PR #9, opened 2019-09-05
[wni] Fix inital framework

Bowenislandsong:
This PR contains 2 commits:
 - fixing make build and lint problems by announcing the package name,
 - pining dependent package version.
These fixes are important for others to contribute.

/cc @aravindhp 
/cc @ravisantoshgudimetla


tools/windows-node-installer/go.mod
```diff
@@ -17,3 +19,9 @@ require (
 	k8s.io/utils v0.0.0-20190809000727-6c36bc71fc4a // indirect
 	sigs.k8s.io/controller-runtime v0.2.0
 )
+
+replace (
+	k8s.io/api => k8s.io/api v0.0.0-20190620084959-7cf5895f2711 // kubernetes-1.15.0
```
aravindhp: I am not sure if this will be used with 4.2 for the drop, so we might want to stick to kube v1.14

tools/windows-node-installer/window-node-installer.go
```diff
@@ -0,0 +1,4 @@
+package windows_node_installer
+// This is a placeholder file to announce package name for `go.mod`.
+// TODO:This file should be replaced by main package in the future and the `windows_node_installer` package should be
```
aravindhp: I am not sure I follow. Anyhow, please create as much of the framework needed that will allow the code to compile while not adding files that would need to removed or replaced.

tools/windows-node-installer/go.mod
ravisantoshgudimetla: Was the comment manually added?

tools/windows-node-installer/window-node-installer.go
ravisantoshgudimetla: I think what @Bowenislandsong meant is he just want to have `windows_node_installer` as package without which go build would throw an error and this file will be renamed as main.go. Is my understanding right @Bowenislandsong? If yes, can you please change the comments?

tools/windows-node-installer/go.mod
Bowenislandsong: Yes, the comment about the version name is manually added. Since go build always changes the version name to time and commit number, I think the comment would help and will not be erased.

tools/windows-node-installer/window-node-installer.go
Bowenislandsong: Ok, I will make the change.

PR #10, opened 2019-09-06
[wni-azure] Azure windows node management code changes - WIP

vinaykns:
This extends to the windows-node-installer project structure
by implementing the two interfaces defined. This is specific
to the windows node management on Azure Platform.


tools/windows-node-installer/pkg/client/openshift.go
```diff
@@ -44,7 +44,7 @@ func (o *OpenShift) GetCloudProvider() (*v1.PlatformStatus, error) {
 	if infra.Status == (v1.InfrastructureStatus{}) || infra.Status.PlatformStatus == nil {
 		return nil, fmt.Errorf("error getting infrastructure status")
 	}
-	return infra.Status.PlatformStatus, nil
+	return infra.Status, nil
```
vinaykns: This is an accidental change, will fix this in the following commits.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
 // this function requires specifying the credentialAccountID of the user's credential account.
 // The resourceTrackerDir is where the `windows-node-installer.json` file which contains IDs of created instance and
 // security group will be created.
-func CloudProviderFactory(kubeconfigPath, credentialPath, credentialAccountID, resourceTrackerDir string) (Cloud, error) {
+
+//struct to hold azure specific information.
+type svc struct {
```
ravisantoshgudimetla: Please change the variable name to something like `azureProvider`, check the existing convention.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
...
+	resourceGroupName     string // name of the resource group
```
ravisantoshgudimetla: Have comments above the field

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -1,7 +1,10 @@
 package cloudprovider
```
ravisantoshgudimetla: You cannot add everything in cloudprovider directory, please add your provider(azure) specific directory

tools/windows-node-installer/pkg/cloudprovider/vm-quickstart-params.json
```diff
@@ -0,0 +1,32 @@
...
+    "osDisk_name": {
+        "value": "_OsDisk_1_2e3ae1ad37414eaca81b432401fcdd75"
+    },
+    "vm_user": {
+        "value": "quickstart"
+    },
+    "vm_password": {
+        "value": "_"
+    }
+}
```
vinaykns: This is a duplicate copy and it need to be deleted in the following changes.

tools/windows-node-installer/pkg/cloudprovider/vm-quickstart-template.json
```diff
@@ -0,0 +1,352 @@
...
+      "apiVersion": "2017-06-01",
+      "scale": null,
+      "properties": {
+        "provisioningState": "Succeeded",
+        "addressPrefix": "10.0.0.0/24"
+      },
+      "dependsOn": [
+        "[resourceId('Microsoft.Network/virtualNetworks', parameters('virtualNetworks_GoQSVM_vnet_name'))]"
+      ]
+    }
```
vinaykns: This is a duplicate copy and will be taken care of in the following commit

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -25,6 +54,25 @@ func CloudProviderFactory(kubeconfigPath, credentialPath, credentialAccountID, r
...
+			log.Fatal("Unable to read template file")
+			return nil, err
+		}
+		params, err := readJSON("windowsnode-params.json")
+		if err != nil {
+			log.Fatal("Unable to read parameters file")
+			return nil, err
+		}
+		azureinfo := svc{resourceGroupName: cloudProvider.Azure.ResourceGroupName,
+			resourceGroupLocation: "centralus", deploymentName: "windowsNode",
```
ravisantoshgudimetla: Remove hardcodings

tools/windows-node-installer/pkg/cloudprovider/vm-quickstart-params.json
```diff
@@ -0,0 +1,32 @@
+{
```
ravisantoshgudimetla: Add some comments as to why we need this

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -1,7 +1,10 @@
 package cloudprovider
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"log"
```
aravindhp: Please use `logr`

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
...
+//struct to hold azure specific information.
+type svc struct {
+	resourceGroupName     string // name of the resource group
+	resourceGroupLocation string // name of the resource location
+
+	deploymentName string                 // name of the deployment
+	templateFile   map[string]interface{} // name of the templateFile (need from user)
+	parametersFile map[string]interface{} // name of the parametersFile (need from user)
+	authdata       map[string]interface{} // Azure auth data
+}
```
aravindhp: Please add the comments on a separate line. Moreover it is obvious that `resourceGroupName` is the name of the resource group. You need to explain what they actually mean from an Azure POV.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
...
+
+func readJSON(path string) (map[string]interface{}, error) {
```
aravindhp: Please add documentation.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
...
-func CloudProviderFactory(kubeconfigPath, credentialPath, credentialAccountID, resourceTrackerDir string) (Cloud, error) {
+
+//struct to hold azure specific information.
+type svc struct {
+	resourceGroupName     string // name of the resource group
+	resourceGroupLocation string // name of the resource location
+
+	deploymentName string                 // name of the deployment
+	templateFile   map[string]interface{} // name of the templateFile (need from user)
+	parametersFile map[string]interface{} // name of the parametersFile (need from user)
```
aravindhp: Are you saying that the user has to provide these JSON files as input? I am a nack on that and I would like to know why you chose this path. I prefer taking whatever inputs are required as command line parameters. More over not all inputs need to be given by the user as we can apply sensible defaults.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -13,7 +16,33 @@ import (
...
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Fatal("failed to read file: %v", err)
+	}
+	contents := make(map[string]interface{})
+	json.Unmarshal(data, &contents)
+	return contents, nil
+}
+
+func CloudProviderFactory(
```
aravindhp: Please don't make such changes. These inputs should remain on the same line. Moreover you inserted new code at the incorrect spot taking away the documentation existing for the function.

tools/windows-node-installer/pkg/cloudprovider/factory.go
aravindhp: These should be inputs that the user provides. Is the `deploymentName` the name of the VM? If yes please match whatever naming convention the installer is following for Linux nodes. Just add `windows` to it. Example: `cluster-name-abc12-windows-worker-us-east-1a-cnpfl`.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -25,6 +54,25 @@ func CloudProviderFactory(kubeconfigPath, credentialPath, credentialAccountID, r
 	}
 
 	switch provider := cloudProvider.Type; provider {
+	case "Azure":
+		//Gather the azure related information.
+		//Read the .auth file to get the subscrptionID and other details.
+		data, err := readJSON("azureauth_location")
```
aravindhp: You should not proceed if there is an error here.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -25,6 +54,25 @@ func CloudProviderFactory(kubeconfigPath, credentialPath, credentialAccountID, r
...
+		template, err := readJSON("windowsnode-template.json")
+		if err != nil {
+			log.Fatal("Unable to read template file")
```
aravindhp: Please remove `[log.Fatal()](https://golang.org/pkg/log/#Fatal)` calls in all cases. They will cause the program to exit making your next return statement meaningless. It is also a bad pattern. Only the main function should end up calling `log.Fatal()`

tools/windows-node-installer/pkg/cloudprovider/factory.go
vinaykns: make sense.

tools/windows-node-installer/pkg/cloudprovider/factory.go
vinaykns: You're right. We could approach this two ways, use the exec plugin to execute the az cli commands  or use the azure-sdk-for-go module and supply the arguments to the function that creates an instance. I'm following the second path to create the instance and it is expecting to supply those template and parameters JSON files. But if we can choose the first path then the job would be way simpler.

tools/windows-node-installer/pkg/cloudprovider/factory.go
vinaykns: will try to add more detail in it.

tools/windows-node-installer/pkg/cloudprovider/factory.go
vinaykns: will fix that.

tools/windows-node-installer/pkg/cloudprovider/factory.go
aravindhp: We should chat about this before you proceed.

PR #11, opened 2019-09-06
[docs] Update CONTRIBUTION guideline

leonjia0112:
Add step to pre-PR content that contributor needs to follow before submitting a PR.

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -71,6 +71,9 @@ git tools.
   - Format your code and fix all the spelling/grammatical mistakes.
   - Limit the column length of the code and the comments within the code to 120 characters.
   - Error messages within the code should be limited to a single line.
+  - Update the documentation if your PR is introducing or changing user facing functionality.
+  - Ensure each PR commit compiles and all unit and e2e tests pass on your local machine.
+    - PRs that have vendor commits are an exception to this rule.
 - Following are the things you should keep in mind once you open a PR:
   - Wait for comments from at least 2 reviewers before pushing changes.
   Open comments in the meanwhile can be worked on locally.


CONTRIBUTION.md
```diff
@@ -71,6 +71,9 @@ git tools.
   - Format your code and fix all the spelling/grammatical mistakes.
   - Limit the column length of the code and the comments within the code to 120 characters.
   - Error messages within the code should be limited to a single line.
+  - Document the details of your PR and put a link inside your PR comment
```
aravindhp: Every PR needn't require documentation. Please replace with `Update the documentation if your PR is introducing or changing user facing functionality.`

CONTRIBUTION.md
```diff
@@ -71,6 +71,9 @@ git tools.
...
+  - Make sure your code compiles on your local machine
```
aravindhp: Please replace with `Ensure each PR commit compiles and all unit and e2e tests pass your local machine.`

CONTRIBUTION.md
```diff
@@ -71,6 +71,9 @@ git tools.
...
+  - Run and make sure it passes all unit testing and e2e testing on your local machine
```
aravindhp: You can remove this bullet.

CONTRIBUTION.md
leonjia0112: Is it better if I say `Ensure each PR commit compiles and all unit and e2e tests pass on your local machine`.

aravindhp: Yes, I missed a `on`

CONTRIBUTION.md
```diff
@@ -71,6 +71,8 @@ git tools.
   - Format your code and fix all the spelling/grammatical mistakes.
   - Limit the column length of the code and the comments within the code to 120 characters.
   - Error messages within the code should be limited to a single line.
+  - Update the documentation if your PR is introducing or changing user facing functionality
+  - Ensure each PR commit compiles and all unit and e2e tests pass on your local machine
```
aravindhp: Both your statements are missing `.` at the end. Please be consistent with existing documentation.

leonjia0112: My bad, didn't notice that.

CONTRIBUTION.md
```diff
@@ -71,6 +71,8 @@ git tools.
   - Format your code and fix all the spelling/grammatical mistakes.
   - Limit the column length of the code and the comments within the code to 120 characters.
   - Error messages within the code should be limited to a single line.
+  - Update the documentation if your PR is introducing or changing user facing functionality.
+  - Ensure each PR commit compiles and all unit and e2e tests pass on your local machine.
```
ravisantoshgudimetla: Not sure, if we want to go this route

aravindhp: Which route? e2e or compile? I definitely want each commit to compile.

ravisantoshgudimetla: e2e tests pass - say if we've a vendor commit and the changes it brings in other commit

aravindhp: I am fine having an exception when we are introducing vendor commits. So we can add a line mentioning that. But in general I don't want to see commits that don't compile.

leonjia0112: Do you want to remove `e2e test` or having another bullet point saying that ` pass e2e test is optional for vendor commit but good to have`?

aravindhp: There should be a sub bullet stating that PRs that have vendor commits are an exception to the rule.

ravisantoshgudimetla:
Trying  the magic incantation :)

CONTRIBUTION.md
leonjia0112: Sub bullet added.

PR #46, opened 2019-10-04
[wsu] WSU playbook

sebsoto:
This pull request introduces the Windows Scale Up Playbook. The purpose of
this playbook is to prepare a Windows instance to join an OpenShift
cluster, run the WMCB, and ensure that the node has successfully
joined the cluster. Specifically, this pull request implements the
first goal, preparing the instance.

An instance is 'prepared', when it is ready to run the WMCB.
Specifically, when it has a kubelet executable, WMCB executable,
and worker ignition file present on the machine.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,6 +28,8 @@ wmcb run --ignition-file $IGNITION_FILE_PATH --kubelet-path $KUBELET_PATH
 
 ## Testing
 
+### Windows Machine Config Bootstrapper
+
 On an existing Windows instance which is ready to join the cluster, copy the worker ignition file to C:\Windows\Temp\worker.ign, and the kubelet to C:\Windows\Temp\kubelet.exe
 
 On the Windows instance, run:
@@ -45,3 +47,7 @@ oc adm certificate approve $CSR_NAME
 Repeat the above steps for the `system:node:$NODE_NAME` csr which will appear shortly. 
 
 You can now do `oc get nodes`, and see the Windows node has joined the cluster.
+
+### Ansible
+
+Follow the instructions in `tools/ansible/README.md`, and ensure the playbook completes successfully.
--- a/tools/ansible/README.md
+++ b/tools/ansible/README.md
@@ -1,0 +1,43 @@
+# Ansible Playbooks
+
+## Prerequisites for using these playbooks
+- Access to an OpenShift cluster
+- A Windows Server 2019 instance within the cluster's subnet, with WinRM [configured](https://docs.ansible.com/ansible/latest/user_guide/windows_setup.html#winrm-setup) for ansible
+    - You can use the [wni](https://github.com/openshift/windows-machine-config-operator/tree/master/tools/windows-node-installer) tool to create a ready Windows instance on certain cloud providers
+    - If you are using a cloud provider, you may have to add a tag to the Windows instance.
+      It will be of the format `key:kubernetes.io/cluster/<infraID>` with `value: owned`.
+      - You can find the infraID in the Ignition config file metadata `metadata.json`
+- A `hosts` file with the required variables defined. See below for an example:
+```
+[win]
+# Address of a node and its Windows password
+<node_ip> ansible_password=<password>
+
+[win:vars]
+# Windows username, typically 'Administrator'
+ansible_user=<username>
+# Address of the OpenShift cluster e.g. "foo.fah.com". This should not include "https://api-" or a port
+cluster_address=<address>
+
+ansible_connection=winrm
+ansible_ssh_port=5986
+# Required if you do not wish to set up a certificate
+#ansible_winrm_server_cert_validation=ignore
+```
+Confirm that you are able to connect your Windows instance with ansible by using the following command:
+```
+$ ansible win -i hosts -m win_ping -v
+```
+
+
+## Windows Scale Up (WSU) Playbook
+This playbook is responsible for three things:
+- Preparing a Windows instance for joining an OpenShift cluster.
+- Running the WMCB
+- Ensuring the Windows node has successfully joined the cluster
+
+### Usage
+Run the WSU playbook:
+```
+$ ansible-playbook -i hosts tasks/wsu/main.yaml -v
+```
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -1,0 +1,58 @@
+---
+- hosts: localhost
+  vars:
+    project_root: "{{ playbook_dir }}/../../../.."
+    kubelet_location: "https://dl.k8s.io/v1.14.0/kubernetes-node-windows-amd64.tar.gz"
+    wmcb_exe: "wmcb.exe"
+
+  tasks:
+    - name: Gather required files
+      block:
+
+        - name: Create directory to store files to be used later
+          tempfile:
+            state: directory
+          register: tmp_dir
+
+        - name: Build WMCB
+          make:
+            target: build
+            chdir: "{{ project_root }}"
+
+        - name: Move WMCB to temporary directory
+          command: mv "{{ project_root }}/{{ wmcb_exe }}" "{{ tmp_dir.path }}/{{ wmcb_exe }}"
+
+        - name: Download Windows node kubelet
+          get_url:
+            url: "{{ kubelet_location }}"
+            dest: "{{ tmp_dir.path }}/kube.tar.gz"
+
+        - name: Extract kubelet
+          unarchive:
+            src: "{{ tmp_dir.path }}/kube.tar.gz"
+            dest: "{{ tmp_dir.path }}"
+
+        - name: Grab kubelet from extracted directories
+          copy:
+            src: "./tmp/kubernetes/node/bin/kubelet.exe"
+            dest: "./tmp/kubelet.exe"
+
+- hosts: win
+  vars:
+    tmp_path: "{{ playbook_dir }}/tmp"
+
+  tasks:
+    - name: Create temporary directory
+      win_tempfile:
+        state: directory
+      register: win_temp_dir
+    - name: Copy required files to Windows host
+      win_copy:
+        src: "{{ hostvars['localhost']['tmp_dir']['path'] }}/"
+        dest: "{{ win_temp_dir.path }}"
+
+    - name: Get ignition file
+      win_get_url:
+        url: "https://api-int.{{ cluster_address }}:22623/config/worker"
+        dest: "{{ win_temp_dir.path }}\\worker.ign"
+        validate_certs: no


tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+    project_root: "{{ playbook_dir }}/../../../.."
+    kubelet_location: "https://dl.k8s.io/v1.14.0/kubernetes-node-windows-amd64.tar.gz"
+    tmp_path: "/tmp/wsu"
+
+  tasks:
+    - name: Gather required files
+      block:
+
+        - name: Create directory to store files to be used later
+          file:
```
suhanime: Can use tempfile module instead and register the variable

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+            path: "{{ tmp_path }}"
+            state: directory
+
+        - name: Build WMCB
+          make:
+            target: build
+            chdir: "{{ project_root }}"
+        - name: Move WMCB to temporary directory
+          command: mv "{{ project_root }}/wmcb.exe" "{{ tmp_path }}/wmcb.exe"
```
suhanime: Can't we register the variable in `Build WMCB` instead of specifying `wmcd.exe`?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+- hosts: win
+  vars:
+    tmp_path: "{{ playbook_dir }}/tmp"
+
+  tasks:
+    - name:  Copy required files to Windows host
+      win_copy:
+        src: "{{ tmp_path }}/"
+        dest: "C:\\Windows\\Temp"
```
suhanime: Windows has a win_tempfile module too. And I think we should delete the tempfile created on Windows host after the task is done

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+        backup: yes
+
+    - name: Get ignition file
+      win_get_url:
```
suhanime: What is the contingency of this failing? Maybe add a timeout?

Makefile
```diff
@@ -14,6 +14,9 @@ GO_BUILD_ARGS=CGO_ENABLED=0 GO111MODULE=on
 build:
 	$(GO_BUILD_ARGS) GOOS=windows go build -o wmcb.exe  $(MAIN_PACKAGE)
 
+run:
```
ravisantoshgudimetla: `run` what? Are you assuming that ansible and other packages are installed as part of `build` then? I wouldn't make changes in the make file unless, I am 100% of how the flow is going to look like.

tools/ansible/tasks/wsu/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+[win:vars]
+# Windows username, typically 'Administrator'
+ansible_user=<username>
+# Address of the OpenShift cluster e.g. "foo.fah.com". This should not include "https://api-" or a port
+cluster_address=<address>
+
+ansible_connection=winrm
+ansible_ssh_port=5986
+ansible_winrm_server_cert_validation=ignore
```
ravisantoshgudimetla: Comment out this is not secure

tools/ansible/tasks/wsu/README.md
```diff
@@ -0,0 +1,38 @@
...
+- A `hosts` file with the required variables defined. See below for an example:
+```
+[win]
+# Address of a node and its Windows password
+<node_ip> ansible_password=<password>
+
+[win:vars]
+# Windows username, typically 'Administrator'
+ansible_user=<username>
+# Address of the OpenShift cluster e.g. "foo.fah.com". This should not include "https://api-" or a port
```
ravisantoshgudimetla: Why this restriction of not providing complete endpoint? isn't it confusing?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
+---
+- hosts: localhost
+  vars:
+    project_root: "{{ playbook_dir }}/../../../.."
```
ravisantoshgudimetla: I'm fine with this for now but things to keep in mind:
How well this plays well with our CI like downloading files onto pods and then transferring them. One option would be have a flag to skip downloading the binaries and let them build locally and transferred.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+        - name: Create directory to store files to be used later
+          file:
+            path: "{{ tmp_path }}"
+            state: directory
+
+        - name: Build WMCB
+          make:
+            target: build
+            chdir: "{{ project_root }}"
```
ravisantoshgudimetla: Should we build this locally? What if we cut a release which has WMCB binary? Shouldn't we be downloading it from there, do you see any problem with this approach?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+            dest: "{{ tmp_path }}"
+
+        - name: Grab kubelet from extracted directories
+          copy:
+            src: "./tmp/kubernetes/node/bin/kubelet.exe"
+            dest: "./tmp/kubelet.exe"
+
+- hosts: win
+  vars:
+    tmp_path: "{{ playbook_dir }}/tmp"
```
ravisantoshgudimetla: Why create a new variable here? Do you forsee any use of this?

Makefile
sebsoto: Run the WMCB. I'll gladly remove it from the makefile.

tools/ansible/tasks/wsu/README.md
sebsoto: Personally I dont think its confusing. I think most people would put the `api` endpoint by default, not `api-int`, and this makes things easier.

tools/ansible/tasks/wsu/main.yaml
sebsoto: There are issues using tempfile with remote copy that I was running into. I'm unable to find the github issue that confirmed my problem, but switching to using /tmp/ fixed it.

tools/ansible/tasks/wsu/main.yaml
sebsoto: I think for now we should build it locally until we have a hosting solution.

sebsoto: Its definitely something we should change down the road though.

tools/ansible/tasks/wsu/main.yaml
sebsoto: We need to recreate the variable for each host.

tools/ansible/tasks/wsu/main.yaml
sebsoto: Looks like theres a default timeout of 10 seconds:
https://docs.ansible.com/ansible/latest/modules/win_get_url_module.html

tools/ansible/tasks/wsu/main.yaml
sebsoto: Good point, will change

tools/ansible/tasks/wsu/main.yaml
sebsoto: Good point. That should definitely be addressed in a future commit

tools/ansible/tasks/wsu/README.md
sebsoto: I will comment it out and document that it may be needed

tools/ansible/tasks/wsu/main.yaml
sebsoto: I think there is some merit in having files be in a consistent spot, for predictability and debugging purposes, since we are interacting with these files with wmcb.exe.
I think we can change this in the future if we don't find the consistency useful though

tools/ansible/tasks/wsu/main.yaml
sebsoto: I was mistaken, changed this.

tools/ansible/tasks/wsu/main.yaml
sebsoto: Got tempfile working in accordance with your instructions!

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: My plan, is to include the WMCB binary as part of release(this handles the hosting), what about having a variable which tells to skip this step(something like `skip if !development mode`)

docs/README.md
```diff
@@ -28,6 +28,8 @@ wmcb run --ignition-file $IGNITION_FILE_PATH --kubelet-path $KUBELET_PATH
 
 ## Testing
 
+### Bootstrapper
```
aravindhp: I am not fan of using bootstrapper. I prefer `Windows Machine Config Bootstrapper`

tools/ansible/tasks/wsu/README.md
aravindhp: Can we provide a command `oc get routes...` that will help the user?

tools/ansible/tasks/wsu/main.yaml
aravindhp: This indeed looks hairy

tools/ansible/tasks/wsu/main.yaml
aravindhp: I would like to have the latest version of the `WMCB` binary to be used. I guess this would become easier once we have the binary published as part of CI like it happens for images.

tools/ansible/tasks/wsu/main.yaml
aravindhp: I prefer that we use the `win_tempfile` module if you are going the temp directory route. If you want a consistent spot, then create one outside of the `temp` directory structure. Otherwise what guarantee do we have for the files to be present there as the admin could clear the global temp directory.

docs/README.md
sebsoto: Changed

tools/ansible/tasks/wsu/main.yaml
sebsoto: Changed

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: I am fine with future commit.

tools/ansible/tasks/wsu/main.yaml
sebsoto: I would like to address this in a future PR

tools/ansible/README.md
```diff
@@ -0,0 +1,42 @@
+# Ansible Playbooks
+
+## Prerequisites for using these playbooks
+- Access to an OpenShift cluster
+- A Windows Server 2019 instance within the cluster's subnet, with WinRM configured for ansible: https://docs.ansible.com/ansible/latest/user_guide/windows_setup.html#winrm-setup
+    - You can use the [wni](https://github.com/openshift/windows-machine-config-operator/tree/master/tools/windows-node-installer) tool to create a ready Windows instance on certain cloud providers
```
aravindhp: Please limit to 120 col. Applies to all instances.

PR #59, opened 2019-10-14
[wni] Fix broken go.sum

ravisantoshgudimetla:
As of now, go.sum on master branch is broken.
This commit should fix it.

/cc @aravindhp @vinaykns


ravisantoshgudimetla:
> Please update your PR and commit header to reflect what you are fixing. i.e. [wni] Fix broken go.sum. Please say which PR introduced the issue and how you are fixing it in the commit and PR message.

PR #24 introduced the problem. The problem is not limited to go.mod and go.sum within wni directory as we initially thought, the go.mod in wmco root is also effected because of the #24 changes. Fixed it using `go mod verify` in both the locations. 

Updated the commit message to reflect it.

go.mod
```diff
@@ -12,16 +12,9 @@ replace (
 )
 
 require (
-	github.com/Azure/azure-sdk-for-go v33.4.0+incompatible // indirect
```
ravisantoshgudimetla: These are not needed in go.sum of wmco dir.

sebsoto:
/lgtm
Ran `make all` and go.sum is the same

sebsoto:
Holding in case other reviewer is needed

PR #86, opened 2019-11-09
[wmcb] Register kubelet with Windows taint

ravisantoshgudimetla:
As of now, linux pods can be scheduled onto Windows
VM because there is no taint on the node which will
prevent it. This commit addresses that problem
Also, updated the readme with a sample pod
spec for the Windows pods.

~~I'll add an e2e once I start working on the `against the cluster` test suite.~~

Added an e2e which will not be running yet.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,6 +7,18 @@ jobs that the bootstrapper does:
 - Ensures that the kubelet gets the correct kubelet config
 - Run the kubelet as a windows service
 
+Once the bootstrapper has been run and the CSR associated with the Windows node is approved, the Windows 
+node will have a taint called `os=Windows:NoSchedule`, only the pods with matching toleration can 
+be scheduled onto the Windows node. An example pod spec with the toleration would be:
+
+```
+tolerations:
+  - key: "os"
+    operator: "Equal"
+    value: "Windows"
+    effect: "NoSchedule"
+```
+
 This will be remotely invoked from a Ansible script or can be run locally
 
 ## Requirements
--- a/internal/test/wmcb_test.go
+++ b/internal/test/wmcb_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"k8s.io/client-go/tools/clientcmd"
 	"log"
 	"os"
 	"testing"
@@ -17,6 +18,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/crypto/ssh"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sclient "k8s.io/client-go/kubernetes"
 )
 
 const (
@@ -29,6 +33,13 @@ const (
 	// winrm port to be used
 	winRMPort = 5986
 )
+
+// windowsTaint is the taint that needs to be applied to the Windows node
+var windowsTaint = v1.Taint{
+	Key:    "os",
+	Value:  "Windows",
+	Effect: v1.TaintEffectNoSchedule,
+}
 
 // testFramework holds the info to run the test suite
 type testFramework struct {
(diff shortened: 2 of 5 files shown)


ravisantoshgudimetla:
Sample Windows worker when tested manually on the node.

```
$ oc describe node/ip-10-0-52-217.ec2.internal
Name:               ip-10-0-52-217.ec2.internal
Roles:              none
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=m4.large
                    beta.kubernetes.io/os=windows
                    failure-domain.beta.kubernetes.io/region=us-east-1
                    failure-domain.beta.kubernetes.io/zone=us-east-1d
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=ec2amaz-83bb2ln
                    kubernetes.io/os=windows
Annotations:        ovn_host_subnet=10.130.2.0/23
                    volumes.kubernetes.io/controller-managed-attach-detach=true
CreationTimestamp:  Fri, 08 Nov 2019 17:54:09 -0500
**Taints:             os=Windows:NoSchedule**
Unschedulable:      false
```

docs/README.md
```diff
@@ -7,6 +7,18 @@ jobs that the bootstrapper does:
 - Ensures that the kubelet gets the correct kubelet config
 - Run the kubelet as a windows service
 
+Once the bootstrapper has been and CSR associated with the Windows node is approved, the Windows node will have a
```
sebsoto: once the bootstrapper has been what?

sebsoto: `and CSR` -> `and the CSR`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -298,6 +310,11 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 		"--windows-service",
 		"--logtostderr=false",
 		"--log-file=" + filepath.Join(wmcb.installDir, "kubelet.log"),
+		// Registers the Kubelet with Windows specific taints so that linux pods won't get scheduled onto
+		// Windows nodes.
+		// TODO: Write a `against the cluster` e2e test which checks for the Windows node object created
```
sebsoto: Why is this a `TODO` and not part of this PR?

ravisantoshgudimetla: I wanted to include this as part of e2e test suite which we don't have yet.

internal/test/wmcb_test.go
```diff
@@ -218,6 +229,41 @@ func TestWMCBUnit(t *testing.T) {
...
+}
+
+// TestWMCBCluster runs the cluster tests for the nodes
+func TestWMCBCluster(t *testing.T) {
+	//TODO: Transfer the WMCB binary to the Windows node and approve CSR for the Windows node.
+	// I want this to be moved to another test. We've another card for this, so let's come back
+	// to that later(WINC-82). As of now, this test is limited to check if the taint has been
+	// applied to the Windows node and skipped for now.
+	client, err := getKubeClient()
+	require.NoError(t, err, "error getting kubeclient")
```
sebsoto: Please print the error that getKubeClient() is returning for future debugging purposes :)
```
require.NoErrorf(t, err, "error getting kubeclient: %s", err)
```

internal/test/wmcb_test.go
```diff
@@ -218,6 +229,41 @@ func TestWMCBUnit(t *testing.T) {
...
+	winNodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "kubernetes.io/os=windows"})
+	require.NoError(t, err, "error while getting Windows node")
```
sebsoto: Same as above

internal/test/wmcb_test.go
ravisantoshgudimetla: Done.

internal/test/wmcb_test.go
ravisantoshgudimetla: Done.

internal/test/wmcb_test.go
```diff
@@ -218,6 +229,41 @@ func TestWMCBUnit(t *testing.T) {
...
+	// applied to the Windows node and skipped for now.
+	client, err := getKubeClient()
+	require.NoErrorf(t, err, "error getting kubeclient: %v", err)
+	winNodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "kubernetes.io/os=windows"})
+	require.NoErrorf(t, err, "error while getting Windows node: %v", err)
+	hasWindowsTaint := false
+	for _, node := range winNodes.Items {
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == windowsTaint.Key && taint.Value == windowsTaint.Value && taint.Effect == windowsTaint.Effect {
+				hasWindowsTaint = true
```
aravindhp: You can break out of the loop once this is set.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -314,7 +331,6 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 	if cloudConfigValue, ok := wmcb.kubeletArgs[cloudConfigOption]; ok {
 		kubeletArgs = append(kubeletArgs, "--"+cloudConfigOption+"="+cloudConfigValue)
 	}
-
```
aravindhp: Why?

internal/test/wmcb_test.go
ravisantoshgudimetla: Done.

ravisantoshgudimetla: Done

internal/test/wmcb_test.go
```diff
@@ -218,6 +229,42 @@ func TestWMCBUnit(t *testing.T) {
...
+				break
+			}
+		}
```
aravindhp: Nit: ideally you need a `if hasWindowsTaint { break }` here

PR #94, opened 2019-11-14
[wmcb] Change node labels during kubelet registration

ravisantoshgudimetla:
Change node labels during kubelet registration
from 'node-role.kubernetes.io' to
'node.openshift.io/os_id=Windows'.The main problem
is https://github.com/kubernetes/kubernetes/issues/84912.
There we're restricting the labels to be added
in kubernetes.io namespace. Now, I am changing
it to label that other entities could
watch for like future controllers or query from like
WSU.

The plan is to use this label in WSU in short term. 
Will create a PR for that soon.

/cc @aravindhp @sebsoto

### DIFF
--- a/internal/test/wmcb_test.go
+++ b/internal/test/wmcb_test.go
@@ -32,14 +32,18 @@ const (
 	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
 	// winrm port to be used
 	winRMPort = 5986
+	// nodeLabels represents the node label that need to be applied to the Windows node created
+	nodeLabel = "node.openshift.io/os_id=Windows"
 )
 
-// windowsTaint is the taint that needs to be applied to the Windows node
-var windowsTaint = v1.Taint{
-	Key:    "os",
-	Value:  "Windows",
-	Effect: v1.TaintEffectNoSchedule,
-}
+var (
+	// windowsTaint is the taint that needs to be applied to the Windows node
+	windowsTaint = v1.Taint{
+		Key:    "os",
+		Value:  "Windows",
+		Effect: v1.TaintEffectNoSchedule,
+	}
+)
 
 // testFramework holds the info to run the test suite
 type testFramework struct {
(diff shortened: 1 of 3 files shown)


pkg/bootstrapper/bootstrapper.go
```diff
@@ -62,8 +62,10 @@ const (
 		      	effect: "NoSchedule"
 	*/
 	windowsTaints = "os=Windows:NoSchedule"
-	// workerLabel contains the label that needs to be applied to the worker nodes in the cluster
-	workerLabel = "node-role.kubernetes.io/worker"
+	// nodeLabel contains the os specific label that will be applied to the Windows node object. This can be used to
+	// identify the nodes managed by WSU and future operators. (We could have gotten this from boostrap kubeconfig too
```
sebsoto: Theres a `(` but no `)` here

internal/test/wmcb_test.go
```diff
@@ -253,16 +255,24 @@ func TestWMCBCluster(t *testing.T) {
...
 	require.NoErrorf(t, err, "error while getting Windows node: %v", err)
-	hasWindowsTaint := false
-	for _, node := range winNodes.Items {
+	assert.Equalf(t, hasWindowsTaint(winNodes.Items), true, "expected Windows Taint to be present on the Windows Node")
+	winNodes, err = client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nodeLabel})
+	require.NoErrorf(t, err, "error while getting Windows node: %v", err)
+	assert.Lenf(t, winNodes.Items, 1, "expected one node to have node label but found: %v", len(winNodes.Items))
+}
+
+// hasWindowsTaints returns true if the given Windows node has
```
sebsoto: If the given windows node has what?

vinaykns: nit: hasWindowsTaint
Also in the func docs, has what..?

internal/test/wmcb_test.go
```diff
@@ -41,6 +41,8 @@ var windowsTaint = v1.Taint{
 	Effect: v1.TaintEffectNoSchedule,
 }
 
+var nodeLabel = "node.openshift.io/os_id=Windows"
```
vinaykns: may be combining all the variable declarations at a single place, looks more uniform..?.
Also the doc missing for the variable.

pkg/bootstrapper/bootstrapper_test.go
```diff
@@ -253,32 +253,6 @@ func TestCloudConfNotPresent(t *testing.T) {
 	assert.False(t, present, "cloud-config option is not present in kubelet args")
 }
 
-// TestKubeletArgs tests if proper kubelet arguments are present after parsing the ignition file
-func TestKubeletArgs(t *testing.T) {
```
vinaykns: just to know why is this test case removed..?

sebsoto: Please call this out in the commit message @ravisantoshgudimetla

ravisantoshgudimetla: I was using this to pass check for the kubelet args obtained after parsing ignition file, specifically nodelabels which I am not doing anymore. I updated the commit message.

PR #97, opened 2019-11-18
[wmcb] Cleanup assert usage

vinaykns:
Currently in wmcb, there are few places we are
causing NPE. So this PR will ensure that assert usage
is replaced with require at appropriate places. Also
Nilf is replaces with NoErrorf for more readability.

### DIFF
--- a/test/e2e/bootstrapper_test.go
+++ b/test/e2e/bootstrapper_test.go
@@ -52,11 +52,11 @@ func TestBootstrapper(t *testing.T) {
 
 	// Run the bootstrapper, which will start the kubelet service
 	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath)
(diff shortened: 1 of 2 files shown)


vinaykns:
Tested the PR with e2e testing instructions specified in the README.

test/e2e/bootstrapper_test.go
```diff
@@ -76,11 +76,11 @@ func TestBootstrapper(t *testing.T) {
 	// Run it again, to ensure it maintains state if the bootstrapper is already started
 	time.Sleep(5 * time.Second)
 	wmcb, err = bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath)
-	assert.Nilf(t, err, "Could not create WinNodeBootstrapper: %s", err)
+	assert.NoErrorf(t, err, "Could not create WinNodeBootstrapper: %s", err)
```
aravindhp: This needs to be `require`. I missed this during the initial review but caught it when rebasing #96.

PR #98, opened 2019-11-19
Bug 1775384: [wsu] Upgrade kubelet version to 1.16

suhanime:
Currently WSU playbook downloads kubelet node binary for Kubernetes version 1.14,
however Kubernetes version is upgraded for OpenShift 4.3, from 1.14 to 1.16.
This commit upgrades Kubernetes node binary version to 1.16.2, and also
introduces an e2e test that checks the checksum of binary downloaded to ensure
the binary was downloaded in its entirety.

### DIFF
--- a/tools/ansible/tasks/wsu/test/e2e/wsu_test.go
+++ b/tools/ansible/tasks/wsu/test/e2e/wsu_test.go
@@ -41,6 +41,11 @@ var (
 	createdInstanceCreds *types.Credentials
 	// Temp directory ansible created on the windows host
 	ansibleTempDir = ""
+	// kubernetes-node-windows-amd64.tar.gz SHA512
+	// Value from https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#node-binaries-1
+	// This value should be updated when we change the kubelet version in WSU
+	expectedKubeTarSha = "a88e7a1c6f72ea6073dbb4ddfe2e7c8bd37c9a56d94a33823f531e303a9915e7a844ac5880097724e44dfa7f4" +
+		"a9659d14b79cc46e2067f6b13e6df3f3f1b0f64"
 )
 
 // createAWSWindowsInstance creates a windows instance and populates the "cloud" and "createdInstanceCreds" global
@@ -118,7 +123,7 @@ func TestWSU(t *testing.T) {
 
 // testFilesCopied tests that the files we attempted to copy to the Windows host, exist on the Windows host
 func testFilesCopied(t *testing.T) {
-	expectedFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe"}
+	expectedFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe", "kube.tar.gz"}
 	endpoint := winrm.NewEndpoint(createdInstanceCreds.GetIPAddress(), 5986, true, true,
 		nil, nil, nil, 0)
 	client, err := winrm.NewClient(endpoint, "Administrator", createdInstanceCreds.GetPassword())
@@ -135,4 +140,19 @@ func testFilesCopied(t *testing.T) {
 		assert.Emptyf(t, stdout.String(), "Missing file: %s", fullPath)
 	}
 
+	// Check the SHA of kube.tar.gz downloaded
+	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
+	// certutil is part of default OS installation Windows 7+
+	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
+	stdout := new(bytes.Buffer)
+	stderr := new(bytes.Buffer)
+	_, err = client.Run(command, stdout, stderr)
+	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
+	require.Equalf(t, stderr.Len(), 0, "Error generating SHA512 for %s", kubeTarPath)
+	// CertUtil output example:
+	// SHA512 hash of <filepath>:\r\n<SHA-output>\r\nCertUtil: -hashfile command completed successfully.
+	// Extracting SHA value from the output
+	actualKubeTarSha := strings.Split(stdout.String(), "\r\n")[1]
+	assert.Equal(t, expectedKubeTarSha, actualKubeTarSha,
+		"kube.tar.gz downloaded does not match expected checksum")
 }
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -2,7 +2,7 @@
 - hosts: localhost
   vars:
     project_root: "{{ playbook_dir }}/../../../.."
-    kubelet_location: "https://dl.k8s.io/v1.14.0/kubernetes-node-windows-amd64.tar.gz"
+    kubelet_location: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
     wmcb_exe: "wmcb.exe"
 
   tasks:


tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -41,6 +43,10 @@ var (
 	createdInstanceCreds *types.Credentials
 	// Temp directory ansible created on the windows host
 	ansibleTempDir = ""
+	// Kubelet node binary SHA512
+	// Value from https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#node-binaries-1
```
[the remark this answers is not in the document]
suhanime: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#node-binaries-1 points to the Node binary shas for 16.2, but I will fix the reference to not point to a heading so it isn't as confusing.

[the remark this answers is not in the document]
aravindhp: OK, I see what is going on. I am fine if you leave it alone as long as it is accurate.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -41,6 +41,10 @@ var (
 	createdInstanceCreds *types.Credentials
 	// Temp directory ansible created on the windows host
 	ansibleTempDir = ""
+	// kubernetes-node-windows-amd64.tar.gz SHA512
+	// Value from https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#node-binaries-1
+	expectedSha = "a88e7a1c6f72ea6073dbb4ddfe2e7c8bd37c9a56d94a33823f531e303a9915e7a844ac5880097724e44dfa7f4a9659" +
```
sebsoto: why is this being broken up into different strings? So it doesnt hit the 120 char limit in our style guidelines?

suhanime: Correct

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -135,4 +139,14 @@ func testFilesCopied(t *testing.T) {
 		assert.Emptyf(t, stdout.String(), "Missing file: %s", fullPath)
 	}
 
+	// Check the SHA of kube.tar.gz downloaded
+	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
+	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
+	stdout := new(bytes.Buffer)
+	_, err = client.Run(command, stdout, os.Stderr)
+	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
+	actualSha := strings.Split(strings.Split(stdout.String(), " ")[3], "\n")[1]
```
sebsoto: Please break this up into multiple lines with descriptive variable names at each step and comment what the original string should look like

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -135,4 +139,14 @@ func testFilesCopied(t *testing.T) {
...
+	assert.Equal(t, expectedSha, actualSha[:len(actualSha)-1],
```
sebsoto: What does actualSha look like? Why do you need to take a portion of it? Please explain this in a comment

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
suhanime: Made the change for cleaner extraction and added an example

sebsoto:
Thanks for the change
/lgtm

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -135,4 +139,17 @@ func testFilesCopied(t *testing.T) {
 		assert.Emptyf(t, stdout.String(), "Missing file: %s", fullPath)
 	}
 
+	// Check the SHA of kube.tar.gz downloaded
+	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
+	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
+	stdout := new(bytes.Buffer)
+	_, err = client.Run(command, stdout, os.Stderr)
```
aravindhp: If `command` returns an error, does that get reflected in `err`? I thought you would need to check `Stderr`.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -135,4 +139,17 @@ func testFilesCopied(t *testing.T) {
 		assert.Emptyf(t, stdout.String(), "Missing file: %s", fullPath)
 	}
 
+	// Check the SHA of kube.tar.gz downloaded
+	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
+	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
```
aravindhp: How are you guaranteeing that `certutil` is available?

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
ravisantoshgudimetla: `expectedSha` of what? If you're downloading multiple items, it'll be hard in future. Also, you've to comment that this sha is something that needs to change whenever you're updating the binaries.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
ravisantoshgudimetla: Do you assume that `certutil` exist? If not, where do you capture the error?

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
ravisantoshgudimetla: IIRC, we `Run` would silently fail, it won't throw an error when the command fails to execute, I think you need to capture the std out as well.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -135,4 +139,17 @@ func testFilesCopied(t *testing.T) {
...
+	stdout := new(bytes.Buffer)
+	_, err = client.Run(command, stdout, os.Stderr)
+	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
+	// CertUtil output example:
+	// SHA512 hash of <filepath>:/r/n<SHA-output>/r/nCertUtil: -hashfile command completed successfully.
+	// Extracting SHA value from the output
+	actualSha := strings.Split(stdout.String(), "/r/n")[1]
```
ravisantoshgudimetla: Check if we need to do something similar to this.

https://github.com/openshift/windows-machine-config-operator/pull/98/files#diff-1c2c1c572f356c7644a7be940f4d187dL131

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
suhanime: This is the expectedSha of the node binary tar. This tar includes kubelet and kubectl - which we transfer on the Windows node as a part of WSU. AFAIK, we are only downloading kubernetes node binary from upstream, as a part of WSU. hybrid-overlay and worker.ign resides on our repo of within the cluster. I'm not sure what you mean by it being hard in the future.
For now I can rename this SHA variable to reflect it is kube-tar specific and call out when it needs to change within the comments

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
suhanime: The documentation isn't particularly great in this regards, but looks like certutil is part of default os installation from Windows 7 onwards.
[A forum answer](https://social.technet.microsoft.com/Forums/en-US/09fb7a70-69c2-4cc0-ad82-173c06774261/availbility-of-certutil-on-different-windows-versions?forum=winserversecurity) is the closest I could find.

suhanime: As I mentioned above, looks like certutil should exists for all Windows Server 2019 images. For now I can capture both stdout and stderr in client.Run command in case the execution fails

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
suhanime: I'll capture and impose the stderr check here along with stdout

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
suhanime: I can add the check to ensure kube.tar actually exists here before running the test for sha check

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -102,16 +109,21 @@ func TestWSU(t *testing.T) {
...
+	_, err = errFile.WriteString(string(out))
 	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, string(out))
 
 	// Ansible will copy files to a temporary directory with a path such as:
 	// C:\\Users\\Administrator\\AppData\\Local\\Temp\\ansible.z5wa1pc5.vhn\\
 	initialSplit := strings.Split(string(out), "C:\\\\Users\\\\Administrator\\\\AppData\\\\Local\\\\Temp\\\\ansible.")
 	require.True(t, len(initialSplit) > 1, "Could not find Windows temp dir: %s", out)
 	ansibleTempDir = "C:\\Users\\Administrator\\AppData\\Local\\Temp\\ansible." + strings.Split(initialSplit[1], "\"")[0]
-
+	time.Sleep(60 * time.Minute)
```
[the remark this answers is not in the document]
suhanime: This was a part of discarded commit - I was using an arbitrary sleep for debugging the windows instance.

sebsoto:
@suhanime Theres a failure within the WSU test that is not a flake

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -143,13 +144,14 @@ func testFilesCopied(t *testing.T) {
 	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
 	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
 	stdout := new(bytes.Buffer)
-	_, err = client.Run(command, stdout, os.Stderr)
+	stderr := new(bytes.Buffer)
+	_, err = client.Run(command, stdout, stderr)
 	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
+	require.Equalf(t, stderr.Len(), 0, "Error generating SHA512 for %s", kubeTarPath)
```
sebsoto: Please use `require.Empty()` instead of `require.Equal` otherwise, changes lgtm!

suhanime: Here stderr isn't getting set when there's no error from command execution, so Empty check is failing. "Empty asserts that the specified object is empty. I.e. nil, "", false, 0 or either a slice or a channel with len == 0." Somehow it doesn't check for bytes.Buffer object?
That's why I am enforcing a check on the length of the stderr instead.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
aravindhp: Yes please rename the variable to reflect this

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
aravindhp: Then please add a comment about it.

ravisantoshgudimetla:
This needs to be backported to release-4.3

PR #103, opened 2019-11-26
Refactor tests

ravisantoshgudimetla:
Small refactor to check if framework can be  re-used.


ravisantoshgudimetla:
If this passes, l'll remove the wsu/e2e_test.go file.

internal/test/framework/framework.go
```diff
@@ -1,26 +1,16 @@
 package test
 
 import (
-	"flag"
```
aravindhp: Why does dep updates have code changes? I will review the rest once this is resolved.

internal/test/go.mod
```diff
@@ -2,13 +2,19 @@ module github.com/openshift/windows-machine-config-operator/internal/test
 
 go 1.12
 
+replace (
```
aravindhp: If you ran a command to do this, please mention that in the commit.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -5,14 +5,13 @@ set -o pipefail
...
 # Build the unit test binary
 cd "${WMCO_ROOT}"
 make build-wmcb-unit-test
 
 # Transfer the binary and run the unit tests
-cd "${INTERNAL_TEST_DIR}"
-CGO_ENABLED=0 GO111MODULE=on go test -c wmcb_test.go -o wmcb_framework
-./wmcb_framework -test.run=TestWMCBUnit -binaryToBeTransferred=../../wmcb_unit_test.exe
+cd "${WMCB_TEST_DIR}"
+CGO_ENABLED=0 GO111MODULE=on go test -v -run=TestWMCBUnit -binaryToBeTransferred=../../../wmcb_unit_test.exe -timeout=20m .
```
aravindhp: @sebsoto bumped the timeout to 30m. I suggest we do the same here too.

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+	WinrmClient *winrm.Client
+	// remoteDir is the directory to which files will be transferred to, on the Windows VM
+	RemoteDir string
+	// SSHClient contains the ssh client information to access the Windows VM via ssh
+	SSHClient *ssh.Client
+	// CloudProvider holds the information related to cloud provider
+	cloudProvider cloudprovider.Cloud
+	// k8sclientset is the kubernetes clientset we will use to query the cluster's status
+	K8sclientset *kubernetes.Clientset
+}
```
aravindhp: Do all the fields in the struct need to exported i.e. capitalized?

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+func (f *TestFramework) createWindowsVM() error {
+	// TODO: Add validation for the environment variables to see if they're properly populated or not
+	kubeconfig := os.Getenv("KUBECONFIG")
+	awsCredentials := os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
+	artifactDir := os.Getenv("ARTIFACT_DIR")
+	privateKeyPath := os.Getenv("KUBE_SSH_KEY_PATH")
+	// Use Windows 2019 server image with containers in us-east1 zone for CI testing.
+	// TODO: Move to environment variable that can be fetched from the cloud provider
+	imageID := "ami-0b8d82dea356226d3"
+	instanceType := "m4.large"
```
aravindhp: Please see #106 and use `m5a.large`

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if _, err := f.WinrmClient.Run(remotePowerShellCmdPrefix+"Start-Service ssh-agent",
+		os.Stdout, os.Stderr); err != nil {
+		return fmt.Errorf("start ssh-agent with error %v", err)
+	}
+	if _, err := f.WinrmClient.Run(remotePowerShellCmdPrefix+"Start-Service sshd",
+		os.Stdout, os.Stderr); err != nil {
+		return fmt.Errorf("failed to start sshd with error %v", err)
+	}
+	return nil
```
aravindhp: In the above message, please replace `with error` with `:` to make the message more succinct.

internal/test/framework/framework.go
ravisantoshgudimetla: Not all of them, for example `cloudProvider` doesn't need to be exported. But as I told at https://github.com/openshift/windows-machine-config-operator/pull/103/files#diff-5465ccc52b3a24461a98f029910cf60aR29, this is not clear, ideally we should have getters and setters around this

internal/test/framework/framework.go
ravisantoshgudimetla: Done

internal/test/go.mod
ravisantoshgudimetla: No, I just manually added them

internal/test/framework/framework.go
ravisantoshgudimetla: Thanks for catching it, I thought I already did this while doing rebase but obviously not and this could be causing the failures we're noticing in CI

ravisantoshgudimetla:
> Sorry for not catching this before. Please replace [test] with [ci] in your commit message.

No problem, I just updated it. I started with `[ci]` but it did not make much sense as I thought this is not specific just specific to CI so I changed it to `test`. I've changed it now.

internal/test/wsu/wsu_test.go
```diff
@@ -178,25 +101,25 @@ func TestWSU(t *testing.T) {
...
 	// Check if each of the files we expect on the Windows host are there
 	for _, filename := range expectedFileList {
 		fullPath := ansibleTempDir + "\\" + filename
 		// This command will write to stdout, only if the file we are looking for does not exist
 		command := fmt.Sprintf("if not exist %s echo fail", fullPath)
 		stdout := new(bytes.Buffer)
-		_, err := winRMClient.Run(command, stdout, os.Stderr)
+
+		_, err := framework.WinrmClient.Run(command, stdout, os.Stderr)
+
```
sebsoto: Is this newline intentional?

internal/test/wsu/wsu_test.go
```diff
@@ -103,69 +70,25 @@ ansible_winrm_server_cert_validation=ignore`, ip, password, clusterAddress))
...
 	// In order to run the ansible playbook we create an inventory file:
 	// https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html
-	hostFilePath, err := createHostFile(createdInstanceCreds.GetIPAddress(), createdInstanceCreds.GetPassword())
+	hostFilePath, err := createHostFile(framework.Credentials.GetIPAddress(), framework.Credentials.GetPassword())
 	require.NoErrorf(t, err, "Could not write to host file: %s", err)
 	cmd := exec.Command("ansible-playbook", "-vvv", "-i", hostFilePath, playbookPath)
 	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, string(out))
 
+	k8sclientset = framework.K8sclientset
```
sebsoto: Shouldnt we delete the k8sclientset and just use framework.K8sclientset at this point?

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+	// TODO: Add some options to skip certain parts of the test
+	if err := f.setupWinRMClient(); err != nil {
+		log.Fatalf("failed to setup winRM client for the Windows VM: %v", err)
+	}
+	// Wait for some time before starting configuring of ssh server. This is to let sshd service be available
+	// in the list of services
+	// TODO: Parse the output of the `Get-Service sshd, ssh-agent` on the Windows node to check if the windows nodes
+	// has those services present
+	time.Sleep(time.Minute)
+	if err := f.configureOpenSSHServer(); err != nil {
```
sebsoto: This is only needed for the WMCB tests, should it be a shared step?

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+		log.Fatalf("failed to configure OpenSSHServer on the Windows VM: %v", err)
+	}
+	if err := f.createRemoteDir(); err != nil {
```
sebsoto: Same as above

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+		log.Fatalf("failed to create remote dir with error: %v", err)
+	}
+	if err := f.getSSHClient(); err != nil {
```
sebsoto: Same as above

internal/test/framework/framework.go
```diff
@@ -0,0 +1,198 @@
...
+	RemoteDir string
+	// SSHClient contains the ssh client information to access the Windows VM via ssh
+	SSHClient *ssh.Client
+	// CloudProvider holds the information related to cloud provider
+	cloudProvider cloudprovider.Cloud
+	// k8sclientset is the kubernetes clientset we will use to query the cluster's status
+	K8sclientset *kubernetes.Clientset
+}
+
+// Setup sets up the initial test cluster for the Windows
```
sebsoto: This comment is confusing as we arent setting up a cluster here

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: No, when I was doing rebase, I accidentally messed up the newlines. I am reverting those changes. Thanks for catching it

sebsoto:
/lgtm
/hold cancel
Nice work @ravisantoshgudimetla !

PR #109, opened 2019-12-10
[wsu] Transfer cni file to the Windows node created

ravisantoshgudimetla:
A WIP PR to test my changes. I'd prefer to get #103 merged and move the tests over there.

### DIFF
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/masterzen/winrm"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/cloudprovider"
 	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/types"
 	"golang.org/x/crypto/ssh"
@@ -40,6 +41,8 @@ type TestFramework struct {
 	cloudProvider cloudprovider.Cloud
 	// k8sclientset is the kubernetes clientset we will use to query the cluster's status
 	K8sclientset *kubernetes.Clientset
+	// OSConfigClient is the OpenShift config client, we will use to query the OpenShift api object status
+	OSConfigClient *configv1client.Clientset
 }
 
 // Setup sets up the Windows node so that it can join the existing OpenShift cluster
@@ -69,6 +72,9 @@ func (f *TestFramework) Setup() {
 	if err := f.getKubeClient(); err != nil {
 		log.Fatalf("failed to get kube client with error: %v", err)
 	}
+	if err := f.getOpenShiftConfigClient(); err != nil {
+		log.Fatalf("failed to get kube client with error: %v", err)
+	}
 }
 
 // getKubeClient setups the kubeclient that can be used across all the test suites.
@@ -83,6 +89,22 @@ func (f *TestFramework) getKubeClient() error {
 		return fmt.Errorf("could not create k8s clientset: %v", err)
 	}
 	f.K8sclientset = clientset
+	return nil
+}
+
+// getOpenShiftConfigClient gets the new OpenShift config v1 client
+func (f *TestFramework) getOpenShiftConfigClient() error {
+	kubeconfig := os.Getenv("KUBECONFIG")
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return fmt.Errorf("could not build config from flags: %v", err)
+	}
+	// Get openshift api config client.
+	configClient, err := configv1client.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("could not create config clientset: %v", err)
+	}
+	f.OSConfigClient = configClient
 	return nil
 }
 
--- a/internal/test/wsu/wsu_test.go
+++ b/internal/test/wsu/wsu_test.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
@@ -63,6 +64,9 @@ var (
 	retryCount = 20
 	// retryInterval is the interval of time until we retry after a failure
 	retryInterval = 5 * time.Second
+	// remotePowerShellCmdPrefix holds the powershell prefix that needs to be prefixed to every command run on the
+	// remote powershell session opened
+	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
 )
 
 // createhostFile creates an ansible host file and returns the path of it
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
 	t.Run("Network annotations were applied to node", testHybridOverlayAnnotations)
 	t.Run("HNS Networks were created", testHNSNetworksCreated)
 	t.Run("East-west networking", testEastWestNetworking)
+	t.Run("Check cni config generated on the Windows host", testCNIConfig)
+}
+
+// testCNIConfig tests if the CNI config has required hostsubnet and servicenetwork CIDR
+// TODO: split this into multiple tests when this grows
+func testCNIConfig(t *testing.T) {
+	// Check if the host subnet matches
+	windowsNodes, err := k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: windowsLabel})
+	require.NoError(t, err, "Could not get list of Windows nodes")
+	assert.Equalf(t, len(windowsNodes.Items), 1, "expected one windows node to be present but found %v",
+		len(windowsNodes.Items))
+	// By the time, we reach here the annotation should be present, so need to validate again
+	hostSubnet := windowsNodes.Items[0].Annotations[hybridOverlaySubnet]
+	cniConfigFilePath := filepath.Join(ansibleTempDir, "cni", "config", "cni.conf")
+	hostSubnetExists, err := doesFileContainMatchingString(cniConfigFilePath, hostSubnet)
+	require.NoError(t, err)
+	assert.Equal(t, hostSubnetExists, true)
+
+	// check if the service CIDR matches
+	networkCR, err := framework.OSConfigClient.ConfigV1().Networks().Get("cluster", metav1.GetOptions{})
+	assert.NoError(t, err, "error querying network object")
+	serviceNetworks := networkCR.Spec.ServiceNetwork
+	// The serviceNetwork should be a singleton slice as of now, let's try accessing the first element in it.
+	require.Equalf(t, len(serviceNetworks), 1, "expected service network to be a singleton but got %v",
+		len(serviceNetworks))
+	requiredServiceNetwork := serviceNetworks[0]
+	requiredServiceNetworkExists, err := doesFileContainMatchingString(cniConfigFilePath, requiredServiceNetwork)
+	require.NoError(t, err)
+	assert.Equal(t, requiredServiceNetworkExists, true)
 }
 
 // testFilesCopied tests that the files we attempted to copy to the Windows host, exist on the Windows host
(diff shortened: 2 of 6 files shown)


ravisantoshgudimetla:
/hold cancel
/approve cancel

This is up for review now barring commit message.

/cc @openshift/openshift-team-windows-containers

tools/ansible/tasks/wsu/templates/cni.j2
```diff
@@ -0,0 +1,32 @@
...
+                    "ExceptionList":[
+                        "10.132.0.0/16"
+                    ]
+                }
+            },
+            {
+                "name":"EndpointPolicy",
+                "value":{
+                    "Type":"ROUTE",
+                    "DestinationPrefix":"10.132.0.0/16",
```
sebsoto: This cant be hardcoded: "The destination prefix would come from the cluster network config objects service CIDR"

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -64,7 +64,8 @@ var (
 	// hybridOverlaySubnet is an annotation applied by the cluster network operator which is used by the hybrid overlay
 	hybridOverlaySubnet = "k8s.ovn.org/hybrid-overlay-hostsubnet"
 	// hybridOverlayMac is an annotation applied by the hybrid overlay
-	hybridOverlayMac = "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
+	hybridOverlayMac          = "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
+	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
```
sebsoto: Please add a comment stating what this does and why its necessary

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -275,6 +276,22 @@ func testHybridOverlayAnnotations(t *testing.T) {
...
+	hostSubnet := windowsNodes.Items[0].Annotations[hybridOverlaySubnet]
+
+	// Read the file from remote host and check if the
+	stdout := os.Stdout
+	r, w, err := os.Pipe()
+	assert.NoError(t, err, "error opening pipe to read stdout")
+	os.Stdout = w
+	_, err = winRMClient.Run(remotePowerShellCmdPrefix+"cat "+ansibleTempDir+"\\"+
+		"cni.conf",
+		os.Stdout, os.Stderr)
```
sebsoto: nit: why put this on a new line?

tools/ansible/tasks/wsu/templates/cni.j2
ravisantoshgudimetla: Hmm, is it, I thought we just need to update `subnet` as per our [doc](https://docs.google.com/document/d/1zAidTs8wbWHzamh4G3pwdaPomyT-doSbo9xTAcIK9no/edit#)

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
ravisantoshgudimetla: All these variables can be re-used from the framework once #103 merges, that's why I didn't comment

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -151,3 +151,16 @@
...
       delay: 5
+
+    # Get the subnet associated with host. We assume the network operator object has been modified to
+    # include hybrid overlay config
+    - name: Get the subnet associated with host
+      delegate_to: localhost
+      shell: "oc get nodes {{ node_name.stdout}} -o=jsonpath='{.metadata.annotations.k8s\\.ovn\\.org\\/hybrid-overlay-hostsubnet}'"
+      register: ovn_host_subnet
+
+    # Generate the cni.j2 file to be transferred to the Windows host
```
aravindhp: `cni.j2`?

tools/ansible/tasks/wsu/templates/cni.j2
```diff
@@ -0,0 +1,32 @@
...
+        "type": "host-local",
+        "subnet": "{{ ovn_host_subnet.stdout }}"
+    },
+        "policies":[
+            {
+                "name":"EndpointPolicy",
+                "value":{
+                    "Type":"OutBoundNAT",
+                    "ExceptionList":[
+                        "10.132.0.0/16"
```
aravindhp: I don't think we can hardcode this either.

tools/ansible/tasks/wsu/test/e2e/wsu_test.go
```diff
@@ -275,6 +276,22 @@ func testHybridOverlayAnnotations(t *testing.T) {
...
+	assert.NoError(t, err, "error while printing output of cni.conf file to stdout")
```
aravindhp: Shouldn't this be a `require`? Not sure if you are going to get anything useful from the lines below if this fails.

tools/ansible/tasks/wsu/templates/cni.j2
ravisantoshgudimetla: What should be the value here?

aravindhp: Please see [WINC-130](https://jira.coreos.com/browse/WINC-130)

sebsoto:
needs rebase

tools/ansible/tasks/wsu/templates/cni.j2
```diff
@@ -0,0 +1,32 @@
+
```
sebsoto: Extra newline?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -186,3 +197,4 @@
       fail:
         msg: "CNI Configuration error"
       when: '"CNI configuration completed successfully" not in bootstrap_out.stderr'
+
```
sebsoto: Extra newline

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -163,19 +163,30 @@
...
+      delegate_to: localhost
+      shell: "oc get nodes {{ node_name.stdout}} -o=jsonpath='{.metadata.annotations.k8s\\.ovn\\.org\\/hybrid-overlay-hostsubnet}'"
+      register: ovn_host_subnet
+
+    # Get the service CIDR associated with the OpenShift network operator object. We assume that network always object
+    # has atleast one entry as per defaulting at
+    # https://github.com/openshift/installer/blob/master/pkg/types/defaults/installconfig.go#L18
+    - name: Get the service CIDR associated with network object
+      delegate_to: localhost
+      shell: "oc get network.operator.openshift.io/cluster -o=jsonpath='{.spec.serviceNetwork[0]}'"
```
sebsoto: Do we always want to use the first element in the array? What are the other elements?

internal/test/wsu/wsu_test.go
```diff
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
 	t.Run("Network annotations were applied to node", testHybridOverlayAnnotations)
 	t.Run("HNS Networks were created", testHNSNetworksCreated)
 	t.Run("East-west networking", testEastWestNetworking)
+	t.Run("Check cni config generated on the Windows host", testCNIConfig)
```
sebsoto: nit: This test should be done before the `East-west networking` test, as the tests are currently being run in a order where the tested elements build on top of the previous test.

internal/test/wsu/wsu_test.go
```diff
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
...
+}
+
+// testCNIConfig tests if the CNI config has required hostsubnet and servicenetwork CIDR
+// TODO: split this into multiple tests when this grows
```
sebsoto: Please create a ticket for this TODO, or track it in an existing ticket :)

internal/test/wsu/wsu_test.go
```diff
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
...
+	// By the time, we reach here the annotation should be present, so need to validate again
+	hostSubnet := windowsNodes.Items[0].Annotations[hybridOverlaySubnet]
+	cniConfigFilePath := filepath.Join(ansibleTempDir, "cni", "config", "cni.conf")
+	hostSubnetExists, err := doesFileContainMatchingString(cniConfigFilePath, hostSubnet)
+	require.NoError(t, err)
+	assert.Equal(t, hostSubnetExists, true)
+
+	// check if the service CIDR matches
+	networkCR, err := framework.OSConfigClient.ConfigV1().Networks().Get("cluster", metav1.GetOptions{})
+	assert.NoError(t, err, "error querying network object")
```
sebsoto: This should be a require

internal/test/wsu/wsu_test.go
```diff
@@ -205,6 +238,27 @@ func testWorkerLabelsArePresent(t *testing.T) {
 		"expected worker label to be present on the Windows node but did not find any")
 }
 
+// doesFileContain checks if the given file on the remote Windows virtual machine contains given string
+func doesFileContainMatchingString(fileName, matchString string) (bool, error) {
+	// Read the file from remote host and check if the hostSubnet is present
+	stdout := os.Stdout
```
sebsoto: Why are you using os.Stdout here and not a new buffer like we do elsewhere in the test suite? See: https://github.com/openshift/windows-machine-config-operator/blob/master/internal/test/wsu/wsu_test.go#L135

This way you wont have to deal with all of the os.Pipe(), ReadAll(), etc.

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: We can have multiple CIDRs, as of now it is limited to 1. Have a look at https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_network.crd.yaml#L136

internal/test/framework/framework.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/masterzen/winrm"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
```
aravindhp: `configclient`

internal/test/wsu/wsu_test.go
aravindhp: This is not a TODO but a NOTE as it is an indicator for the next person changing this to split it up if adding new tests.

internal/test/wsu/wsu_test.go
```diff
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
...
+	windowsNodes, err := k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: windowsLabel})
+	require.NoError(t, err, "Could not get list of Windows nodes")
+	assert.Equalf(t, len(windowsNodes.Items), 1, "expected one windows node to be present but found %v",
+		len(windowsNodes.Items))
+	// By the time, we reach here the annotation should be present, so need to validate again
+	hostSubnet := windowsNodes.Items[0].Annotations[hybridOverlaySubnet]
+	cniConfigFilePath := filepath.Join(ansibleTempDir, "cni", "config", "cni.conf")
+	hostSubnetExists, err := doesFileContainMatchingString(cniConfigFilePath, hostSubnet)
+	require.NoError(t, err)
+	assert.Equal(t, hostSubnetExists, true)
```
aravindhp: [`assert.True`](https://godoc.org/github.com/stretchr/testify/assert#True)

internal/test/wsu/wsu_test.go
```diff
@@ -111,6 +115,35 @@ func TestWSU(t *testing.T) {
...
+
+	// check if the service CIDR matches
+	networkCR, err := framework.OSConfigClient.ConfigV1().Networks().Get("cluster", metav1.GetOptions{})
+	assert.NoError(t, err, "error querying network object")
+	serviceNetworks := networkCR.Spec.ServiceNetwork
+	// The serviceNetwork should be a singleton slice as of now, let's try accessing the first element in it.
+	require.Equalf(t, len(serviceNetworks), 1, "expected service network to be a singleton but got %v",
+		len(serviceNetworks))
+	requiredServiceNetwork := serviceNetworks[0]
+	requiredServiceNetworkExists, err := doesFileContainMatchingString(cniConfigFilePath, requiredServiceNetwork)
```
aravindhp: It would be nice if the file is read just once and the contents passed in here

internal/test/wsu/wsu_test.go
```diff
@@ -205,6 +238,27 @@ func testWorkerLabelsArePresent(t *testing.T) {
 		"expected worker label to be present on the Windows node but did not find any")
 }
 
+// doesFileContain checks if the given file on the remote Windows virtual machine contains given string
+func doesFileContainMatchingString(fileName, matchString string) (bool, error) {
```
aravindhp: It would be nice if the file is read just once and not every time this function is called

tools/ansible/tasks/wsu/main.yaml
aravindhp: I think so at least for now. There could be a more complex network setup that has multiple service networks but lets cross that bridge later. Make comment about that here.

aravindhp:
Continued in #114

PR #111, opened 2019-12-10
[wni][azure] Fix rule handling

aravindhp:
**Problem**:
WNI when creating VMs were overwriting the rules present in the  worker NSG causing due to which we were losing access to the OpenShift  console as the existing rules are required for console access.

**Solution:** 
Rather than updating the security group, the code has been  modified to just create / update the rules. To achieve this, the following was done:
- Remove all instances of updating the worker security group
- Introduce a struct that encapsulates an Azure NSG security rule from a WNI perspective
- Add methods to the struct that create/update and delete the rules
- GetMyIP() was exported to allow it be called from the e2e test code

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
@@ -3,8 +3,8 @@ package azure
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
-	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/types"
 	"io/ioutil"
 	"math/rand"
 	"net/http"
@@ -21,6 +21,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/client"
 	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/resource"
+	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/types"
 	logger "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
@@ -29,12 +30,24 @@ const (
 	sourcePortRange = "*"
 	// the '*' is used to match all the source IP addresses
 	destinationAddressPrefix = "*"
-	// winRM HTTPS port
-	winRM = "5986"
-	// includes the priority of opening winRm port in security group rules.
-	winRMPortPriority = 300
-	// winRM security group rule name
-	winRMRule = "WINRM"
+	// winRMPort is the secure WinRM port
+	winRMPort = "5986"
+	// winRMPortPriority is the priority for the WinRM rule
+	winRMPortPriority = 600
+	// winRMRuleName security group rule name for the WinRM rule
+	winRMRuleName = "WinRM"
+	// rdpPort is the RDP port
+	rdpPort = "3389"
+	// rdpRulePriority is the priority for the RDP rule
+	rdpRulePriority = 601
+	// rdpRuleName is the security group rule name for the RDP rule
+	rdpRuleName = "RDP"
+	// vnetPorts is the port range for vnet rule
+	vnetPorts = "1-65535"
+	// vnetRulePriority is the priority for the vnet traffic rule
+	vnetRulePriority = 602
+	// vnetRuleName is the security group rule name for vnet traffic within the cluster
+	vnetRuleName = "vnet_traffic"
 )
 
 var log = logger.Log.WithName("azure-vm")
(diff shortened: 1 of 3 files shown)


tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -13,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/windows-machine-config-operator/tools/windows-node-installer/pkg/types"
+
```
ravisantoshgudimetla: nit: Remove white space

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -329,84 +379,76 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
...
-		return fmt.Errorf("cannot get location of the openshift cluster: %v", err)
+// ensureSecurityGroupRules ensures that the required security group rules are present in the given NSG
+func (az *AzureProvider) ensureSecurityGroupRules(ctx context.Context, nsgRules []network.SecurityRule) error {
+
+	// Collect the existing rules
+	for _, nsgRule := range nsgRules {
+		fmt.Printf("rule %s\n", *nsgRule.Name)
+		switch *nsgRule.Name {
+		case rdpRuleName:
+			fmt.Println("found RDP rule")
```
ravisantoshgudimetla: nit: Please remove these println's

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -77,6 +90,27 @@ type AzureProvider struct {
 	instanceType string
 	// workspace to store all the results.
 	resourceTrackerDir string
+	// rdp
+	rdp *nsgRuleWrapper
```
ravisantoshgudimetla: While I am fine being specific, do you want to create a slice of `*nsgRuleWrapper`? just to make extension easier because I don't see any validation specific to any protocol like `rdp`, `winrm` etc?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -847,45 +885,16 @@ func (az *AzureProvider) deleteSpecificRule(s []network.SecurityRule, name strin
...
-		return fmt.Errorf("cannot delete the security group rules of the worker subnet: %v", err)
-	}
-	err = future.WaitForCompletionRef(ctx, az.nsgClient.Client)
-	if errorCheck(err) {
-		return fmt.Errorf("cannot delete the security group rules of the worker subnet: %v", err)
-	}
-	_, err = future.Result(az.nsgClient)
+
+	err = az.rdp.delete(ctx, nsgName)
+	err = az.winRM.delete(ctx, nsgName)
```
ravisantoshgudimetla: I think you already mentioned about covering validation errors.

tools/windows-node-installer/pkg/cloudprovider/azure/vm_test.go
```diff
@@ -0,0 +1,32 @@
...
+	getFileSettings, err := auth.GetSettingsFromFile()
+	if err != nil {
+		return "", err
+	}
+	subscriptionID = getFileSettings.GetSubscriptionID()
+	return
+}
+
+func TestConstructNetworkProfile(t *testing.T) {
+	oc, err := client.GetOpenShift("/home/aravindh/devel/installer/cluster0/auth/kubeconfig")
```
ravisantoshgudimetla: I hope you'll remove these as well :)

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -622,24 +664,20 @@ func (az *AzureProvider) constructNetworkProfile(ctx context.Context,
...
-		nsgName = az.generateResourceName("nsg", vmRandString)
-		err := az.createSecurityGroupRules(ctx, nsgName)
-		if errorCheck(err) {
-			return nil, fmt.Errorf("failed to update security group rules: %v", err)
-		}
+	if checkForNil(nsgStruct.ID) {
+		return nil, fmt.Errorf("failed to find worker NSG")
 	}
+
+	vmRandString := strings.Split(vmName, "-")[1]
```
sebsoto: nit: it took me a while to figure out the purpose of vmRandString, `vmRandomID` or something similar could be more intuitive.

aravindhp: This is part of existing code and renaming it would require changing parts of code that has nothing to do with this change.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -77,6 +90,25 @@ type AzureProvider struct {
 	instanceType string
 	// workspace to store all the results.
 	resourceTrackerDir string
+	// requiredRules is the set of SG rules that need to be created or deleted
+	requiredRules map[string]*nsgRuleWrapper
```
aravindhp: >While I am fine being specific, do you want to create a slice of *nsgRuleWrapper? just to make extension easier because I don't see any validation specific to any protocol like rdp, winrm etc?

Making it a slice would require looping through it when I am assigning existing `SecurityRule` to the wrapper. So instead I made it a map. Hope that satisfies your request.

ravisantoshgudimetla: SGTM

ravisantoshgudimetla:
For letting others to comment.

aravindhp:
/hold
for @sebsoto and @suhanime reviews

ravisantoshgudimetla:
Whoever is reviewing next, please remove the hold

sebsoto:
Thanks for addressing my comments
/hold cancel

PR #134, opened 2020-01-20
[wni][aws] Fix subnet selection

aravindhp:
**Problem**
To launch the instance, the first public subnet was being picked. In the case where the instance type is not supported in the availability zone the subnet is in, the instance creation would fail.

**Solution**
Rather picking the first public subnet, we further cross reference with the list of instance offerings and ensure we pick the subnet that is in a zone that supports the instance type.

Jira: [WINC-149](https://issues.redhat.com/browse/WINC-149)

### DIFF
--- a/tools/windows-node-installer/README.md
+++ b/tools/windows-node-installer/README.md
@@ -41,9 +41,12 @@ Available Commands:
 
 ```bash
 ./wni aws create --kubeconfig <path to OpenShift cluster>/kubeconfig --credentials <path to aws>/credentials 
---credential-account default --image-id ami-06a4e829b8bbad61e --instance-type m4.large --ssh-key <name of the 
+--credential-account default --image-id ami-06a4e829b8bbad61e --instance-type m5a.large --ssh-key <name of the 
 existing ssh keypair in aws> --private-key <private key to decrypt the aws instance password.>
 ```
+
+*Note*: Due to a bug in the Intel 82599 network adapter used in most Intel based instances that causes issues with
+overlay networks, we suggest using AMD based instances like `m5a.large`
 
 The default properties of the created instance are:
  - Instance name <OpenShift cluster\'s infrastructure ID>-windows-worker-\<zone\>-<random 4 characters string>
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -725,16 +725,54 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
 		return "", err
 	}
 
+	// Get the instance offerings that support Windows instances
+	scope := "Availability Zone"
+	productDescription := "Windows"
+	f := false
+	offerings, err := a.EC2.DescribeReservedInstancesOfferings(&ec2.DescribeReservedInstancesOfferingsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("scope"),
+				Values: []*string{&scope},
+			},
+		},
+		IncludeMarketplace: &f,
+		InstanceType:       &a.instanceType,
+		ProductDescription: &productDescription,
+	})
+	if err != nil {
+		return "", fmt.Errorf("error checking instance offerings of %s: %v", a.instanceType, err)
+	}
+	if offerings.ReservedInstancesOfferings == nil {
+		return "", fmt.Errorf("no instance offerings returned for %s", a.instanceType)
+	}
+
 	// Finding public subnet within the vpc.
+	foundPublicSubnet := false
 	for _, subnet := range subnets.Subnets {
 		for _, tag := range subnet.Tags {
 			// TODO: find public subnet by checking igw gateway in routing.
 			if *tag.Key == "Name" && strings.Contains(*tag.Value, infraID+"-public-") {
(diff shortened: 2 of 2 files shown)


tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -725,16 +725,54 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
 		return "", err
 	}
 
+	// Get the instance offerings that support Windows instances
+	scope := "Availability Zone"
+	productDescription := "Windows"
+	f := false
```
suhanime: Can we make this variable more meaningful? Maybe by adding a description or a meaningful name

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -725,16 +725,54 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
...
+					if *instanceOffering.AvailabilityZone == *subnet.AvailabilityZone {
+						return *subnet.SubnetId, nil
+					}
+				}
 			}
 		}
 	}
-	return "", fmt.Errorf("could not find a public subnet in VPC: %v", *vpc.VpcId)
+
+	errMsg := fmt.Sprintf("could not find a public subnet in VPC: %v", *vpc.VpcId)
```
sebsoto: why not use fmt.Errorf here?

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -725,16 +725,54 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
...
+	if !foundPublicSubnet {
+		errMsg = fmt.Sprintf("could not find a public subnet in a zone that supports %s instance type",
```
sebsoto: Same as above. You wouldnt have to cast it as an error in the return statement.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -725,16 +725,54 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
...
+	productDescription := "Windows"
+	f := false
+	offerings, err := a.EC2.DescribeReservedInstancesOfferings(&ec2.DescribeReservedInstancesOfferingsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("scope"),
+				Values: []*string{&scope},
+			},
+		},
+		IncludeMarketplace: &f,
```
suhanime: Since we are not using `f` anywhere else, can we just provide this value as false, with reasoning?

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: PS: I am not casting but passing it as an arg to `Errorf`

sebsoto:
/lgtm
/hold
holding for @suhanime

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
vinaykns: we can use to.BoolPtr(false) or for string to.StringPtr("example"), but internally it also uses the same approach i.e sending the address of the variable.

aravindhp: I am happy to do that

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/pem"
 	"fmt"
+	"github.com/Azure/go-autorest/autorest/to"
```
sebsoto: it feels a little weird importing an azure library into the aws wni. But oh well

tools/windows-node-installer/README.md
```diff
@@ -41,10 +41,13 @@ Available Commands:
 
 ```bash
 ./wni aws create --kubeconfig <path to OpenShift cluster>/kubeconfig --credentials <path to aws>/credentials 
---credential-account default --image-id ami-06a4e829b8bbad61e --instance-type m4.large --ssh-key <name of the 
+--credential-account default --image-id ami-06a4e829b8bbad61e --instance-type m5a.large --ssh-key <name of the
```
vinaykns: just a final question, should we also update the instance-type in the aws e2e test file.? https://github.com/openshift/windows-machine-config-operator/blob/master/tools/windows-node-installer/test/e2e/aws_test.go#L28

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: Thanks for catching this. I did not realize this. I will revert the change.

tools/windows-node-installer/README.md
aravindhp: The e2e test for WNI does not involve adding it to a cluster so it is fine to leave it as is.

PR #141, opened 2020-02-07
[ci] Save log files from output run

ravisantoshgudimetla:
As of now, we are not retrieving the log files for
kubelet and wsu after wsu, wmcb are run. This commit
should address the issue.

- Added a basic implementation of retrieve files in Windows VM
- Updated TestMain to call retrieve files in WSU, WMCB
- Add functionality to collect wsu output

As of now, we're not collecting log files
related to wsu runs. This commit should
address this issue.

Co-authored-by: Sebastian Soto <ssoto@redhat.com>

### DIFF
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -3,8 +3,10 @@ package framework
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -28,6 +30,9 @@ const (
 
 	// awsUsername is the default windows username on AWS
 	awsUsername = "Administrator"
+	// remoteLogPath is the directory where all the log files related to components that we need are generated on the
+	// Windows VM
+	remoteLogPath = "C:\\k\\log\\"
 )
 
 var (
@@ -292,6 +297,76 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
 		return nil, fmt.Errorf("could not find node with IP: %s", externalIP)
 	}
 	return matchedNode, nil
+}
+
+// WriteToArtifactDir will write contents to $ARTIFACT_DIR/subDirName/filename. If subDirName is empty, contents
+// will be written to $ARTIFACT_DIR/filename
+func (f *TestFramework) WriteToArtifactDir(contents []byte, subDirName, filename string) error {
+	path := filepath.Join(artifactDir, subDirName, filename)
+	dir, _ := filepath.Split(path)
+	err := os.MkdirAll(dir, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("could not create %s: %s", dir, err)
+	}
+	return ioutil.WriteFile(path, contents, os.ModePerm)
+}
+
+// GetNode uses external IP and finds out the name associated with the node
+func (f *TestFramework) GetNodeName(externalIP string) (string, error) {
+	node, err := f.GetNode(externalIP)
+	if err != nil {
+		return "", fmt.Errorf("error while getting required kubernetes node object: %v", err)
+	}
+	return node.Name, nil
+}
+
+// RetrieveArtifacts should retrieve artifacts related the test run. Ideally this should retrieve all the logs related
+// to the Windows VM. This shouldn't return an error but should print the failures as log collection is nice to have
+// rather than a must have.
+// TODO: Think about how we can retrieve stdout from ansible out within this function
+func (f *TestFramework) RetrieveArtifacts() {
+	for i, vm := range f.WinVMs {
+		if vm == nil {
+			continue
+		}
+		if vm.GetCredentials() == nil {
+			log.Printf("no credentials provided for vm %d ", i)
+			continue
+		}
+
+		instanceID := vm.GetCredentials().GetInstanceId()
+		if len(instanceID) == 0 {
+			log.Printf("no instance id provided for vm %d", i)
+			continue
+		}
+
+		externalIP := vm.GetCredentials().GetIPAddress()
+		if len(externalIP) == 0 {
+			log.Printf("no external ip address found for the vm with instance ID %s", instanceID)
+			continue
+		}
+
+		nodeName, err := f.GetNodeName(externalIP)
+		if err != nil {
+			log.Printf("error while getting node name associated with the vm %s: %v", instanceID, err)
+		}
+
+		// We want a format like "nodes/ip-10-0-141-99.ec2.internal/logs/wsu/kubelet"
+		localKubeletLogPath := filepath.Join(artifactDir, "nodes", nodeName, "logs")
+
+		// Let's reinitialize the ssh client as hybrid overlay is known to cause ssh connections to be dropped
+		// TODO: Reduce the usage of Reinitialize as much as possible, this is to ensure that when we move to operator
+		// 		model, the reconnectivity should be handled automatically.
+		if err := vm.Reinitialize(); err != nil {
+			log.Printf("failed re-initializing ssh connectivity with on vm %s: %v", instanceID, err)
+		}
+		// Get the VM's private ip and populate log files in the test container.
+		// Make this a map["'"artifact_that_we_want_to_pull"]="log_file.name"
+		if err := vm.RetrieveFiles(remoteLogPath, localKubeletLogPath); err != nil {
+			log.Printf("failed retrieving log files on vm %s: %v", instanceID, err)
+			continue
+		}
+	}
 }
 
 // TearDown destroys the resources created by the Setup function
--- a/internal/test/framework/windowsvm.go
+++ b/internal/test/framework/windowsvm.go
@@ -46,6 +46,9 @@ type WindowsVM interface {
 	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
 	// does not exist
 	CopyFile(string, string) error
+	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
+	// now, we're limiting every file in the remote directory to be written to single directory on the local host
+	RetrieveFiles(string, string) error
 	// Run executes the given command remotely on the Windows VM and returns the output of stdout and stderr. If the
 	// bool is set, it implies that the cmd is to be execute in PowerShell.
 	Run(string, bool) (string, string, error)
@@ -143,6 +146,74 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
 
 	// Forcefully close it so that we can execute the binary later
 	dstFile.Close()
+	return nil
+}
+
+// RetrieveFiles retrieves list of files from remote directory to the local directory.
+// The implementation can be changed if the use-case arises. As of now, we're doing a best effort
+// to collect every log possible. If a retrieval of file fails, we would proceed with retrieval
+// of other log files.
+func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
+	if w.sshClient == nil {
+		return fmt.Errorf("RetrieveFile cannot be called without a ssh client")
+	}
+
+	// Create local dir
+	err := os.MkdirAll(localDir, os.ModePerm)
+	if err != nil {
+		log.Printf("could not create %s: %s", localDir, err)
+	}
+
+	sftp, err := sftp.NewClient(w.sshClient)
+	if err != nil {
+		return fmt.Errorf("sftp initialization failed: %v", err)
+	}
+	defer sftp.Close()
+
+	// Get the list of all files in the directory
+	remoteFiles, err := sftp.ReadDir(remoteDir)
+	if err != nil {
+		return fmt.Errorf("error opening remote file: %v", err)
+	}
+
+	for _, remoteFile := range remoteFiles {
+		// Assumption: We ignore the directories here the reason being RetrieveFiles should just retrieve files
+		// in a directory, if this is directory, we should have called RetrieveFiles on this directory
+		if remoteFile.IsDir() {
+			continue
+		}
+		fileName := remoteFile.Name()
+		dstFile, err := os.Create(filepath.Join(localDir, fileName))
+		if err != nil {
+			log.Printf("error creating file locally: %v", err)
+			continue
+		}
+		// TODO: Check if there is some performance implication of multiple Open calls.
+		srcFile, err := sftp.Open(remoteDir + "\\" + fileName)
+
+		if err != nil {
+			log.Printf("error while opening remote directory on the Windows VM: %v", err)
+			continue
+		}
+		_, err = io.Copy(dstFile, srcFile)
+		if err != nil {
+			log.Printf("error retrieving file %v from Windows VM: %v", fileName, err)
+			continue
+		}
+		// flush memory
+		if err = dstFile.Sync(); err != nil {
+			log.Printf("error flusing memory: %v", err)
+			continue
+		}
+		if err := srcFile.Close(); err != nil {
+			log.Printf("error closing file on the remote host %s", fileName)
+			continue
+		}
+		if err := dstFile.Close(); err != nil {
+			log.Printf("error closing file %s locally", fileName)
+			continue
+		}
+	}
 	return nil
 }
 
--- a/internal/test/wsu/wsu_test.go
+++ b/internal/test/wsu/wsu_test.go
@@ -3,6 +3,7 @@ package wsu
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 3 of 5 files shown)


ravisantoshgudimetla:
Noticing the following error:

`2020/02/07 04:06:26 failed retrieving log files: sftp initialization failed: read tcp 172.16.190.86:42904->3.84.91.216:22: read: connection timed out on vm i-051c85a747518b359`

I may have to use @aravindhp's commit from here https://github.com/openshift/windows-machine-config-operator/pull/139/commits/e4970b1fa4076326a1a4a44a2bda30a2b334035c

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+			log.Printf("error while getting external ip address of the vm with instance ID %s", instanceID)
+			continue
+		}
+
+		nodeName := f.GetNodeName(externalIP)
+		if len(nodeName) == 0 {
+			log.Printf("error while getting internal ip address of the node object: %s", instanceID)
+		}
+
+		// We want a format like "ip-10-0-141-99.ec2.internal/nodes/logs/wsu/kubelet"
```
suhanime: nodes/ip-10-0-141-99.ec2.internal/logs...

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,63 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+	}
+
+	for _, remoteFile := range remoteFiles {
+		// Assumption: We ignore the directories here the reason being RetrieveFiles should just retrieve files
+		// in a directory, if this is directory, we should have called RetrieveFiles on this directory
+		if remoteFile.IsDir() {
+			continue
+		}
+		fileName := remoteFile.Name()
+		dstFile, err := os.Create(localDir + "/" + fileName)
```
suhanime: Can we use filepath instead for os.Create and sftp.Open below?

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
 	return matchedNode, nil
 }
 
+// GetNode uses external IP and finds out the name associated with the node
+func (f *TestFramework) GetNodeName(externalIP string) string {
```
sebsoto: Why not just return an error here, and then handle the error how you wish in the calling function? You end up checking if  `len(nodeName) == 0` anyway.

I'd also remove the log messages within the function itself and log when you error handle. Doing it this way limits the way this function can be used

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+// to the Windows VM. This shouldn't return an error but should print the failures as log collection is nice to have
+// rather than a must have.
+// TODO: Think about how we can retrieve stdout from ansible out within this function
+func (f *TestFramework) RetrieveArtifacts() {
+	for _, vm := range f.WinVMs {
+		if vm == nil {
+			continue
+		}
+		if vm.GetCredentials() == nil {
+			log.Print("error while getting credentials of the vm")
```
sebsoto: Please log which VM

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+			continue
+		}
+
+		instanceID := vm.GetCredentials().GetInstanceId()
+		if len(instanceID) == 0 {
+			log.Print("error while getting instance id of the vm")
```
sebsoto: Log which VM, even if it is just 'VM 1'/'VM 2'/etc

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+		nodeName := f.GetNodeName(externalIP)
+		if len(nodeName) == 0 {
+			log.Printf("error while getting internal ip address of the node object: %s", instanceID)
+		}
+
+		// We want a format like "ip-10-0-141-99.ec2.internal/nodes/logs/wsu/kubelet"
+		localKubeletLogPath := filepath.Join(artifactDir, "nodes", nodeName, "logs", "wsu", "kubelet")
+
+		// Let's reinitialize the ssh client as over hybrid overlay is known to cause ssh connections to be dropped
+		vm.Reinitialize()
```
sebsoto: Why is this being done here and not where the hybrid overlay is being setup?

sebsoto:
Commit message has a typo in it `update windows interface to retrievefiles ` -> `retrieve files` also please specify `windowsVM interface`

internal/test/framework/framework.go
ravisantoshgudimetla: Thanks for catching it.

ravisantoshgudimetla:
Sample output from previous run here:

https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/141/pull-ci-openshift-windows-machine-config-operator-master-e2e-wsu/619/artifacts/e2e-wsu/nodes/ip-10-0-21-228.ec2.internal/logs/wsu/

internal/test/framework/framework.go
```diff
@@ -294,6 +301,73 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
 	return matchedNode, nil
 }
 
+// GetNode uses external IP and finds out the name associated with the node
+func (f *TestFramework) GetNodeName(externalIP string) string {
+	node, err := f.GetNode(externalIP)
+	if err != nil {
+		log.Printf("error while getting required kubernetes node object: %v", err)
```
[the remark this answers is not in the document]
suhanime: Didn't check the calling function. My bad

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,63 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+// As of now, I think there is no need to have a recursive copy as the log directory should contain only log files
+//but no child directories. I am fine change the implementation if the use-case arises.
+func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
+	// Create local dir
+	err := os.MkdirAll(localDir, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("could not create %s: %s", localDir, err)
+	}
+
+	if w.sshClient == nil {
```
sebsoto: This should be done before creating the local dir

internal/test/wsu/wsu_test.go
```diff
@@ -98,32 +98,45 @@ func testAllVMs(t *testing.T) {
...
+	wsuOut, err := runWSU(vm)
+	wsuStringOutput := string(wsuOut)
+	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuStringOutput)
+	// TODO: Think of a better way to refactor this function later to get just output that can be consumed by framework.
+	logFileName := "wsu.log"
+	externalIP := vm.GetCredentials().GetIPAddress()
+	nodeName, err := framework.GetNodeName(externalIP)
+	if err != nil {
+		// TODO: Change all fmt.Printfs to log.Printfs
+		fmt.Printf("could not print the internal ip for the vm %s", vm.GetCredentials().GetInstanceId())
```
sebsoto: shouldn't this be `could not get the node name of the vm`

internal/test/wsu/wsu_test.go
```diff
@@ -98,32 +98,45 @@ func testAllVMs(t *testing.T) {
...
+	vm := framework.WinVMs[vmNum]
 	// Run the WSU against the VM
-	wsuOut, err := runWSU(framework.WinVMs[vmNum])
-	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuOut)
-
+	wsuOut, err := runWSU(vm)
+	wsuStringOutput := string(wsuOut)
+	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuStringOutput)
+	// TODO: Think of a better way to refactor this function later to get just output that can be consumed by framework.
+	logFileName := "wsu.log"
```
sebsoto: Won't this result in the first run getting overwritten? This function is run twice per VM. You need to differentiate between the two runs. You can use the test name for that.

internal/test/framework/framework.go
```diff
@@ -28,6 +31,9 @@ const (
 
 	// awsUsername is the default windows username on AWS
 	awsUsername = "Administrator"
+	// remoteKubeletLogPath is the directory where all the log files related to kubelet are generated on the remote Windows
+	// Machine
```
aravindhp: This is the directory where we are dropping all the logs not just kubelet, so please reflect that here.

internal/test/framework/framework.go
```diff
@@ -294,6 +300,62 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+		}
+
+		instanceID := vm.GetCredentials().GetInstanceId()
+		if len(instanceID) == 0 {
+			log.Printf("error while getting instance id of the %d vm", i)
+			continue
+		}
+
+		externalIP := vm.GetCredentials().GetIPAddress()
+		if len(externalIP) == 0 {
```
aravindhp: Given you are validating, please validate that is a proper IP address.

internal/test/framework/framework.go
```diff
@@ -294,6 +300,62 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+// to the Windows VM. This shouldn't return an error but should print the failures as log collection is nice to have
+// rather than a must have.
+// TODO: Think about how we can retrieve stdout from ansible out within this function
+func (f *TestFramework) RetrieveArtifacts() {
+	for i, vm := range f.WinVMs {
+		if vm == nil {
+			continue
+		}
+		if vm.GetCredentials() == nil {
+			log.Printf("error while getting credentials of the %d vm", i)
```
aravindhp: There is no err per se but that the credentials were empty. `no credentials provided for vm %d` would be more accurate. This applies to the other logs here.

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,63 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
 	return nil
 }
 
+// RetrieveFiles retrieves list of files from remote directory to the local directory.
+// As of now, I think there is no need to have a recursive copy as the log directory should contain only log files
```
aravindhp: Remove `I think`

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,63 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+//but no child directories. I am fine change the implementation if the use-case arises.
```
aravindhp: Remove `I am fine` and make it a statement. `The implementation can be changed if the use-case arises`.

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,63 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+		if remoteFile.IsDir() {
+			continue
+		}
+		fileName := remoteFile.Name()
+		dstFile, err := os.Create(filepath.Join(localDir, fileName))
+		if err != nil {
+			return fmt.Errorf("error creating file locally: %v", err)
+		}
+		// TODO: Check if there is some performance implication of multiple Open calls.
+		srcFile, err := sftp.Open(remoteDir + "\\" + fileName)
```
aravindhp: I don't know about performance implications but you should close them after the copy is complete.

internal/test/framework/framework.go
```diff
@@ -294,6 +300,62 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+			continue
+		}
+
+		nodeName, err := f.GetNodeName(externalIP)
+		if err != nil {
+			log.Printf("error while getting node name associated with the vm %s: %v", instanceID, err)
+		}
+
+		// We want a format like "nodes/ip-10-0-141-99.ec2.internal/logs/wsu/kubelet"
+		localKubeletLogPath := filepath.Join(artifactDir, "nodes", nodeName, "logs", "wsu", "kubelet")
```
aravindhp: What about the `wmcb` case? You cannot hard code this to `wsu`. More over there is going to be more than just the kubelet log here, so I suggest leaving out the `kubelet` directory.

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: You're right. I did not think of that case. I'll update the log file's name to include the test name

internal/test/framework/framework.go
ravisantoshgudimetla: Ok, I'll remove it.

internal/test/framework/framework.go
ravisantoshgudimetla: Good Point. Will change it.

internal/test/framework/windowsvm.go
ravisantoshgudimetla: Good point. I missed it. Thanks for catching it.

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: Update: There is a bug in the current code where we're not running tests twice on each node. @sebsoto will handle it in a separate PR. I'll add the test name.

internal/test/wsu/wsu_test.go
```diff
@@ -98,32 +98,45 @@ func testAllVMs(t *testing.T) {
...
+	vm := framework.WinVMs[vmNum]
 	// Run the WSU against the VM
-	wsuOut, err := runWSU(framework.WinVMs[vmNum])
-	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuOut)
-
+	wsuOut, err := runWSU(vm)
+	wsuStringOutput := string(wsuOut)
+	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuStringOutput)
+	// TODO: Think of a better way to refactor this function later to get just output that can be consumed by framework.
+	logFileName := t.Name() + "wsu.log"
```
ravisantoshgudimetla: @sebsoto - Let me know, if this addresses your concern of running WSU multiple times.

internal/test/framework/windowsvm.go
ravisantoshgudimetla: Done

internal/test/wsu/wsu_test.go
sebsoto: Be aware that the test name contains foward slashes, which will cause sub directories to be created.

ravisantoshgudimetla: We should be good with that as long as we've VM name being generated for test. Look at the log file that got generated here:

https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/141/pull-ci-openshift-windows-machine-config-operator-master-e2e-wsu/628/artifacts/e2e-wsu/nodes/ip-10-0-19-99.ec2.internal/logs/TestWSU/VM_specific_tests/

internal/test/framework/framework.go
```diff
@@ -294,6 +299,76 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+
+		nodeName, err := f.GetNodeName(externalIP)
+		if err != nil {
+			log.Printf("error while getting node name associated with the vm %s: %v", instanceID, err)
+		}
+
+		// We want a format like "nodes/ip-10-0-141-99.ec2.internal/logs/wsu/kubelet"
+		localKubeletLogPath := filepath.Join(artifactDir, "nodes", nodeName, "logs")
+
+		// Let's reinitialize the ssh client as over hybrid overlay is known to cause ssh connections to be dropped
```
aravindhp: This comment is not phrased correctly

internal/test/framework/windowsvm.go
```diff
@@ -146,6 +149,68 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+		}
+		// flush memory
+		if err = dstFile.Sync(); err != nil {
+			return fmt.Errorf("error flusing memory: %v", err)
+		}
+		if err := srcFile.Close(); err != nil {
+			return fmt.Errorf("error closing file on the remote host %s", fileName)
+		}
+		if err := dstFile.Close(); err != nil {
+			return fmt.Errorf("error closing file %s locally", fileName)
```
aravindhp: Should we error on these scenarios? I suggest we try to collect as many logs as we can and just log the failures. More over there is no reason to not close `srcFile` just because there was an error in closing `dstFile`.

internal/test/wsu/wsu_test.go
```diff
@@ -98,32 +98,45 @@ func testAllVMs(t *testing.T) {
...
+	externalIP := vm.GetCredentials().GetIPAddress()
+	nodeName, err := framework.GetNodeName(externalIP)
+	if err != nil {
+		// TODO: Change all fmt.Printfs to log.Printfs
+		fmt.Printf("could not node name associated with the vm %s", vm.GetCredentials().GetInstanceId())
+	}
+	require.NoErrorf(t, err, "Error getting node associated with vm %s", vm.GetCredentials().GetInstanceId())
+	localLogDirLocation := filepath.Join("nodes", nodeName, "logs")
+	if err = framework.WriteToArtifactDir(wsuOut, localLogDirLocation, logFileName); err != nil {
+		fmt.Printf("could not write %s to artifact dir: %s", logFileName, err)
```
sebsoto: Need a `\n` at the end of the print

ravisantoshgudimetla: Done

internal/test/wsu/wsu_test.go
```diff
@@ -98,32 +98,45 @@ func testAllVMs(t *testing.T) {
...
-
+	wsuOut, err := runWSU(vm)
+	wsuStringOutput := string(wsuOut)
+	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuStringOutput)
+	// TODO: Think of a better way to refactor this function later to get just output that can be consumed by framework.
+	logFileName := t.Name() + "wsu.log"
+	externalIP := vm.GetCredentials().GetIPAddress()
+	nodeName, err := framework.GetNodeName(externalIP)
+	if err != nil {
+		// TODO: Change all fmt.Printfs to log.Printfs
```
aravindhp: This is the first case of logging in this file. Please use `log` for consistency.

internal/test/framework/framework.go
```diff
@@ -294,6 +299,76 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+// to the Windows VM. This shouldn't return an error but should print the failures as log collection is nice to have
+// rather than a must have.
+// TODO: Think about how we can retrieve stdout from ansible out within this function
+func (f *TestFramework) RetrieveArtifacts() {
+	for i, vm := range f.WinVMs {
+		if vm == nil {
+			continue
+		}
+		if vm.GetCredentials() == nil {
+			log.Printf("no credentials provided for vm %d vm", i)
```
sebsoto: `no credentials provided for vm %d vm` -> `no credentials provided for vm %d`

internal/test/framework/framework.go
```diff
@@ -294,6 +299,76 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+			continue
+		}
+
+		instanceID := vm.GetCredentials().GetInstanceId()
+		if len(instanceID) == 0 {
+			log.Printf("no instance id provided for %d vm", i)
```
sebsoto: `no instance id provided for %d vm` -> `no instance id provided for vm %d`

internal/test/framework/framework.go
```diff
@@ -294,6 +299,76 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+		// Let's reinitialize the ssh client as hybrid overlay is known to cause ssh connections to be dropped
+		// TODO: Reduce the usage of Reinitialize as much as possible, this is to ensure that when we move to operator
+		// 		model, the reconnectivity should be handled automatically.
+		if err := vm.Reinitialize(); err != nil {
+			log.Printf("failed re-initializing ssh connetivity with error: %v on vm %s", err, instanceID)
+		}
+		// Get the VM's private ip and populate log files in the test container.
+		// Make this a map["'"artifact_that_we_want_to_pull"]="log_file.name"
+		if err := vm.RetrieveFiles(remoteLogPath, localKubeletLogPath); err != nil {
+			log.Printf("failed retrieving log files: %v on vm %s", err, instanceID)
```
sebsoto: err should trail the rest of log statement:
` log.Printf("failed retrieving log files on vm %s: %v", instanceID, err)`

ravisantoshgudimetla:
/hold

Something wrong about test infra

aravindhp:
/hold cancel

CI and build issues have been resolved

PR #142, opened 2020-02-11
[wsu] Check if hybrid overlay is enabled

sebsoto:
This commit causes the WSU to fail early if either the cluster is not
ovn-kubernetes, or the hybrid overlay patch has not been applied to
the network.operator CR.


internal/test/framework/framework.go
```diff
@@ -294,6 +295,28 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+	return nil
+}
+
+// patchClusterNetworkOperator patches the "network.operator cluster" resource, with the provided JSON patch
+func (f *TestFramework) patchClusterNetworkOperator(patch []byte) error {
+	return f.OSConfigClient.RESTClient().Patch(k8stypes.MergePatchType).
+		AbsPath("/apis/operator.openshift.io/v1/networks/cluster").
+		Body(patch).
+		Do().
+		Error()
```
aravindhp: Did you ping #forum-sdn as to the best way to make this API call?

internal/test/framework/framework.go
```diff
@@ -294,6 +295,28 @@ func (f *TestFramework) GetNode(externalIP string) (*v1.Node, error) {
...
+	err := f.patchClusterNetworkOperator(jsonPatch)
+	if err != nil {
+		return fmt.Errorf("could not apply patch %s to cluster network: %s", string(jsonPatch), err)
+	}
+
+	return nil
+}
+
+// patchClusterNetworkOperator patches the "network.operator cluster" resource, with the provided JSON patch
+func (f *TestFramework) patchClusterNetworkOperator(patch []byte) error {
```
ravisantoshgudimetla: Should we use the same logic for `wmcb` and remove patching cno from shell script?

aravindhp: Sure but that should be done in a separate PR

ravisantoshgudimetla: Yeah, mention it somewhere in the code, so that we won't forget.

internal/test/wsu/wsu_test.go
```diff
@@ -76,14 +76,30 @@ ansible_winrm_server_cert_validation=ignore
...
+	require.NoError(t, err, "Could not apply hybrid overlay patch")
+	t.Run("VM specific tests", testAllVMs)
 
 	// Run cluster wide tests
 	t.Run("Pending CSRs were approved", testNoPendingCSRs)
 }
 
+// testWithoutHybridOverlay ensures that the WSU fails early when hybrid overlay is not enabled
+func testWithoutHybridOverlay(t *testing.T) {
+	for i := range framework.WinVMs {
```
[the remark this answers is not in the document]
ravisantoshgudimetla: nit: As discussed yesterday, why can't you just use values from the iteration?

internal/test/framework/framework.go
sebsoto: @ravisantoshgudimetla it makes more sense to open a ticket if we want this work done. I've created https://issues.redhat.com/browse/WINC-224, wasnt sure what to put in the description so please fill it out :smile:

internal/test/wsu/wsu_test.go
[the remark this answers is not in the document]
sebsoto: I will be removing the loop, which will resolve both these comments

internal/test/framework/framework.go
sebsoto: I will be switching to using the operator client library, this change is being blocked by #146. Will push revisions once that merges

internal/test/go.mod
```diff
@@ -3,9 +3,11 @@ module github.com/openshift/windows-machine-config-bootstrapper/internal/test
 go 1.12
 
 replace (
-	k8s.io/api => k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
-	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
-	k8s.io/client-go => k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200205145930-e9d93e317dd1 // OpenShift 4.3
```
aravindhp: Wasn't this done in #146 and #151?

internal/test/go.sum
```diff
@@ -1,81 +1,57 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
-github.com/Azure/azure-sdk-for-go v34.1.0+incompatible h1:uW/dgSzmRQEPXwaRUN8WzBHJy5J2cp8cw1ea908uFj0=
```
aravindhp: A bit confused by this. Why are we seeing a change in `go.sum`?

internal/test/framework/framework.go
```diff
@@ -369,6 +391,19 @@ func (f *TestFramework) RetrieveArtifacts() {
 	}
 }
 
+// ApplyHybridOverlayPatch will enable the hybrid overlay on the cluster
+func (f *TestFramework) ApplyHybridOverlayPatch() error {
+	jsonPatch := []byte(`{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"hybridOverlayConfig":` +
+		`{"hybridClusterNetwork":[{"cidr":"10.132.0.0/14","hostPrefix":23}]}}}}}`)
+
+	_, err := f.OSOperatorClient.Networks().Patch("cluster", k8stypes.MergePatchType, jsonPatch)
```
aravindhp: Is this patching the network CR or the network.operator config CR? How are they being differentiated?

internal/test/wsu/wsu_test.go
```diff
@@ -77,14 +77,26 @@ ansible_winrm_server_cert_validation=ignore
...
 
-	t.Run("VM specific tests", func(t *testing.T) {
-		testAllVMs(t)
-	})
+	// Run the WSU before applying hybrid overlay patch, expecting it to fail with a verbose error message
+	t.Run("Expect failure when hybrid overlay is not enabled", testWithoutHybridOverlay)
+
+	// Apply the hybrid overlay patch, and run the full VM test suite for each VM
+	err := framework.ApplyHybridOverlayPatch()
+	require.NoError(t, err, "Could not apply hybrid overlay patch")
```
aravindhp: Do we need a delay after this given that the patch takes a while to take effect? Ideally we wait until the OVN pods are ready again or the hybrid annotations are present on all master nodes.

internal/test/go.mod
sebsoto: This was done to the WNI's dependencies, not the framework's.

internal/test/go.sum
sebsoto: We're seeing a change in go.sum as the k8s and openshift libraries have been updated, changing the subdependencies they're bringing in. That change is reflected here in the go.sum

internal/test/framework/framework.go
sebsoto: This is patching the network.operator config CR. They are differentiated as the network CR is accessed through the OSConfigClient and the network.operator is accessed through the OSOperatorClient

internal/test/wsu/wsu_test.go
sebsoto: Yes, we should probably be more thorough about waiting for things to finish setting up. I will come up with something

internal/test/go.sum
aravindhp: So why isn't this going into the previous bump commit?

internal/test/framework/framework.go
```diff
@@ -194,6 +201,21 @@ func (f *TestFramework) getOpenShiftConfigClient() error {
 	return nil
 }
 
+// getOpenShiftConfigClient gets a new OpenShift operator v1 client
+func (f *TestFramework) getOpenShiftOperatorClient() error {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
ravisantoshgudimetla: Can we try reducing the usage of `kubeconfig` global var and see if we want that var in our framework. I am fine using the `KUBECONFIG` environment var for now.

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: I agree instead of introducing delay, it'd be better to go with the approach that @aravindhp suggested.

internal/test/framework/framework.go
```diff
@@ -164,14 +177,20 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
 	return nil
 }
 
-// getKubeClient setups the kubeclient that can be used across all the test suites.
-func (f *TestFramework) getKubeClient() error {
+// getKubeConfig sets the kubeconfig in the test framework, using the KUBECONFIG environment variable
+func (f *TestFramework) getKubeconfig() error {
```
aravindhp: Why are you introducing this change that introduces a subtle ordering requirement with the `get*()` functions that is not called out anywhere? I am not sure @ravisantoshgudimetla's [comment](https://github.com/openshift/windows-machine-config-bootstrapper/pull/142#discussion_r380317780) implied that you do this.

internal/test/framework/framework.go
```diff
@@ -164,14 +177,20 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
...
 	kubeconfig := os.Getenv("KUBECONFIG")
```
aravindhp: If at all you are going to make this change, please remove this redundant line.

internal/test/framework/framework.go
```diff
@@ -369,6 +395,100 @@ func (f *TestFramework) RetrieveArtifacts() {
...
+		time.Sleep(RetryInterval)
+	}
+	return fmt.Errorf("timed out waiting for pods in namespace \"openshift-ovn-kubernetes\" to be ready")
+
+}
+
+// waitUntilMasterNodesAnnotated returns when either all master nodes have had the hybrid overly gateway mac annotation
+// applied to them, or reaches a timeout limit.
+func (f *TestFramework) waitUntilMasterNodesAnnotated() error {
+	masterLabel := "node-role.kubernetes.io/master"
```
aravindhp: I misspoke about it being just master nodes. These annotations are applied on both `master` and `worker` nodes so they all need to be checked.

internal/test/framework/framework.go
```diff
@@ -369,6 +395,100 @@ func (f *TestFramework) RetrieveArtifacts() {
...
+	ovnAnnotation := "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
```
aravindhp: Shouldn't we also check for `k8s.ovn.org/hybrid-overlay-node-subnet`? And please use the global variables here.

internal/test/framework/framework.go
sebsoto: That is how I interpreted his comment, reading it again I think I misinterpreted it

internal/test/framework/framework.go
sebsoto: We can if you feel that it is necessary. I wanted to avoid using the other annotation as it is dependent on the openshift version. I see the ovn pods being ready is the real test of if things have fully set up.

aravindhp: >We can if you feel that it is necessary. I wanted to avoid using the other annotation as it is dependent on the openshift version.
We are checking for `k8s.ovn.org/hybrid-overlay-node-subnet` [elsewhere](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/internal/test/wsu/wsu_test.go#L303) in our e2e so we are already dependent on the OpenShift version.

>I see the ovn pods being ready is the real test of if things have fully set up.
I would say we need to check for the pods and annotations as it is unclear when those annotations get applied i.e. before the pods go to running or during.

internal/test/framework/framework.go
```diff
@@ -369,6 +395,100 @@ func (f *TestFramework) RetrieveArtifacts() {
...
+// waitUntilHybridOverlayReady returns once OVN is ready again, after applying the hybrid overlay patch.
+func (f *TestFramework) waitUntilHybridOverlayReady() error {
+	err := f.waitUntilMasterNodesAnnotated()
+	if err != nil {
+		return err
+	}
+
+	// This is being done after we wait for the master nodes, as we need to make sure the patch is being acted on,
+	// and we are not checking the pods before they begin to restart with the hybrid overlay changes
+	err = f.waitUntilOVNPodsReady()
```
ravisantoshgudimetla: nit: You can use `if err:=f.waitUntil..` pattern here.

internal/test/framework/framework.go
ravisantoshgudimetla: > I see the ovn pods being ready is the real test of if things have fully set up.

That may not guarantee you that the annotation has been applied what if patch failed or no-op and still it returned successful. Again nothing wrong in being defensive.

internal/test/framework/framework.go
```diff
@@ -369,6 +395,100 @@ func (f *TestFramework) RetrieveArtifacts() {
...
+		if len(masterNodes.Items) == 0 {
+			return fmt.Errorf("master node list returned 0 entries")
+		}
+
+		allAnnotated := true
+		for _, node := range masterNodes.Items {
+			_, ok := node.Annotations[ovnAnnotation]
+			if !ok {
+				allAnnotated = false
+				break
```
ravisantoshgudimetla: What's wrong with returning here?

sebsoto: We want to be able to retry, returning there doesn't allow it

internal/test/framework/framework.go
```diff
@@ -149,12 +155,19 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
 			return fmt.Errorf("unable to instantiate Windows VM: %v", err)
 		}
 	}
-	if err := f.getKubeClient(); err != nil {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
aravindhp: Have you confirmed that `config` can be reused across different clients?

internal/test/framework/framework.go
aravindhp: I would say this is not a nit and is a pattern we should follow where possible.

internal/test/framework/framework.go
sebsoto: I've looked at each of the clients and I dont see anything that suggests that it can't be used across them.
Each client makes a copy of the config and then uses the copy and not the original.

internal/test/framework/framework.go
```diff
@@ -369,6 +384,100 @@ func (f *TestFramework) RetrieveArtifacts() {
 	}
 }
 
+// ApplyHybridOverlayPatch will enable the hybrid overlay on the cluster
+func (f *TestFramework) ApplyHybridOverlayPatch() error {
```
ravisantoshgudimetla: nit: I am wondering if we want to specialize here with wsuFramework struct similar to what @suhanime is doing https://github.com/openshift/windows-machine-config-bootstrapper/pull/148. The main reason is ideally, we would want to take the `internal/framework` package and re-use it in our new wmco repo.

internal/test/framework/framework.go
ravisantoshgudimetla: Got it, I misread. Thank you :)

PR #148, opened 2020-02-12
[wsu] Autodownload binaries based on cluster version

suhanime:
This PR enables WSU to download the latest WMCB, hybrid-overlay and kube node binaries based on the cluster version

To achieve this, first commit introduces TestWindowsVM interface that composes WindowsVM interface. The second commit then introduces a property to TestWindowsVM interface that is used for e2e-wsu tests

Since CI cluster does not give `oc version` in the format of 4.x, kubernetes (server) version is used to map to the corresponding OpenShift version
WSU can now choose to build wmcb or download it based on a boolean argument
E2E SHA check now happens for wmcb (downloaded), hybrid-overlay and kube node binaries, as a part of the playbook

- e2e-wsu tests changed to bring up 2 VMs
- Add property `buildWMCB` to Windows VM interface
- Specialize framework setup for wsu_test.go
- Add extra guards in the playbook to fail early
 
[WINC-131](https://issues.redhat.com/browse/WINC-131)

### DIFF
--- a/internal/test/framework/windowsvm.go
+++ b/internal/test/framework/windowsvm.go
@@ -39,6 +39,9 @@ type windowsVM struct {
 	sshClient *ssh.Client
 	// winrmClient to access the Windows VM created
 	winrmClient *winrm.Client
+	// buildWMCB indicates if WSU should build WMCB and use it
+	// TODO This is a WSU specific property and should be moved to wsu_test -> https://issues.redhat.com/browse/WINC-249
+	buildWMCB bool
 }
 
 // WindowsVM is the interface for interacting with a Windows VM in the test framework
@@ -64,6 +67,11 @@ type WindowsVM interface {
 	Reinitialize() error
 	// Destroy destroys the Windows VM
 	Destroy() error
+	// BuildWMCB returns the value of buildWMCB. It can be used by WSU to decide if it should build WMCB before using it
+	BuildWMCB() bool
+	// SetBuildWMCB sets the value of buildWMCB. Setting buildWMCB to true would indicate WSU will build WMCB instead of
+	// downloading the latest as per the cluster version. False by default
+	SetBuildWMCB(bool)
 }
 
 // newWindowsVM creates and sets up a Windows VM in the cloud and returns the WindowsVM interface that can be used to
@@ -356,4 +364,12 @@ func (w *windowsVM) getSSHClient() error {
 	}
 	w.sshClient = sshClient
 	return nil
+}
+
+func (w *windowsVM) BuildWMCB() bool {
+	return w.buildWMCB
+}
+
+func (w *windowsVM) SetBuildWMCB(buildWMCB bool) {
+	w.buildWMCB = buildWMCB
 }
--- a/internal/test/wsu/main_test.go
+++ b/internal/test/wsu/main_test.go
@@ -8,11 +8,13 @@ import (
 	"testing"
 )
 
-// framework holds the instantiation of test suite being executed. As of now, temp dir is hardcoded.
 var (
-	framework = &e2ef.TestFramework{}
+	// Initialize wsuFramework which specializes TestFramework by adding some properties specific to WSU tests
+	framework = wsuFramework{}
 	// TODO: expose this to the end user as a command line flag
 	// vmCount is the number of VMs the test suite requires
+	// Bring up 2 vms to test wsu run with automatic download of WMCB based on cluster version as well as built from
+	// source version
 	vmCount = 2
 )
 
--- a/internal/test/wsu/wsu_test.go
+++ b/internal/test/wsu/wsu_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/openshift/windows-machine-config-bootstrapper/internal/test"
 	e2ef "github.com/openshift/windows-machine-config-bootstrapper/internal/test/framework"
+	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
@@ -28,12 +29,6 @@ import (
 var (
 	// Path of the WSU playbook
 	playbookPath = os.Getenv("WSU_PATH")
-
-	// kubernetes-node-windows-amd64.tar.gz SHA512
-	// Value from https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#node-binaries-1
-	// This value should be updated when we change the kubelet version in WSU
-	expectedKubeTarSha = "a88e7a1c6f72ea6073dbb4ddfe2e7c8bd37c9a56d94a33823f531e303a9915e7a844ac5880097724e44dfa7f4" +
-		"a9659d14b79cc46e2067f6b13e6df3f3f1b0f64"
 	// workerLabel is the worker label that needs to be applied to the Windows node
 	workerLabel = "node-role.kubernetes.io/worker"
 	// hybridOverlayMac is an annotation applied by the hybrid overlay
@@ -43,6 +38,43 @@ var (
 	// ubi8Image is the name/location of the linux image we will use for testing
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
+
+type wsuFramework struct {
+	// TestFramework holds the instantiation of test suite being executed
+	*e2ef.TestFramework
+	// Number of VMs which should use built version of WMCB
+	// TODO Expose this option to the user along with vmCount -> https://issues.redhat.com/browse/WINC-240
+	vmCountWithBuiltWMCB int
+}
+
+// Setup initializes the wsuFramework.
+func (f *wsuFramework) Setup(vmCount int, credentials []*types.Credentials, skipVMsetup bool) error {
+	f.TestFramework = &e2ef.TestFramework{}
+
+	// If vmCount is 3 and vmCountWithBuiltWMCB is 2, 2 VMs will run WSU that will build WMCB and 1 VM will
+	// auto-download the latest WMCB based on the cluster version
+	f.vmCountWithBuiltWMCB = 1
+	// vmCountWithBuiltWMCB should not be greater than vmCount
+	if f.vmCountWithBuiltWMCB > vmCount {
+		return fmt.Errorf("tried to run WSU against %d VMs but vmCount set to %d", f.vmCountWithBuiltWMCB, vmCount)
+	}
+
+	// Set up the framework
+	err := f.TestFramework.Setup(vmCount, credentials, skipVMsetup)
+	if err != nil {
+		fmt.Errorf("framework setup failed: %v", err)
+		return err
+	}
+
+	// Set 'buildWMCB' property of Windows VM
+	// Not ideal to set a generic property in a specific implementation. This is a temporary workaround and will be
+	// updated in https://issues.redhat.com/browse/WINC-249
+	for i := 0; i < f.vmCountWithBuiltWMCB; i++ {
+		f.WinVMs[i].SetBuildWMCB(true)
+	}
+
+	return nil
+}
 
 // createhostFile creates an ansible host file for the VMs we have spun up
 func createHostFile(vmList []e2ef.WindowsVM) (string, error) {
@@ -104,7 +136,6 @@ func testAllVMs(t *testing.T) {
 			runTests(t, framework.WinVMs[i])
 		})
 	}
-
 }
 
 // runTests runs all the tests required for a specific VM
@@ -161,7 +192,14 @@ func runWSU(vm e2ef.WindowsVM) ([]byte, error) {
 	}
 
 	// Run the WSU against the VM
-	ansibleCmd = exec.Command("ansible-playbook", "-v", "-i", hostFilePath, playbookPath)
+	if vm.BuildWMCB() {
+		// Build WMCB
+		ansibleCmd = exec.Command("ansible-playbook", "-v", "-i", hostFilePath, playbookPath, "-e",
+			"{build_wmcb: True}")
+	} else {
+		// Download latest released version of WMCB based on cluster version
+		ansibleCmd = exec.Command("ansible-playbook", "-v", "-i", hostFilePath, playbookPath)
+	}
 
 	// Run the playbook
 	wsuOut, err := ansibleCmd.CombinedOutput()
@@ -195,6 +233,15 @@ func runE2ETestSuite(t *testing.T, vm e2ef.WindowsVM, ansibleOutput string) {
 	t.Run("Files copied to Windows node", func(t *testing.T) {
 		testFilesCopied(t, vm, tempDirPath)
 	})
+	if vm.BuildWMCB() {
+		t.Run("Check if wmcb was built", func(t *testing.T) {
+			testBuiltWMCB(t, ansibleOutput)
+		})
+	} else {
+		t.Run("Check if wmcb was downloaded", func(t *testing.T) {
+			testDownloadedWMCB(t, ansibleOutput)
+		})
+	}
 	t.Run("Node is in ready state", func(t *testing.T) {
 		testNodeReady(t, node)
 	})
@@ -216,6 +263,28 @@ func runE2ETestSuite(t *testing.T, vm e2ef.WindowsVM, ansibleOutput string) {
 	t.Run("North-south networking", func(t *testing.T) {
 		testNorthSouthNetworking(t, node, vm)
 	})
+}
+
+// testDownloadedWMCB checks if the task 'Download WMCB' in the Ansible output was executed, not skipped
+func testDownloadedWMCB(t *testing.T, ansibleOutput string) {
+	// split the output for each task
+	tasks := strings.Split(ansibleOutput, "TASK")
+	// grab the output for the task "Download WMCB"
+	downloadWmcbTaskOutput := ""
+	for i := 0; i < len(tasks); i++ {
+		if strings.Contains(tasks[i], "Download WMCB") {
+			downloadWmcbTaskOutput = tasks[i]
+		}
+	}
+	// check if the outcome was changed i.e. the task was executed ("changed": true)
+	require.Contains(t, downloadWmcbTaskOutput, "\"changed\": true")
+}
+
+// testBuiltWMCB checks the Ansible output to ensure WMCB was built
+func testBuiltWMCB(t *testing.T, ansibleOutput string) {
+	require.Contains(t, ansibleOutput,
+		"CGO_ENABLED=0 GO111MODULE=on GOOS=windows go build -o wmcb.exe  "+
+			"github.com/openshift/windows-machine-config-bootstrapper/cmd/bootstrapper")
 }
 
 // testCNIConfig tests if the CNI config has required hostsubnet and servicenetwork CIDR
@@ -255,20 +324,6 @@ func testFilesCopied(t *testing.T, vm e2ef.WindowsVM, ansibleTempDir string) {
 		assert.NoError(t, err, "Error looking for %s: %s", fullPath, err)
 		assert.Emptyf(t, stdout, "Missing file: %s", fullPath)
 	}
-
-	// Check the SHA of kube.tar.gz downloaded
-	kubeTarPath := ansibleTempDir + "\\" + "kube.tar.gz"
-	// certutil is part of default OS installation Windows 7+
-	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
-	stdout, stderr, err := vm.Run(command, false)
-	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
-	require.Equalf(t, "", stderr, "Error generating SHA512 for %s", kubeTarPath)
-	// CertUtil output example:
-	// SHA512 hash of <filepath>:\r\n<SHA-output>\r\nCertUtil: -hashfile command completed successfully.
-	// Extracting SHA value from the output
-	actualKubeTarSha := strings.Split(stdout, "\r\n")[1]
-	assert.Equal(t, expectedKubeTarSha, actualKubeTarSha,
-		"kube.tar.gz downloaded does not match expected checksum")
 }
 
 // testNodeReady tests that the bootstrapped node was added to the cluster and is in the ready state
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -2,8 +2,8 @@
 - hosts: localhost
   vars:
     project_root: "{{ playbook_dir }}/../../../.."
-    kubelet_location: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
     wmcb_exe: "wmcb.exe"
+    hybrid_overlay_exe : "hybrid-overlay.exe"
 
   tasks:
     - name: Check if cluster is using ovn-kubernetes
@@ -20,6 +20,79 @@
         msg: "Cluster not patched for hybrid overlay"
       when: "hybrid_query.stdout == ''"
 
+    # Expected value of kubernetes_version in the form of 'v1.17.2'
+    - name: Get kubernetes version
+      shell: "oc version -o json | jq -r '.serverVersion.gitVersion'"
+      register: kubernetes_version
+      failed_when: kubernetes_version.stdout == ""
+
+    # kubernetes_version v1.17.2 should correspond to 1.17 for getting changelog url
+    - name: Get kubernetes version without patch version
+      shell: "version={{ kubernetes_version.stdout }} ; echo ${version%.*} | tr -d v"
+      register: kubernetes_version_without_patch
+      failed_when: kubernetes_version_without_patch.stdout == ""
+
+    # Changelog url to lookup SHA for kube node binaries
+    # Example: https://github.com/kubernetes/kubernetes/blob/release-1.17/CHANGELOG/CHANGELOG-1.17.md
+    - name: Get kubernetes changelog url
+      shell: |
+        echo "https://github.com/kubernetes/kubernetes/blob/release-"\
+        "{{ kubernetes_version_without_patch.stdout }}/CHANGELOG/CHANGELOG-"\
+        "{{ kubernetes_version_without_patch.stdout }}.md"
+      register: changelog
+
+    # Get SHA512 for kubernetes-node-windows-amd64 binary released for the kubernetes patch version of the cluster
+    - name: Get kube node binary SHA
+      shell: |
+        curl -s "{{ changelog.stdout }}" | \
+        grep -A 1 "{{ kubernetes_version.stdout }}" | \
+        grep -A 1 node-windows | tail -n 1 | sed -e 's/.*<code>\(.*\)<\/code>.*/\1/'
+      register: kube_node_sha
+      failed_when: kube_node_sha.stdout == ""
+
+    - name: Set kubelet location
+      set_fact:
+        # Example of kubelet download url: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
+        kubelet_location: "https://dl.k8s.io/{{ kubernetes_version.stdout }}/kubernetes-node-windows-amd64.tar.gz"
+
+    - name: Get kubernetes minor version
+      shell: "echo {{ kubernetes_version_without_patch.stdout }} | cut -d '.' -f2"
+      register: kubernetes_minor_version
+      failed_when: kubernetes_minor_version.stdout == ""
+
+    # This translates a kubernetes minor version to an OpenShift version. This works under the assumption that each
+    # OpenShift release corresponds with a kubernetes release. This is being done this way, and not with oc get
+    # clusterversion, as OpenShift CI doesn't have the actual version attached to its clusters, instead replacing it
+    # with 0.0.1 and information about the release creation date
+    # TODO Track the validity of this script when OpenShift version is bumped and the corresponding Kubernetes version is not
+    - name: Get cluster version
+      shell: |
+        MINORVERSION={{ kubernetes_minor_version.stdout }}
+        BASEK8SVERSION=16
+        BASEOPENSHIFTVERSION=3
+        if [ \"$MINORVERSION\" -lt \"$BASEK8SVERSION\" ]; then
+        echo unsupported kubernetes version; exit 1
+        fi
+        VERSIONINCREMENTS=$((MINORVERSION - BASEK8SVERSION))
+        echo 4.$((BASEOPENSHIFTVERSION + VERSIONINCREMENTS))
+        exit 0
+      register: cluster_version
+      failed_when: cluster_version.stdout == ""
+
+    # curl -s https://api.github.com/repos/openshift/windows-machine-config-bootstrapper/releases gets the list of all
+    # releases, we use the cluster version from previous ansible task to select the release we need.
+    # TODO Ensure latest release is available for every new OpenShift version
+    - name: Get release
+      shell: |
+        RELEASES="$(curl -s https://api.github.com/repos/openshift/windows-machine-config-bootstrapper/releases)"
+        TAGNAME="$(echo $RELEASES | jq '.[]' | jq '.tag_name' | \
+        grep -i {{ cluster_version.stdout }} |\
+        sort -r | awk 'NR==1')"
+        echo $RELEASES | \
+        jq '.[] | select (.tag_name == '$TAGNAME')'
+      register: release
+      failed_when: release is not defined or release.stdout == ""
+
     - name: Gather required files
       block:
         - name: Create directory to store files to be used later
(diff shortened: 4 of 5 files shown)


internal/test/wsu/wsu_test.go
```diff
@@ -100,15 +96,15 @@ func setupAndTest(t *testing.T, vmNum int) {
 	t.Parallel()
 
 	// Run the WSU against the VM
-	wsuOut, err := runWSU(framework.WinVMs[vmNum])
+	wsuOut, err := runWSU(framework.WinVMs[vmNum], vmNum)
```
ravisantoshgudimetla: nit: Ideally this should be property of the Windows VM. Please think about adding 10 more VM, would numbering help then?

internal/test/wsu/wsu_test.go
```diff
@@ -100,15 +96,15 @@ func setupAndTest(t *testing.T, vmNum int) {
...
 	require.NoError(t, err, "WSU playbook returned error: %s, with output: %s", err, wsuOut)
 
 	// Run our VM test suite
-	runTest(t, framework.WinVMs[vmNum], wsuOut)
+	runTest(t, framework.WinVMs[vmNum], vmNum, wsuOut)
 }
 
 // runWSU runs the WSU playbook against a VM. Returns WSU stdout
-func runWSU(vm e2ef.WindowsVM) (string, error) {
+func runWSU(vm e2ef.WindowsVM, vmNum int) (string, error) {
```
ravisantoshgudimetla: If you have VM struct to additional field called `testLocalBuild`, you wouldn't need to pass vmNum as an argument to any function.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -2,29 +2,131 @@
...
+          when: not build_wmcb or build_wmcb is not defined
+          block:
+            # The wmcb release tags are supposed to conform to the following semver format:
+            # v<openshift-major-version>.<openshift-minor-version>.<wmcb_build_number>-<release_type>. For example, v4.4-1-alpha.
+            # For a <openshift-major-version>.<openshift-minor-version>, there can be many wmcb builds available, for
+            # example, OpenShift 4.4 can map to v4.4.1-alpha and v4.4.2-alpha, we will always pick up the latest build available
+            # which is v4.4.2-alpha,
+            # In future, when we cut beta and GA releases, the playbook should always go for GA version when available.
+            - name: Get wmcb url
+              shell: |
```
ravisantoshgudimetla: Have you tried [`get_url` ](https://docs.ansible.com/ansible/latest/modules/get_url_module.html)module in ansible, it seems you can use checksum as parameter to it, which might save steps for us, like checking for check sum, if ansible is already doing that for us.

internal/test/wsu/wsu_test.go
sebsoto: I agree with this and would like to see this change. In a seperate commit of this PR you can add a getter to the interface which will return the purpose of the VM. This proposed commit would go before this one.

internal/test/wsu/wsu_test.go
sebsoto: I'd argue a more general field than `testLocalBuild` should be used

internal/test/wsu/wsu_test.go
```diff
@@ -224,20 +256,6 @@ func testFilesCopied(t *testing.T, vm e2ef.WindowsVM, ansibleTempDir string) {
...
-	command := fmt.Sprintf("certutil -hashfile %s SHA512", kubeTarPath)
-	stdout, stderr, err := vm.Run(command, false)
-	require.NoError(t, err, "Error generating SHA512 for %s", kubeTarPath)
-	require.Equalf(t, "", stderr, "Error generating SHA512 for %s", kubeTarPath)
-	// CertUtil output example:
-	// SHA512 hash of <filepath>:\r\n<SHA-output>\r\nCertUtil: -hashfile command completed successfully.
-	// Extracting SHA value from the output
-	actualKubeTarSha := strings.Split(stdout, "\r\n")[1]
-	assert.Equal(t, expectedKubeTarSha, actualKubeTarSha,
-		"kube.tar.gz downloaded does not match expected checksum")
```
sebsoto: Why Is this block being removed?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -2,29 +2,131 @@
...
     project_root: "{{ playbook_dir }}/../../../.."
-    kubelet_location: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
     wmcb_exe: "wmcb.exe"
+    hybrid_overlay_exe : "hybrid-overlay.exe"
+    build: False
+    # Dictionary mapping Kubernetes version to the corresponding OpenShift Version
+    map_kubernetes_to_openshift_version:
+      '1.16': 4.3
+      '1.17': 4.4
+      '1.18': 4.5
```
sebsoto: Why did you remove the `Get cluster version` task and replace it with this? Doing this makes the playbook unusable after kubernetes 1.18. 
While I dont see it being used at that point, I'd rather not have a timebomb built into our code.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -2,29 +2,131 @@
...
+    - name: Get kube node binary SHA
+      shell: |
+        curl -s "{{ changelog.stdout }}" | \
+        grep -A 1 "{{ kubernetes_version.stdout }}" | \
+        grep -A 1 node-windows | tail -n 1 | sed -e 's/.*<code>\(.*\)<\/code>.*/\1/'
+      register: kube_node_sha
+
+    # OpenShift cluster version
+    - name: Register cluster version
+      shell: "echo {{ map_kubernetes_to_openshift_version[kubernetes_version_without_patch.stdout] }}"
```
aravindhp: What happens if the version is not present in the map?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -2,29 +2,131 @@
...
+
+        - name: Download WMCB binary
+          when: not build_wmcb or build_wmcb is not defined
+          block:
+            # The wmcb release tags are supposed to conform to the following semver format:
+            # v<openshift-major-version>.<openshift-minor-version>.<wmcb_build_number>-<release_type>. For example, v4.4-1-alpha.
+            # For a <openshift-major-version>.<openshift-minor-version>, there can be many wmcb builds available, for
+            # example, OpenShift 4.4 can map to v4.4.1-alpha and v4.4.2-alpha, we will always pick up the latest build available
+            # which is v4.4.2-alpha,
+            # In future, when we cut beta and GA releases, the playbook should always go for GA version when available.
```
aravindhp: +1 for comments like this.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -2,29 +2,131 @@
...
+            - name: Get wmcb url
+              shell: |
+                echo '{{ release.stdout }}' | \
+                jq ' .assets[] | select (.name == "{{ wmcb_exe }}" )' | \
+                jq -r .browser_download_url
+              register: wmcb_download_url
```
aravindhp: There are multiple places where you are looking at `release.stdout`. What happens if the release is malformed and you don't the expected value? Do you have guards against this down the line?

internal/test/wsu/wsu_test.go
suhanime: For now I have used `buildWmcb` since that is the flag we are passing through wsu. I am looking for suggestions for a more generalized name

tools/ansible/tasks/wsu/main.yaml
suhanime: That is correct - `get_url` along with `win_copy` modules should ensure checksums are checked. I mimicked the SHA check of WMCB from that of hybrid-overlay, thinking it should cover the edge case of redirects - if that situation ever arises.

internal/test/wsu/wsu_test.go
suhanime: Since we are downloading kube node binaries dynamically now, I moved the SHA check to WSU playbook

tools/ansible/tasks/wsu/main.yaml
suhanime: Your point is valid. My thinking when deciding to move to map was to consider the scenario when OpenShift version skips a kubernetes version, like the way it happened in 4.3, and it would be easier to maintain a map rather than changing the script

internal/test/wsu/wsu_test.go
```diff
@@ -41,6 +35,8 @@ var (
 	windowsServerImage = "mcr.microsoft.com/windows/servercore:ltsc2019"
 	// ubi8Image is the name/location of the linux image we will use for testing
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
+	// builtWmcbVM indicates which VM will use built wmcb
+	builtWmcbVm = 1
```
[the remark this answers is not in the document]
suhanime: I have changed this option to 'vmCountWithBuiltWmcb' option in main_test, so it is now capable of running tests for wsu with built wmcb on multiple vms if needed

tools/ansible/tasks/wsu/main.yaml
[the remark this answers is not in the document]
suhanime: I have replaced the existing shell script with a map - we can remove 1.18 mapping for now

tools/ansible/tasks/wsu/main.yaml
suhanime: Added a guard for this

tools/ansible/tasks/wsu/main.yaml
suhanime: I have added a guard to ensure we have correct cluster version (by checking the presence of kubernetes version in the map) and I am checking if the release object was defined and not empty before proceeding

internal/test/wsu/wsu_test.go
suhanime: Thank you for pointing it out. I have added a property to the interface

internal/test/framework/windowsvm.go
```diff
@@ -64,6 +66,8 @@ type WindowsVM interface {
 	Reinitialize() error
 	// Destroy destroys the Windows VM
 	Destroy() error
+	GetBuildWmcb() bool
+	SetBuildWmcb(bool)
```
aravindhp: Please add docs

internal/test/wsu/main_test.go
```diff
@@ -8,12 +8,15 @@ import (
...
-	framework = &e2ef.TestFramework{}
+	wsuFramework = WsuFramework{}
 	// TODO: expose this to the end user as a command line flag
 	// vmCount is the number of VMs the test suite requires
-	vmCount = 1
+	// Bring up 2 vms to test wsu run with automatic download of WMCB based on cluster version as well as built from
+	// source version
+	vmCount = 2
+	// Number of VMs which should use built version of WMCB
+	vmCountWithBuiltWmcb = 1
```
aravindhp: Shouldn't this be a property of the `WsuFramework`?

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,38 @@ var (
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
 
+type WsuFramework struct {
+	framework *e2ef.TestFramework
```
aravindhp: You can just embed the framework here. There is no need to name it as `framework`. That will make the referencing of its fields in later code more succinct.

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,38 @@ var (
...
+}
+
+func (f *WsuFramework) RetrieveArtifacts() {
+	f.framework.RetrieveArtifacts()
+}
+
+func (f *WsuFramework) TearDown() {
+	f.framework.TearDown()
+}
+
```
aravindhp: If you embed the framework, there will be no need of these wrappers unless we want to specialize the retrieval and tear down.

tools/ansible/tasks/wsu/main.yaml
aravindhp: I don't follow you reasoning. Are you saying that `get_url` and `win_copy` do support redirects?

tools/ansible/tasks/wsu/main.yaml
aravindhp: If the `release.stdout` is malformed, it could potentially have data but just not in the format you expect. In those scenarios the above code will fail. That will result in variables like `wmcb_download_url` having incorrect or no data.

tools/ansible/tasks/wsu/main.yaml
sebsoto: @suhanime I am :-1: on this change, if you think this is a better way of doing things, we can get a third opinion, else please revert this change :)

suhanime: I think it is easier to manage versions we support via a map. @ravisantoshgudimetla @aravindhp @akhil-rane please weigh in to help us decide. The discussion is about dropping (Get cluster version task)[https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/tools/ansible/tasks/wsu/main.yaml#L88-L104] and using a map of kubernetes version to cluster version instead

aravindhp: I am not a fan of either method mainly because we are using them so that things would work in CI. Given we are back here, I suggest to add a proper way that works for the "product" and we have a fall back hack that works for CI. That fall back hack should be simple and easy to maintain.

internal/test/wsu/main_test.go
suhanime: I can change to make it so - I figured keeping this option here (and eventually expose it to command line) would be a nice-to-have

tools/ansible/tasks/wsu/main.yaml
suhanime: We can use `oc version` here to get openshift version as well, and use the hack if the output is not in the format we expect (like 4.x) - does that seem good enough for the product?

tools/ansible/tasks/wsu/main.yaml
suhanime: I am not 100% sure if `get_url` doesn't automatically follow redirects (although `win_get_url` doesn't - that's why we have added the option [here](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/tools/ansible/tasks/wsu/main.yaml#L135). While exploring with [checksum](https://docs.ansible.com/ansible/latest/modules/get_url_module.html#parameter-checksum) parameter of `get_url`, I discovered that it doesn't work the way we want it to. It uses this checksum so it can skip download if the file is already present, but if the file downloaded doesn't meet the checksum provided, then it fails with "The checksum for <file_downloaded> did not match <expected_checksum>; it was <actual_checksum>". If we are okay with this, I can make the changes.

tools/ansible/tasks/wsu/main.yaml
suhanime: Since I only moved this task from `win` host to `localhost` without making any significant changes, I think this is beyond the scope of this ticket

internal/test/wsu/main_test.go
aravindhp: You would need to separate out the two concerns, i.e. command line option vs framework and then use what is giving in the CLI to instantiate the framework

tools/ansible/tasks/wsu/main.yaml
aravindhp: Using the `oc version` is fine with me. Others can chime in here too.

tools/ansible/tasks/wsu/main.yaml
aravindhp: Isn't this the behavior that we want?

tools/ansible/tasks/wsu/main.yaml
aravindhp: That is a fair statement. It just means that you need to ensure that the guard you added for `release.stdout` is good enough which I don't think is the case.

internal/test/wsu/main_test.go
ravisantoshgudimetla: The variable will be mapped to a field in `WsuFramework` so that endusers can access it via cli. Correct me if I am wrong @suhanime

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: We sort of discussed about this, @suhanime has some ideas around getting log files specific to wsu and wmcb runs.

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: > It uses this checksum so it can skip download if the file is already present,

What's wrong with `force` option if you want the download to happen always.

> but if the file downloaded doesn't meet the checksum provided, then it fails with "The checksum for <file_downloaded> did not match <expected_checksum>; it was <actual_checksum>". If we are okay with this, I can make the changes.

Don't we fail, if the checksum fails even today?

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: Hey, that's my comment :P

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: To be clear, we shouldn't add something to product for it to work in CI. More code means more bugs. Given that, I am fine with having a map which gets updated for every release. Ideally, we would cut a new release in our repo for every release of OpenShift so that there is one-to-one mapping. We don't have to maintain this map going forward.

tools/ansible/tasks/wsu/main.yaml
ravisantoshgudimetla: The playbook will fail when the actual download happens here https://github.com/openshift/windows-machine-config-bootstrapper/pull/148/files#diff-def35121485b59e044decc98a81c6402R118 and increasing the ansible log-level will help in getting the error. We can have some checks at get_url or let run ansible in dry-run mode to get the info.

internal/test/wsu/main_test.go
suhanime: Correct @ravisantoshgudimetla . For now I am separating the concerns though - I moved the variable to be a property of wsuFramework and hardcoding it to 1, but the building blocks are in place if we wanted to expose it to CLI (with vmCount as well) in the future

internal/test/wsu/wsu_test.go
suhanime: Yes - we can specialize the RetrieveArtifacts for WsuFramework when [WINC-155](https://issues.redhat.com/browse/WINC-155) is picked up

tools/ansible/tasks/wsu/main.yaml
suhanime: After discussing it with @ravisantoshgudimetla , I have decided to use `get_url` module's checksum parameter, which eliminates the need to check the checksum on windows instance. Implemented for both kube node binaries and downloaded wmcb

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+        grep -A 1 node-windows | tail -n 1 | sed -e 's/.*<code>\(.*\)<\/code>.*/\1/'
+      register: kube_node_sha
+
+    - name: Set kubelet location
+      set_fact:
+        # Example of kubelet download url: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
+        kubelet_location: "https://dl.k8s.io/{{ kubernetes_version.stdout }}/kubernetes-node-windows-amd64.tar.gz"
+
+    - name: Get kubernetes minor version
+      shell: "oc version -o json | jq -r '.serverVersion.minor' | tr -d +"
```
sebsoto: We are calling oc version multiple times, we should try and avoid that if possible, and parse the output of the original command as we need it.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+          # 'build_wmcb' is used as a boolean - it has to be provided in JSON format
+          # https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#passing-variables-on-the-command-line
+          when: build_wmcb is defined and build_wmcb
+          block:
+          - name: Build WMCB
+            make:
+              target: build
+              chdir: "{{ project_root }}"
+
+          - name: Move WMCB to temporary directory
```
sebsoto: We should build WMCB directly in the temp directory. Thats a solution to the bug @akhil-rane is reporting.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+            # For a <openshift-major-version>.<openshift-minor-version>, there can be many wmcb builds available, for
+            # example, OpenShift 4.4 can map to v4.4.1-alpha and v4.4.2-alpha, we will always pick up the latest build available
+            # which is v4.4.2-alpha,
+            # In future, when we cut beta and GA releases, the playbook should always go for GA version when available.
+            - name: Get wmcb url
+              shell: |
+                echo '{{ release.stdout }}' | \
+                jq ' .assets[] | select (.name == "{{ wmcb_exe }}" )' | \
+                jq -r .browser_download_url
+              register: wmcb_download_url
```
sebsoto: We need to make sure this will fail as well, when stdout is empty

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+
+            # We'll get the SHA associated with the wmcb.exe in the body of the release tag.
+            - name: Get the SHA for wmcb binary. We assume that body of WMCB release has the info.
+              shell: |
+                echo '{{ release.stdout }}' | \
+                jq -r .body | \
+                grep -i {{ wmcb_exe }}| \
+                awk '{print $1}'
+              register: wmcb_sha
```
sebsoto: Make sure stdout isnt empty

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+                dest: "{{ tmp_dir.path }}/wmcb.exe"
+                checksum: "sha256:{{ wmcb_sha.stdout }}"
+
+        # Get url to download latest hybrid-overlay for the OpenShift cluster version
+        - name: Get hybrid overlay url
+          shell: |
+            echo '{{ release.stdout }}' | \
+            jq ' .assets[] | select (.name == "{{ hybrid_overlay_exe }}" )' | \
+            jq -r .browser_download_url
+          register: hybrid_overlay_download_url
```
sebsoto: Make sure stdout isnt empty

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
...
+
+        # We'll get the SHA associated with the hybrid_overlay.exe in the body of the release tag.
+        - name: Get the SHA for hybrid-overlay binary. We assume that body of WMCB release has the info.
+          shell: |
+            echo '{{ release.stdout }}' | \
+            jq -r .body | \
+            grep -i {{ hybrid_overlay_exe }}| \
+            awk '{print $1}'
+          register: hybrid_overlay_sha
```
sebsoto: Make sure stdout isnt empty

internal/test/framework/windowsvm.go
```diff
@@ -64,6 +66,12 @@ type WindowsVM interface {
 	Reinitialize() error
 	// Destroy destroys the Windows VM
 	Destroy() error
+	// GetBuildWmcb gets the value of buildWmcb. If value is true, WSU will build WMCB before transferring. If value is
+	// false, WSU will download the latest released version of WMCB based on the cluster version
+	GetBuildWmcb() bool
```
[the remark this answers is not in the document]
aravindhp: I agree with Sebastain. @ravisantoshgudimetla and I were discussing moving `WindowsVM` to WNI so that we can reuse it in WMCO. Once we do that, we will remove this from the framework. Then in the framework it can be specialized as `TestWindowsVM`.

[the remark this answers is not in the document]
ravisantoshgudimetla: Or compose your interface with other interface called TestWindowsVM and have these fields there. In either case, that's a good call.

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,35 @@ var (
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
 
+type WsuFramework struct {
```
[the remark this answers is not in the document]
ravisantoshgudimetla: Why should we make this public? Is it being used anywhere else?

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,35 @@ var (
...
+}
+
+// Setup initializes the framework. 'vmCountWithBuildWmcb' indicates how many VMs out of 'vmCount' created would run WSU
+// that uses built version of WMCB. Others will download the latest released version of WMCB as per the cluster version
+func (f *WsuFramework) Setup(vmCount int, credentials []*types.Credentials, skipVMsetup bool) error {
+	f.TestFramework = &e2ef.TestFramework{}
+	// vmCountWithBuiltWmcb should not be greater than vmCount
+	// If vmCount is 2, 1 VM will run WSU that will build WMCB and the other will auto-download the latest WMCB based on
+	// the cluster version
+	f.vmCountWithBuiltWmcb = 1
```
ravisantoshgudimetla: You're hardcoding this, any reason for doing so or do you want to handle computing what all VMs need local build version in a later stage? If yes, please add a TODO on how want to do it. Perhaps you can create a story around it as well.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,151 @@
         msg: "Cluster not patched for hybrid overlay"
       when: "hybrid_query.stdout == ''"
 
+    - name: Get kubernetes version
+      shell: "oc version -o json | jq -r '.serverVersion.gitVersion'"
```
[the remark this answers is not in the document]
ravisantoshgudimetla: Good point, also make sure that you're getting the entire version and store it a variable so that you can avoid making unnecessary api calls.

internal/test/wsu/wsu_test.go
suhanime: Changed comment and added a TODO in the struct to explain my thought process

tools/ansible/tasks/wsu/main.yaml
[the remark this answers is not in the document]
suhanime: It was already registered in a variable, added failed_when condition
There was another API call not added in my PR, in task "Get kubernetes minor version" - I have changed it so it now parses the input from the variable instead of making another API call

internal/test/framework/windowsvm.go
```diff
@@ -39,9 +39,28 @@ type windowsVM struct {
 	sshClient *ssh.Client
 	// winrmClient to access the Windows VM created
 	winrmClient *winrm.Client
+	// windowsVMLabels is a map of test specific labels for the Windows VM
+	*windowsVMLabels
```
aravindhp: This should not be here. What you need to do is create a `testWindowsVM` struct and embed the bool in there.
```
type testWindowsVM struct {
    windowsVM
    buildWMCB bool
}
```

internal/test/wsu/wsu_test.go
```diff
@@ -44,8 +39,38 @@ var (
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
 
+type wsuFramework struct {
+	// TestFramework holds the instantiation of test suite being executed. As of now, temp dir is hardcoded.
+	*e2ef.TestFramework
+	// Number of VMs which should use built version of WMCB
+	// TODO Expose this option to the user along with vmCount
```
aravindhp: Please add a Jira story and link to it here.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,160 @@
...
+        "{{ kubernetes_version_without_patch.stdout }}/CHANGELOG/CHANGELOG-"\
+        "{{ kubernetes_version_without_patch.stdout }}.md"
+      register: changelog
+
+    - name: Get kube node binary SHA
+      shell: |
+        curl -s "{{ changelog.stdout }}" | \
+        grep -A 1 "{{ kubernetes_version.stdout }}" | \
+        grep -A 1 node-windows | tail -n 1 | sed -e 's/.*<code>\(.*\)<\/code>.*/\1/'
+      register: kube_node_sha
```
aravindhp: Please add a `failed_when` here.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,160 @@
...
+
+    - name: Set kubelet location
+      set_fact:
+        # Example of kubelet download url: "https://dl.k8s.io/v1.16.2/kubernetes-node-windows-amd64.tar.gz"
+        kubelet_location: "https://dl.k8s.io/{{ kubernetes_version.stdout }}/kubernetes-node-windows-amd64.tar.gz"
+
+    - name: Get kubernetes minor version
+      shell: "echo {{ kubernetes_version_without_patch.stdout }} | cut -d '.' -f2"
+      register: kubernetes_minor_version
```
aravindhp: Please add a `failed_when` here.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,160 @@
...
+        MINORVERSION={{ kubernetes_minor_version.stdout }}
+        BASEK8SVERSION=16
+        BASEOPENSHIFTVERSION=3
+        if [ \"$MINORVERSION\" -lt \"$BASEK8SVERSION\" ]; then
+        echo unsupported kubernetes version; exit 1
+        fi
+        VERSIONINCREMENTS=$((MINORVERSION - BASEK8SVERSION))
+        echo 4.$((BASEOPENSHIFTVERSION + VERSIONINCREMENTS))
+        exit 0
+      register: cluster_version
```
aravindhp: Please add a `failed_when` here.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,160 @@
...
+        grep -i {{ cluster_version.stdout }} |\
+        sort -r | awk 'NR==1')"
+        echo $RELEASES | \
+        jq '.[] | select (.tag_name == '$TAGNAME')'
+      register: release
+
+    - name: Check release object
+      fail:
+        msg: "Cannot find the latest release"
+      when: release is not defined or release.stdout == ""
```
aravindhp: What is the Ansible standard for using this vs `failed_when`?

tools/ansible/tasks/wsu/main.yaml
aravindhp: > which eliminates the need to check the checksum on windows instance

Why? What if there is an error when transferring the file to the Windows instance?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -20,25 +20,160 @@
...
+        echo "https://github.com/kubernetes/kubernetes/blob/release-"\
+        "{{ kubernetes_version_without_patch.stdout }}/CHANGELOG/CHANGELOG-"\
+        "{{ kubernetes_version_without_patch.stdout }}.md"
+      register: changelog
+
+    - name: Get kube node binary SHA
+      shell: |
+        curl -s "{{ changelog.stdout }}" | \
+        grep -A 1 "{{ kubernetes_version.stdout }}" | \
+        grep -A 1 node-windows | tail -n 1 | sed -e 's/.*<code>\(.*\)<\/code>.*/\1/'
```
ravisantoshgudimetla: What does this mean?Please add a comment here?

suhanime: I have added a comment here to give an overview of what we are trying to get

tools/ansible/tasks/wsu/main.yaml
suhanime: Doesn't make much difference except `fail` gives us a custom error message. I must have missed this as a part of my revisions - I am removing this module and adding a `failed_when` in the task above with the same condition

tools/ansible/tasks/wsu/main.yaml
suhanime: `win_copy` will ensure they are copied over correctly or fail

aravindhp: Please define "correctly" here. It is not clear how and when the cheksum is done as per the [docs](https://docs.ansible.com/ansible/latest/modules/win_copy_module.html).

suhanime: I added some comments with `Download wmcb binary` task and `Copy required files to Windows host` task that should ensure the integrity of downloaded as well as copied file.
It is implicitly defined in the [docs for force parameter](https://docs.ansible.com/ansible/latest/modules/win_copy_module.html#parameter-force) that default run is doing the checksum

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,35 @@ var (
...
+	*e2ef.TestFramework
+	// Number of VMs which should use built version of WMCB
+	// TODO Expose this option to the user along with vmCount -> https://issues.redhat.com/browse/WINC-240
+	vmCountWithBuiltWmcb int
+}
+
+// Setup initializes the framework.
+func (f *wsuFramework) Setup(vmCount int, credentials []*types.Credentials, skipVMsetup bool) error {
+	f.TestFramework = &e2ef.TestFramework{}
+	// vmCountWithBuiltWmcb should not be greater than vmCount
```
aravindhp: Please add a check for this

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -82,6 +217,7 @@
       debug:
         msg: "Windows temporary directory: {{ win_temp_dir.path }}"
 
+    # Uses winRM to transfer files over. Copies entire content or fails
```
aravindhp: Please call out that checksum is performed given that it is not documented clearly.

internal/test/wmcb/wmcb_test.go
```diff
@@ -73,9 +73,9 @@ var (
 	}
 )
 
-// wmcbVM is a wrapper for the WindowsVM interface that associates it with WMCB specific testing
+// wmcbVM is a wrapper for the TestWindowsVM interface that associates it with WMCB specific testing
```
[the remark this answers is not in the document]
suhanime: As per our discussion, it is simpler to keep `buildWMCB` property in `WindowsVM` for now, and I have created https://issues.redhat.com/browse/WINC-249 to tackle more graceful handling of specializing WindowsVM for both e2e tests.

internal/test/framework/windowsvm.go
```diff
@@ -64,6 +67,11 @@ type WindowsVM interface {
 	Reinitialize() error
 	// Destroy destroys the Windows VM
 	Destroy() error
+	// BuildWMCB returns the value of buildWMCB. It can be used by WSU to decide if it should build WMCB before using it
+	BuildWMCB() bool
```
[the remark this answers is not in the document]
suhanime: Because https://golang.org/doc/effective_go.html#Getters lists 

> it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner. The use of upper-case names for export provides the hook to discriminate the field from the method.

internal/test/wsu/wsu_test.go
```diff
@@ -44,6 +39,41 @@ var (
...
+	}
+
+	// Set up the framework
+	err := f.TestFramework.Setup(vmCount, credentials, skipVMsetup)
+	if err != nil {
+		fmt.Errorf("framework setup failed: %v", err)
+		return err
+	}
+
+	// Set 'buildWMCB' property of Windows VM
```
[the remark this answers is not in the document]
suhanime: `buildWMCB` declaration in `windowsVM` has a comment that is pointing to the JIRA ticket

> // TODO This is a WSU specific property and should be moved to wsu_test -> https://issues.redhat.com/browse/WINC-249

internal/test/wsu/wsu_test.go
```diff
@@ -218,6 +263,28 @@ func runE2ETestSuite(t *testing.T, vm e2ef.WindowsVM, ansibleOutput string) {
...
 }
 
+// testDownloadedWMCB checks if the task 'Download WMCB' in the Ansible output was executed, not skipped
+func testDownloadedWMCB(t *testing.T, ansibleOutput string) {
+	// split the output for each task
+	tasks := strings.Split(ansibleOutput, "TASK")
+	// grab the output for the task "Download WMCB"
+	downloadWmcbTaskOutput := ""
+	for i := 0; i < len(tasks); i++ {
+		if strings.Contains(tasks[i], "Download WMCB") {
```
ravisantoshgudimetla: Sorry for not catching this earlier but are you sure, you'd always have the string to be same or do you want to ignore the case and see if that matches what we have?

internal/test/wsu/wsu_test.go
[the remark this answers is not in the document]
ravisantoshgudimetla: No, I mean copying the values from a generic method back to specific method is considered wrong right? Without this, I would not understand why this bad.

internal/test/wsu/wsu_test.go
suhanime: It is the name of the Task it is testing, so it should always be present, unless the playbook changes - in which the test should be rewritten anyways
I don't mind changing the case, but I do not see the point of doing so

internal/test/wsu/wsu_test.go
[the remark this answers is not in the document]
suhanime: Added, although I must say I am very cautious about adding repetitive comments across the code base

PR #154, opened 2020-02-21
Bug 1805513: Resolve parallel execution of playbook

akhil-rane:
Problem: When we run wsu playbook parrallely for 2 windows vms, the ansible task
'Move wmcb to temporary directory' fails for the second vm. This is because
currently we are building wmcb.exe in the project root directory in both the runs.
So when the run for the first vm has moved the built wmcb.exe to a temporary
directory, run for the second vm cannot find it.

Solution: Instead of moving built wmcb.exe to a temporary directory, we are
copying it

### DIFF
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -32,8 +32,8 @@
             target: build
             chdir: "{{ project_root }}"
 
-        - name: Move WMCB to temporary directory
-          command: mv "{{ project_root }}/{{ wmcb_exe }}" "{{ tmp_dir.path }}/{{ wmcb_exe }}"
+        - name: Copy WMCB to temporary directory
+          command: cp "{{ project_root }}/{{ wmcb_exe }}" "{{ tmp_dir.path }}/{{ wmcb_exe }}"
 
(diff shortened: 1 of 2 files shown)


tools/ansible/tasks/wsu/main.yaml
```diff
@@ -30,10 +30,8 @@
         - name: Build WMCB
           make:
             target: build
-            chdir: "{{ project_root }}"
-
-        - name: Move WMCB to temporary directory
-          command: mv "{{ project_root }}/{{ wmcb_exe }}" "{{ tmp_dir.path }}/{{ wmcb_exe }}"
```
aravindhp: Couldn't you fix it by copying the file instead?

akhil-rane: We could, but that will leave a copy of wmcb.exe in the project root directory.

aravindhp: Is that a problem? Space is cheap :wink:

akhil-rane: Addressed this

suhanime:
/lgtm
Thank you for the PR @akhil-rane

PR #155, opened 2020-02-21
[wni] Introduce WindowsVM interface

ravisantoshgudimetla:
As of now, we don't have a way to interact with the windows VM object
we created. This change introduces the Windows VM interface and adds
methods to access to the Windows VM object created using SSH and WinRM.
 Most of the code is moved from framework in the internal package.
This should help in re-using WNI to launch instances in our Windows
Machine Config controller.

/cc @aravindhp @suhanime @sebsoto


tools/windows-node-installer/pkg/types/types.go
```diff
@@ -9,9 +23,215 @@ const (
...
+type Windows struct {
+	// Credentials to access the Windows VM created
+	Credentials Credentials
+	// SSHClient contains the ssh client information to access the Windows VM via ssh
+	SSHClient *ssh.Client
+	// WinrmClient to access the Windows VM created
+	WinrmClient *winrm.Client
+}
+
+// WindowsVM is the interface for interacting with a Windows VM in the test framework
```
ravisantoshgudimetla: note to me: Remove test framework :(

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -197,14 +198,31 @@ func (a *AwsProvider) CreateWindowsVM() (credentials *types.Credentials, err err
...
+	// TODO: Parse the output of the `Get-Service sshd, ssh-agent` on the Windows node to check if the windows nodes
+	// has those services present
+	time.Sleep(time.Minute)
+	if err := w.ConfigureOpenSSHServer(); err != nil {
+		return w, fmt.Errorf("failed to configure OpenSSHServer on the Windows VM: %v", err)
+	}
+	if err := w.GetSSHClient(); err != nil {
+		return w, fmt.Errorf("failed to get ssh client for the Windows VM created: %v", err)
+	}
+
```
aravindhp: This function is getting large and unwieldy. Please move this to a helper.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -777,6 +778,27 @@ func (az *AzureProvider) CreateWindowsVM() (*types.Credentials, error) {
...
+
+	// Wait for some time before starting configuring of ssh server. This is to let sshd service be available
+	// in the list of services
+	// TODO: Parse the output of the `Get-Service sshd, ssh-agent` on the Windows node to check if the windows nodes
+	// has those services present
+	time.Sleep(time.Minute)
+	if err := w.GetSSHClient(); err != nil {
+		return w, fmt.Errorf("failed to get ssh client for the Windows VM created: %v", err)
+	}
+
```
aravindhp: Same comment here and if possible make it a common helper.

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -19,7 +19,7 @@ import (
 type Cloud interface {
 	// CreateWindowsVM creates a Windows VM for a given cloud provider
 	// TODO: CreateWindowsVM should return a provider object for further interaction with the created instance.
```
aravindhp: Please remove this TODO and update the doc.

tools/windows-node-installer/pkg/types/types.go
```diff
@@ -9,9 +23,215 @@ const (
...
 	FirewallRuleName = "ContainerLogsPort"
+	// remotePowerShellCmdPrefix holds the PowerShell prefix that needs to be prefixed  for every remote PowerShell
+	// command executed on the remote Windows VM
+	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
+	// winRMPort is port used for WinRM communication
+	winRMPort = 5986
 )
 
+// Windows represents a Windows host.
+type Windows struct {
```
aravindhp: Does this need to be exported? If not I would name this `windowsVM`.

tools/windows-node-installer/pkg/types/types.go
```diff
@@ -9,9 +23,215 @@ const (
...
+	sshClient, err := ssh.Dial("tcp", w.Credentials.GetIPAddress()+":22", config)
+	if err != nil {
+		return fmt.Errorf("failed to dial to ssh server: %s", err)
+	}
+	w.SSHClient = sshClient
+	return nil
+}
+
 // Credentials holds the information to access the Windows instance created.
+// TODO: Make this part of WindowsVM struct later
```
aravindhp: Please add a Jira story to track this and link it here.

tools/windows-node-installer/pkg/types/types.go
aravindhp: I realize looking at #157, that this is required.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: How about we make the helper public and keeps these other methods as private? Is there a reason to expose all these methods? That was the code in the internal framework will become smaller too.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: Added a TODO. https://github.com/openshift/windows-machine-config-bootstrapper/pull/155/files#diff-356b22f90392ac8fc81c91247ebede89R205

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -197,14 +198,34 @@ func (a *AwsProvider) CreateWindowsVM() (credentials *types.Credentials, err err
...
 	// the credentials are being generated properly. This method won't guarantee the existence of credentials
 	// if the VM is spun up
-	credentials = types.NewCredentials(instanceID, publicIPAddress, decryptedPassword, winUser)
+	credentials := types.NewCredentials(instanceID, publicIPAddress, decryptedPassword, winUser)
+	w.Credentials = credentials
+
+	// Setup Winrm and SSH client so that we can interact with the Windows Object we created
+	// TODO: Add a struct called Connectivity which has information related to Winrm, SSH and have
+	//		getters and setters for it so that it can be exposed a public method
+	// 		jira: https://issues.redhat.com/browse/WINC-239
```
aravindhp: Thanks for adding this. However it would be better if you added this as a comment in the `Windows` struct where these fields are defined. That seems like the best common place to add this comment as the placement here could imply it is ec2 specific.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -129,7 +129,8 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 // resourceTrackerDir.
 // On success, the function outputs RDP access information in the commandline interface. It also returns the
 // the credentials to access the Windows VM created,
-func (a *AwsProvider) CreateWindowsVM() (credentials *types.Credentials, err error) {
+func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
```
suhanime: func docs aren't representing the correct parameters

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -715,7 +715,8 @@ func (az *AzureProvider) constructNetworkProfile(ctx context.Context,
 // resourceGroupName as the existing OpenShift
 // TODO: If it fails during the instance creation process it has to delete the resources created
 // untill that step.
-func (az *AzureProvider) CreateWindowsVM() (*types.Credentials, error) {
+func (az *AzureProvider) CreateWindowsVM() (types.WindowsVM, error) {
```
suhanime: docs for this func need to be updated too

tools/windows-node-installer/pkg/types/types.go
```diff
@@ -9,8 +23,217 @@ const (
...
+	// remotePowerShellCmdPrefix holds the PowerShell prefix that needs to be prefixed  for every remote PowerShell
+	// command executed on the remote Windows VM
+	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
+	// winRMPort is port used for WinRM communication
+	winRMPort = 5986
 )
 
+// Windows represents a Windows host.
+// TODO: Add a struct called Connectivity which has information related to Winrm, SSH and have
+//		getters and setters for it so that it can be exposed a public method
```
suhanime: as a public method*

tools/windows-node-installer/test/e2e/aws_test.go
```diff
@@ -127,19 +127,19 @@ func setupAWSCloudProvider() error {
...
 	// Create the instance
-	credentials, err = awsProvider.CreateWindowsVM()
+	windowsVM, err = awsProvider.CreateWindowsVM()
 	if err != nil {
 		return fmt.Errorf("error creating Windows instance: %s", err)
 	}
 
 	// Ensure we have the login info for the instance
-	if credentials == nil {
+	if windowsVM == nil {
```
suhanime: Does the error message below represent the intended check?

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
ravisantoshgudimetla: Good catch. Updated. PTAL.

PR #159, opened 2020-02-25
[wni] Add a DestroyWindowsVM method to cloud provider interface

ravisantoshgudimetla:
As of now, we're deleting the entire Windows VMs created via WNI using
windows-node-installer.json. We'd want a way to just delete a single
Windows VM created via WNI so that we can use it in WMCO. This commit expands
the interface to have that method.

/cc @aravindhp @openshift/openshift-team-windows-containers

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -275,6 +275,20 @@ func (a *AwsProvider) getDecodedPassword(instanceID string) ([]byte, error) {
 
 }
 
+// DestroyWindowsVM destroys the given Windows VM
+func (a *AwsProvider) DestroyWindowsVM(instanceID string) error {
+	if err := a.TerminateInstance(instanceID); err != nil {
+		return fmt.Errorf("failed deleting the instance %s: %v", instanceID, err)
+	}
+	if err := a.waitUntilInstanceTerminated(instanceID); err != nil {
+		// As of now, I want the error thrown to be consistent across all cloud providers, so I am not specializing
+		// the information in the error message string but the err will have more context. This is cause our operator
+		// to block for a while
+		return fmt.Errorf("failed deleting the instance %s: %v", instanceID, err)
+	}
+	return nil
+}
+
 // waitUntilPasswordDataIsAvailable waits till the ec2 password data is available.
 // AWS sdk's WaitUntilPasswordDataAvailable is returning inspite of password data being available.
 // So, building this function as a wrapper around AWS sdk's GetPasswordData method with constant back-off
--- a/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
@@ -933,6 +933,45 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
 	return
 }
 
+// DestroyWindowsVMWithResources destroys the Windows VMs along with all the errors associated with it.
+func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
+	ctx := context.Background()
+	vmInfo, err := az.vmClient.Get(ctx, az.resourceGroupName, vmName, compute.InstanceView)
+	if err != nil {
+		log.Printf("error getting vminfo for %s: %v", vmName, err)
+	}
+	err, nicName := az.getNICname(ctx, vmName)
+	if err != nil {
+		log.Printf("error getting the nic name for %s: %v", vmName, err)
+	}
+	err, ipName := az.getIPname(ctx, vmName)
+	if err != nil {
+		log.Printf("error getting the ip name for the instance %s: %v", vmName, err)
+	}
+
+	err = az.destroyInstance(ctx, vmName)
+	if err != nil {
+		log.Printf("error destroying the instance %s: %v", vmName, err)
+	}
+
+	err = az.destroyNIC(ctx, nicName)
+	if err != nil {
+		log.Printf("error deleting the NIC of instance %s: %v", vmName, err)
+	}
+
+	err = az.destroyIP(ctx, ipName)
+	if err != nil {
+		log.Printf("error deleting the IP of instance %s: %v", vmName, err)
+	}
+
+	// This may still leak resources as we'll not be able to get vm info once the VM gets deleted.
+	err = az.destroyDisk(ctx, vmInfo)
+	if err != nil {
+		log.Printf("error deleting the disk attached to the instance %s: %v", vmName, err)
+	}
+	return nil
+}
+
 // DestroyWindowsVMs destroys all the resources created by the CreateWindows method.
 func (az *AzureProvider) DestroyWindowsVMs() error {
 	// Read from `windows-node-installer.json` file
@@ -950,33 +989,8 @@ func (az *AzureProvider) DestroyWindowsVMs() error {
 	var rdpFilePath string
 
 	for _, vmName := range installerInfo.InstanceIDs {
-
-		_, nicName := az.getNICname(ctx, vmName)
-		_, ipName := az.getIPname(ctx, vmName)
-
-		vmInfo, _ := az.vmClient.Get(ctx, az.resourceGroupName, vmName, compute.InstanceView)
-
-		log.Printf("deleting the resources associated with instance %s: %s", vmName, err)
-
-		err = az.destroyInstance(ctx, vmName)
-		if !errorCheck(err) {
-			log.Printf("deleted the instance '%s'", vmName)
-		}
-
-		err = az.destroyNIC(ctx, nicName)
-		if !errorCheck(err) {
-			log.Printf("deleted the NIC of instance")
-		}
-
-		err = az.destroyIP(ctx, ipName)
-		if !errorCheck(err) {
-			log.Printf("deleted the IP of instance")
-		}
-
-		err = az.destroyDisk(ctx, vmInfo)
-		if !errorCheck(err) {
-			log.Printf("deleted the disk attached to the instance")
-		}
+		log.Printf("deleting the resources associated with instance %s", vmName)
+		az.DestroyWindowsVM(vmName)
 
 		rdpFilePath = az.resourceTrackerDir + vmName
 		err = resource.DeleteCredentialData(rdpFilePath)
--- a/tools/windows-node-installer/pkg/cloudprovider/factory.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/factory.go
@@ -29,6 +29,10 @@ type Cloud interface {
 	// It deletes the security group only if the group is not associated with any instance.
 	// The association between the instance and security group are available from individual cloud provider.
 	DestroyWindowsVMs() error
+	// DestroyWindowsVM destroys a specific instance that was passed to it. It returns an error when the WindowsVM
+	// deletion fails. It takes the instanceID as argument depending on the cloud provider implementation the
+	// instanceID. Let's if we want to slice of instances instead of individual instances
+	DestroyWindowsVM(string) error
 }
 
 // CloudProviderFactory returns cloud specific interface for performing necessary functions related to creating or


tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,15 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
 	return
 }
 
+// DestroyWindowsVM destroys the given Windows VM
+func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
+	ctx := context.Background()
+	if err := az.destroyInstance(ctx, vmName); err != nil {
```
aravindhp: This is not sufficient for azure. You will have to do what is being done in `DestroyWindowsVMs()` otherwise you will leak those other resources. Please make a common helper that you call from both the places.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -275,6 +275,20 @@ func (a *AwsProvider) getDecodedPassword(instanceID string) ([]byte, error) {
...
+func (a *AwsProvider) DestroyWindowsVM(instanceID string) error {
+	var err error
+	if err = a.TerminateInstance(instanceID); err != nil {
+		return fmt.Errorf("failed deleting the instance %s: %v", instanceID, err)
+	}
+	if err := a.waitUntilInstanceTerminated(instanceID); err != nil {
+		// As of now, I want the error thrown to be consistent across all cloud providers, so I am not specializing
+		// the information in the error message string but the err will have more context
+		return fmt.Errorf("failed deleting the instance %s: %v", instanceID, err)
+	}
```
aravindhp: This can be made into a helper that is also called from `DestroyWindowsVMs()`

ravisantoshgudimetla: We've a fire & forget operation within DestrotWindowsVMs method, so we cannot share the code.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: Thanks for catching, I've added it now.

ravisantoshgudimetla:
WSU failed twice for now. Latest failure:

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-bootstrapper/159/pull-ci-openshift-windows-machine-config-bootstrapper-master-e2e-wsu/165

/retest

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -275,6 +275,21 @@ func (a *AwsProvider) getDecodedPassword(instanceID string) ([]byte, error) {
 
 }
 
+// DestroyWindowsVM destroys the given Windows VM
+func (a *AwsProvider) DestroyWindowsVM(instanceID string) error {
+	var err error
```
sebsoto: I dont think this line is necessary

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
 	return
 }
 
+// DestroyWindowsVMWithResources destroys the Windows VMs along with all the errors associated with it
+func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
+	ctx := context.Background()
+	_, nicName := az.getNICname(ctx, vmName)
```
sebsoto: Why are we dropping these errors

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	_, ipName := az.getIPname(ctx, vmName)
```
sebsoto: Don't drop these errors

sebsoto: Don't drop these errors

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	vmInfo, _ := az.vmClient.Get(ctx, az.resourceGroupName, vmName, compute.InstanceView)
```
sebsoto: Don't drop the error

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+
+	log.Printf("deleting the resources associated with instance %s", vmName)
```
sebsoto: nit: I'm not a fan of this pattern of logging the current action within functions like this. If we want to log this I would rather the caller of `DestroyWindowsVM` log it.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
 	return
 }
 
+// DestroyWindowsVMWithResources destroys the Windows VMs along with all the errors associated with it
+func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
```
sebsoto: nit: Should this function have a check to see if the VM actually exists? In the case it doesn't exist, return nil.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	log.Printf("deleting the resources associated with instance %s", vmName)
+
+	err := az.destroyInstance(ctx, vmName)
+	if err != nil {
+		return fmt.Errorf("error deleting the instance %s: %v", vmName, err)
+	}
+
+	err = az.destroyNIC(ctx, nicName)
+	if err != nil {
+		return fmt.Errorf("error deleting the NIC of instance %s: %v", vmName, err)
```
sebsoto: If this fails the other resources (IP/disk) will be left around and will be unfindable since we use the VM to find them.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,37 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	}
+
+	err = az.destroyIP(ctx, ipName)
+	if err != nil {
+		return fmt.Errorf("error deleting the IP of instance %s: %v", vmName, err)
```
sebsoto: Same as above, but for disk only

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: Old pattern, I just copied it. It's interesting that the azure api guys have error as first argument and necessary return value as second argument. I don't mind adding it.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: I don't like it either, this is a problem of the way the code is structured. I've moved it to the caller.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: Not exactly, I think you'll still be able to find resources based on the VM name from console. The other approach is log and continue which I am not a fan either. I think if the api call fail, we should fail early instead of collecting all the failures and then reporting back(if that's what you're suggesting).

sebsoto: I meant findable through our getNIC functions - but yeah not a great solution etiher way. Im fine with leaving it how it is

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -936,13 +936,17 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
 // DestroyWindowsVMWithResources destroys the Windows VMs along with all the errors associated with it
 func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
 	ctx := context.Background()
-	_, nicName := az.getNICname(ctx, vmName)
-	_, ipName := az.getIPname(ctx, vmName)
+	err, nicName := az.getNICname(ctx, vmName)
+	if err != nil {
+		return fmt.Errorf("error deleting the instance %s: %v", vmName, err)
```
sebsoto: error message doesn't provide context about whats failing - `fmt.Errorf("error getting NIC name: %v", err)`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -936,13 +936,17 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	}
+	err, ipName := az.getIPname(ctx, vmName)
+	if err != nil {
+		return fmt.Errorf("error deleting the instance %s: %v", vmName, err)
```
sebsoto: fmt.Errorf("error getting IP name: %v", err)

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
aravindhp: I am nack on returning errors here. For the short term I prefer logging and ensuring that we don't leak resources. This becomes more important when we enable CI for Azure. Once we are closer to the product we can add these methods to as interface and explicitly destroy them from the Operator and use the same process for both AWS and Azure. But I don't see us getting there anytime soon.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,44 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+
+	err = az.destroyIP(ctx, ipName)
+	if err != nil {
+		log.Printf("error deleting the IP of instance %s: %v", vmName, err)
+	}
+	vmInfo, err := az.vmClient.Get(ctx, az.resourceGroupName, vmName, compute.InstanceView)
+	if err != nil {
+		log.Printf("error getting vminfo for %s: %v", vmName, err)
+	}
+	// This will still leak resources as we'll not be able to get vm info once the VM gets deleted.
```
sebsoto: I dont understand this comment, what azure resources are being leaked here?

sebsoto: Are you saying this will always fail?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: > I prefer logging and ensuring that we don't leak resources.

I made the changes to log and continue.
We'd still leak resources when the VM gets deleted and operator calls the `DestroyWindowsVM` method here -
https://github.com/openshift/windows-machine-config-bootstrapper/pull/159/files#diff-33ef6d4eee7a2e0183e61be55d8e3e79R967


For the record, I'd let the end-user deal with it. Even in our CI, if we properly create resources, the destroy should provide enough guarantees for VM deletion not the resources associated with it. Either we should we have external entity deal with it which is capable of deleting the lingering resources. If we start providing guarantees around the cloud provider api, we'll have to write a controller for wni :(.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: The disk deletion won't happen as it needs vminfo and vminfo would be nil for the deleted vm.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -933,6 +933,44 @@ func (az *AzureProvider) destroyDisk(ctx context.Context, vmInfo compute.Virtual
...
+	err = az.destroyNIC(ctx, nicName)
+	if err != nil {
+		log.Printf("error deleting the NIC of instance %s: %v", vmName, err)
+	}
+
+	err = az.destroyIP(ctx, ipName)
+	if err != nil {
+		log.Printf("error deleting the IP of instance %s: %v", vmName, err)
+	}
+	vmInfo, err := az.vmClient.Get(ctx, az.resourceGroupName, vmName, compute.InstanceView)
```
aravindhp: How is this working today? Won't the VM be destroyed by this point?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
aravindhp: >We'd still leak resources when the VM gets deleted and operator calls the DestroyWindowsVM method here 

I don't follow this scenario. Why would the operator call the `DestroyWindowsVM()` method on a VM that has been deleted.

>For the record, I'd let the end-user deal with it. Even in our CI, if we properly create resources, the destroy should provide enough guarantees for VM deletion not the resources >associated with it. Either way we should have external entity deal with it which is capable of deleting the lingering resources. 

I agree with you and the point I am making is once we have this external entity capable of deleting lingering resources, we can change this code.

>If we start providing guarantees around the cloud >provider api, we'll have to write a controller for wni :(

I am not saying we will ever provide such guarantees.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
sebsoto: Then shouldn't you move it before the VM is deleted? I don't understand why we are calling a function that is guaranteed to fail

ravisantoshgudimetla: I just moved it back. Even if I move it, if the Delete operation is called on non-existent VM instead of failing hard where are proceeding further which will cause problems :(.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: > I don't follow this scenario. Why would the operator call the DestroyWindowsVM() method on a VM that has been deleted.

If our node objects slice which has info of instanceIDs is stale, we can call the DestoryVM. Remember kubernetes is a eventually consistent system, so there will be stages where we get errors but we should be able to auto correct over time.

ravisantoshgudimetla:
/approve cancel

@sebsoto and @aravindhp - Addressed your comments. PTAL

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
sebsoto: It seems odd to me to call a function, expecting it to fail.

ravisantoshgudimetla:
OVN failing to come up

```
 level=error msg="Cluster operator network Degraded is True with RolloutHung: DaemonSet \"openshift-ovn-kubernetes/ovnkube-node\" rollout is not making progress - last change 2020-03-02T12:46:18Z"
level=info msg="Cluster operator network Progressing is True with Deploying: DaemonSet \"openshift-multus/multus-admission-controller\" is not yet scheduled on any nodes\nDaemonSet \"openshift-ovn-kubernetes/ovnkube-node\" is not available (awaiting 3 nodes)" ```

This has been the case for atleast 3 of the jobs that I looked at.

aravindhp:
/override ci/prow/e2e-wsu

Overriding the test as we are running into a OVN blocker and this PR has to merge for us to move forward with the operator work. Moreover this PR is not making any changes to the WSU playbook.

aravindhp:
/override ci/prow/e2e-wsu

Overriding as this test has nothing to do with the PR and more importantly this blocks the operator work from proceeding.

aravindhp:
/override ci/prow/e2e-wmcb

Not sure why the e2e-wmcb is being run again as it [passed](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-bootstrapper/159/pull-ci-openshift-windows-machine-config-bootstrapper-master-e2e-wmcb/181) before.
Overriding as this test has nothing to do with the PR and more importantly this blocks the operator work from proceeding.

PR #162, opened 2020-03-03
[wsu] [CI] Capture hybrid-overlay and kube-proxy output in artifacts_dir

mansikulkarni96:
This PR captures the kube-proxy logs in C:\k\log\kubeProxy\ and the hybrid-overlay logs in C:\k\log\hybridOverlay\ in the remote VM logs directory .

Kube proxy logs are already being captured so they will just be redirected to this new directory and hybrid overlay logs are captured using RedirectStandardError and RedirectStandardOutput arguments in the WSU playbook while the binaries are running. 

This PR also consists addition of  a method RetrieveDirectories to the WindowsVM interface, this was required as the remote log directory on the VM can now consist of a nested directory structure which should be copied to the artifacts directory. 

The E2E test  make sure that the log directories and files are being captured in C:\K\log on the VM.

[[WINC-155]](https://issues.redhat.com/browse/WINC-155)

### DIFF
--- a/internal/test/framework/windowsvm.go
+++ b/internal/test/framework/windowsvm.go
@@ -35,9 +35,9 @@ type testWindowsVM struct {
 // TestWindowsVM is the interface for interacting with a Windows VM in the test framework. This will hold the
 // specialized information related to test suite
 type TestWindowsVM interface {
-	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
-	// now, we're limiting every file in the remote directory to be written to single directory on the local host
-	RetrieveFiles(string, string) error
+	// RetrieveDirectories recursively copies the files and directories from the directory in the remote Windows VM
+	// to the given directory on the local host.
+	RetrieveDirectories(string, string) error
 	// Destroy destroys the Windows VM
 	// TODO: Remove this and move it to framework or other higher level object capable of doing deletions.
 	//		jira: https://issues.redhat.com/browse/WINC-243
@@ -104,19 +104,15 @@ func newWindowsVM(imageID, instanceType string, credentials *types.Credentials, 
 	return w, nil
 }
 
-// RetrieveFiles retrieves list of files from remote directory to the local directory.
-// The implementation can be changed if the use-case arises. As of now, we're doing a best effort
-// to collect every log possible. If a retrieval of file fails, we would proceed with retrieval
-// of other log files.
-func (w *testWindowsVM) RetrieveFiles(remoteDir, localDir string) error {
+func (w *testWindowsVM) RetrieveDirectories(remoteDir string, localDir string) error {
 	if w.SSHClient == nil {
-		return fmt.Errorf("RetrieveFile cannot be called without a ssh client")
+		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
 	}
 
-	// Create local dir
+	// creating a local directory to store the files and directories from remote directory.
 	err := os.MkdirAll(localDir, os.ModePerm)
 	if err != nil {
-		log.Printf("could not create %s: %s", localDir, err)
+		return fmt.Errorf("could not create %s: %v", localDir, err)
 	}
 
 	sftp, err := sftp.NewClient(w.SSHClient)
@@ -128,46 +124,52 @@ func (w *testWindowsVM) RetrieveFiles(remoteDir, localDir string) error {
 	// Get the list of all files in the directory
 	remoteFiles, err := sftp.ReadDir(remoteDir)
 	if err != nil {
-		return fmt.Errorf("error opening remote file: %v", err)
+		return fmt.Errorf("error opening remote directory %s: %v", remoteDir, err)
 	}
 
 	for _, remoteFile := range remoteFiles {
-		// Assumption: We ignore the directories here the reason being RetrieveFiles should just retrieve files
-		// in a directory, if this is directory, we should have called RetrieveFiles on this directory
+		remotePath := filepath.Join(remoteDir, remoteFile.Name())
+		localPath := filepath.Join(localDir, remoteFile.Name())
+		// check if it is a directory, call itself again
 		if remoteFile.IsDir() {
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
+				log.Printf("error while retrieving %s directory from Windows : %v", remotePath, err)
+			}
+		} else {
+			// logging errors as a best effort to retrieve files from remote directory
+			if err = w.copyFileFrom(sftp, remotePath, localPath); err != nil {
+				log.Printf("error while retrieving %s file from Windows : %v", remotePath, err)
+			}
 		}
-		fileName := remoteFile.Name()
-		dstFile, err := os.Create(filepath.Join(localDir, fileName))
-		if err != nil {
-			log.Printf("error creating file locally: %v", err)
-			continue
-		}
-		// TODO: Check if there is some performance implication of multiple Open calls.
-		srcFile, err := sftp.Open(remoteDir + "\\" + fileName)
+	}
+	return nil
+}
 
-		if err != nil {
-			log.Printf("error while opening remote directory on the Windows VM: %v", err)
-			continue
-		}
-		_, err = io.Copy(dstFile, srcFile)
-		if err != nil {
-			log.Printf("error retrieving file %v from Windows VM: %v", fileName, err)
-			continue
-		}
-		// flush memory
-		if err = dstFile.Sync(); err != nil {
-			log.Printf("error flusing memory: %v", err)
-			continue
-		}
-		if err := srcFile.Close(); err != nil {
-			log.Printf("error closing file on the remote host %s", fileName)
-			continue
-		}
-		if err := dstFile.Close(); err != nil {
-			log.Printf("error closing file %s locally", fileName)
-			continue
-		}
+// copyFileFrom copies a file from remote directory to the local directory.
+func (w *testWindowsVM) copyFileFrom(sftp *sftp.Client, remotePath, localPath string) error {
+	localFile, err := os.Create(localPath)
+	if err != nil {
+		return fmt.Errorf("error creating file locally: %v", err)
+	}
+	// TODO: Check if there is some performance implication of multiple Open calls.
+	remoteFile, err := sftp.Open(remotePath)
+	if err != nil {
+		return fmt.Errorf("error while opening remote file on the Windows VM: %v", err)
+	}
+	// logging the errors instead of returning to allow closing of files
+	_, err = io.Copy(localFile, remoteFile)
+	if err != nil {
+		log.Printf("error retrieving file %v from Windows VM: %v", localPath, err)
+	}
+	// flush memory
+	if err = localFile.Sync(); err != nil {
+		log.Printf("error flusing memory: %v", err)
+	}
+	if err := remoteFile.Close(); err != nil {
+		log.Printf("error closing file on the remote host %s", localPath)
+	}
+	if err := localFile.Close(); err != nil {
+		log.Printf("error closing file %s locally", localPath)
 	}
 	return nil
 }
--- a/internal/test/wsu/wsu_test.go
+++ b/internal/test/wsu/wsu_test.go
@@ -39,6 +39,9 @@ var (
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
 
+const hybridOverlayDir = "C:\\k\\log\\hybrid-overlay"
+const kubeProxyDir = "C:\\k\\log\\kube-proxy"
+
 type wsuFramework struct {
 	// TestFramework holds the instantiation of test suite being executed
 	*e2ef.TestFramework
@@ -244,11 +247,15 @@ func runE2ETestSuite(t *testing.T, vm e2ef.TestWindowsVM, ansibleOutput string) 
 	tempDirPath, err := getAnsibleTempDirPath(ansibleOutput)
 	require.NoError(t, err, "Could not get path of Ansible temp directory")
 
+	binaryFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe", "hybrid-overlay.exe", "kube.tar.gz"}
+	hybridOverlaylogList := []string{"hybridOverlayStdout.log", "hybridOverlayStderr.log"}
+	kubeProxylogList := []string{"kube-proxy.exe.INFO", "kube-proxy.exe.WARNING"}
+
 	node, err := framework.GetNode(vm.GetCredentials().GetIPAddress())
 	require.NoError(t, err, "Could not get Windows node object")
 
 	t.Run("Files copied to Windows node", func(t *testing.T) {
-		testFilesCopied(t, vm, tempDirPath)
+		testRemoteFilesExist(t, vm, binaryFileList, tempDirPath)
 	})
 	if vm.BuildWMCB() {
 		t.Run("Check if wmcb was built", func(t *testing.T) {
@@ -261,6 +268,10 @@ func runE2ETestSuite(t *testing.T, vm e2ef.TestWindowsVM, ansibleOutput string) 
 	}
 	t.Run("Node is in ready state", func(t *testing.T) {
 		testNodeReady(t, node)
+	})
+	t.Run("check if hybrid-overlay and kube-proxy log exists", func(t *testing.T) {
+		testRemoteFilesExist(t, vm, kubeProxylogList, kubeProxyDir)
+		testRemoteFilesExist(t, vm, hybridOverlaylogList, hybridOverlayDir)
 	})
 	t.Run("Check if worker label has been applied to the Windows node", func(t *testing.T) {
 		testWorkerLabelsArePresent(t, node)
@@ -328,13 +339,11 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.TestWindowsVM, ansibleTe
 	assert.Contains(t, cniConfigFileContents, requiredServiceNetwork, "CNI config does not contain service network")
 }
 
-// testFilesCopied tests that the files we attempted to copy to the Windows host, exist on the Windows host
-func testFilesCopied(t *testing.T, vm e2ef.TestWindowsVM, ansibleTempDir string) {
-	expectedFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe", "hybrid-overlay.exe", "kube.tar.gz"}
-
+// testRemoteFilesExist tests that the files we expect, exist on the Windows host
+func testRemoteFilesExist(t *testing.T, vm e2ef.TestWindowsVM, expectedFileList []string, directoryPath string) {
 	// Check if each of the files we expect on the Windows host are there
 	for _, filename := range expectedFileList {
-		fullPath := ansibleTempDir + "\\" + filename
+		fullPath := directoryPath + "\\" + filename
 		// This command will write to stdout, only if the file we are looking for does not exist
 		command := fmt.Sprintf("if not exist %s echo fail", fullPath)
 		stdout, _, err := vm.Run(command, false)
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -206,6 +206,7 @@
   vars:
     tmp_path: "{{ playbook_dir }}/tmp"
     ovn_annotation: "{{ 'hostsubnet' if  ( hostvars['localhost']['cluster_version']['stdout']  == '4.3' ) else  'node-subnet' }}"
+    log_dir: "c:\\k\\log"
 
   tasks:
     - name: Create temporary directory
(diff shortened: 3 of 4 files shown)


internal/test/framework/windowsvm.go
```diff
@@ -157,21 +158,20 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
-		return fmt.Errorf("RetrieveFile cannot be called without a ssh client")
+		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
 	}
 
-	// Create local dir
+	// Create a local dir
 	err := os.MkdirAll(localDir, os.ModePerm)
 	if err != nil {
-		log.Printf("could not create %s: %s", localDir, err)
+		return fmt.Errorf("could not create %s: %v", localDir, err)
```
suhanime: In case the `os.MkdirAll` fails because directory already exists, can we just append the contents (or replace the existing logs)?

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
-		}
-		if err := srcFile.Close(); err != nil {
-			log.Printf("error closing file on the remote host %s", fileName)
-			continue
-		}
-		if err := dstFile.Close(); err != nil {
-			log.Printf("error closing file %s locally", fileName)
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
+				log.Printf("error while retrieving directiories from Windows VM %s : %v", remotePath, err)
```
suhanime: *error while retrieving %s directory from...

internal/test/framework/windowsvm.go
```diff
@@ -49,9 +49,10 @@ type WindowsVM interface {
 	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
 	// does not exist
 	CopyFile(string, string) error
-	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
-	// now, we're limiting every file in the remote directory to be written to single directory on the local host
-	RetrieveFiles(string, string) error
+	// RetrieveDirectories retrieves the list of files and directories from the directory in the remote Windows VM to
```
sebsoto: `retrieves the list of files and directories` This is saying that the function serves the same function as a remote `ls`.

This comment can be reworded as
```
RetrieveDirectories recursively copies the files and directories from the directory in the remote Windows VM
to the given directory on the local host.
```

internal/test/framework/windowsvm.go
```diff
@@ -49,9 +49,10 @@ type WindowsVM interface {
 	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
 	// does not exist
 	CopyFile(string, string) error
-	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
-	// now, we're limiting every file in the remote directory to be written to single directory on the local host
-	RetrieveFiles(string, string) error
```
sebsoto: Do we really want to remove this method fully and replace it with a recursive version? I'd think we'd want to be able to copy the contents of a directory without recursing.

internal/test/framework/windowsvm.go
```diff
@@ -49,9 +49,10 @@ type WindowsVM interface {
...
+	// RetrieveDirectories retrieves the list of files and directories from the directory in the remote Windows VM to
+	// the local host. As of now, we're limiting every file in the remote directory to be written to single directory
+	// on the local host
+	RetrieveDirectories(string, string) error
```
sebsoto: nit: I think the name should be changed to reflect that this is recursive, but that might just be me

internal/test/framework/windowsvm.go
```diff
@@ -157,21 +158,20 @@ func (w *windowsVM) CopyFile(filePath, remoteDir string) error {
...
+// if it encounters a file.
+func (w *windowsVM) RetrieveDirectories(remoteDir string, localDir string) error {
+
 	if w.sshClient == nil {
-		return fmt.Errorf("RetrieveFile cannot be called without a ssh client")
+		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
 	}
 
-	// Create local dir
+	// Create a local dir
```
sebsoto: `Create a local dir` that is whats happening, but the line below is self explanatory
I suggest saying what the directory is for :)

internal/test/wsu/wsu_test.go
```diff
@@ -244,11 +246,15 @@ func runE2ETestSuite(t *testing.T, vm e2ef.WindowsVM, ansibleOutput string) {
 	tempDirPath, err := getAnsibleTempDirPath(ansibleOutput)
 	require.NoError(t, err, "Could not get path of Ansible temp directory")
 
+	binaryFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe", "hybrid-overlay.exe", "kube.tar.gz"}
+	hybridOverlaylogList := []string{"hybridOverlayStdout.log", "hybridOverlayStderr.log"}
+	kubeProxylogList := []string{"kube-proxy.exe.INFO", "kube-proxy.exe.WARNING"}
```
suhanime: Isn't there any error file generated for kube-proxy?

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
-		}
-		if err := dstFile.Close(); err != nil {
-			log.Printf("error closing file %s locally", fileName)
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
+				log.Printf("error while retrieving directiories from Windows VM %s : %v", remotePath, err)
+			}
+		} else {
+			// best effort to retrieve files
+			w.retrieveFiles(sftp, remotePath, localPath)
```
sebsoto: This wont work for nested directories. You need to make sure the parent directory is created within retireveFiles

internal/test/framework/windowsvm.go
```diff
@@ -49,9 +49,10 @@ type WindowsVM interface {
 	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
 	// does not exist
 	CopyFile(string, string) error
-	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
-	// now, we're limiting every file in the remote directory to be written to single directory on the local host
-	RetrieveFiles(string, string) error
+	// RetrieveDirectories retrieves the list of files and directories from the directory in the remote Windows VM to
+	// the local host. As of now, we're limiting every file in the remote directory to be written to single directory
+	// on the local host
```
aravindhp: With this limitation, files with the same name from different remote directories will be overwritten locally. What is the downside in creating local directories?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -329,8 +330,22 @@
       win_shell: "Stop-Process -Name \"hybrid-overlay\""
       when: 'get_process.stderr == ""'
 
+    # The bootstrapper runs in one of the tasks above and makes C:\\k\\log directory created before hand
+    # Creating a hybrid-overlay directory to store hybrid-overlay specific logs
+    - name: check if hybridOverlay directory already exists
```
suhanime: Maybe we should mention log directory for hybrid overlay?

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
+			w.retrieveFiles(sftp, remotePath, localPath)
 		}
 	}
 	return nil
 }
 
+// retrieveFiles retrieves list of files from remote directory to the local directory.
+// we're doing a best effort to collect every log possible. If a retrieval of file fails,
+// we would proceed with retrieval of other log files.
+func (w *windowsVM) retrieveFiles(sftp *sftp.Client, remotePath, localPath string) {
```
sebsoto: This function should return an error instead of logging. We want to bubble errors up as much as possible.

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
+
+	localFile, err := os.Create(localPath)
+	if err != nil {
+		log.Printf("error creating file locally: %v", err)
```
sebsoto: The rest of the function depends on localFile existing, you need to return at this point. If there was an error this would cause a panic

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
+	}
+	if err := localFile.Close(); err != nil {
+		log.Printf("error closing file %s locally", localPath)
+	}
+}
+
 func (w *windowsVM) Run(cmd string, psCmd bool) (string, string, error) {
 	if w.winrmClient == nil {
-		return "", "", fmt.Errorf("Run cannot be called without a WinRM client")
+		return "", "", fmt.Errorf("run cannot be called without a WinRM client")
```
sebsoto: I would also love for this to be changed, but its unrelated to the PR
Its good that you caught this, but we want to avoid doing random fixes within tickets. :)
But maybe its ok in this case? I'll let others chime in here

aravindhp: Please revert.

internal/test/framework/windowsvm.go
sebsoto: This seems like it only copies a single file, but its called `retrieveFiles` (plural). The description also isnt accurately describing whats going on. Please update the function name and comment.

internal/test/wsu/wsu_test.go
```diff
@@ -39,6 +39,9 @@ var (
 	ubi8Image = "registry.access.redhat.com/ubi8/ubi:latest"
 )
 
+const hybridOverlayLogPath = "C:\\k\\log\\hybridOverlay"
```
aravindhp: `LogPath` seems to imply it could be a path for a single log file. What is the downside of using `hybridOverlayDir`?

Please use `hybrid-overlay` for the directory name.

internal/test/wsu/wsu_test.go
```diff
@@ -39,6 +39,9 @@ var (
...
+const kubeProxyLogPath = "C:\\k\\log\\kubeProxy"
```
aravindhp: `kube-proxy` for directory name.

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
 	// Get the list of all files in the directory
 	remoteFiles, err := sftp.ReadDir(remoteDir)
 	if err != nil {
-		return fmt.Errorf("error opening remote file: %v", err)
+		return fmt.Errorf("error opening remote directory %s: %v", remoteDir, err)
 	}
 
 	for _, remoteFile := range remoteFiles {
```
sebsoto: @mansikulkarni96 Have you looked at the ftp client `Walk()` function? Whats the downside of using that here instead of writing your own?

internal/test/wsu/wsu_test.go
```diff
@@ -328,13 +338,11 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.WindowsVM, ansibleTempDi
 	assert.Contains(t, cniConfigFileContents, requiredServiceNetwork, "CNI config does not contain service network")
 }
 
-// testFilesCopied tests that the files we attempted to copy to the Windows host, exist on the Windows host
-func testFilesCopied(t *testing.T, vm e2ef.WindowsVM, ansibleTempDir string) {
-	expectedFileList := []string{"kubelet.exe", "worker.ign", "wmcb.exe", "hybrid-overlay.exe", "kube.tar.gz"}
-
+// testFilesExist tests that the files we expect, exist on the Windows host
+func testFilesExist(t *testing.T, vm e2ef.WindowsVM, expectedFileList []string, directoryPath string) {
```
aravindhp: Given you are testing if remote files exists, how about calling this `testRemoteFilesExist`?

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -329,8 +330,22 @@
...
+      win_stat:
+        path: "{{ log_dir }}\\hybridOverlay"
```
aravindhp: `hybrid-overlay`

internal/test/framework/windowsvm.go
sebsoto: Actually, it looks like you're creating this instead of using the `CopyFile` function we have for the windowsVM. Please use the existing `CopyFile()` function if possible

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -407,10 +422,23 @@
         name: "kube-proxy"
         state: absent
 
+    # The bootstrapper runs in one of the above tasks and makes C:\\k\\log directory created before hand
+    # Creating a kubeProxy directory to store kubeProxy specific logs
+    - name: check if kubeProxy directory already exists
+      win_stat:
+        path: "{{ log_dir }}\\kubeProxy"
```
aravindhp: `kube-proxy`

internal/test/framework/windowsvm.go
mansikulkarni96: I can log the error instead, so it passes to the next step. Even if the directory exists, retrieveFiles can copy the file to already existing local directory.

internal/test/wsu/wsu_test.go
mansikulkarni96: missed this, as it wasn't created for one of the tests I ran, can verify again and update.

internal/test/framework/windowsvm.go
```diff
@@ -181,53 +181,58 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
-			continue
-		}
-		if err := srcFile.Close(); err != nil {
-			log.Printf("error closing file on the remote host %s", fileName)
-			continue
-		}
-		if err := dstFile.Close(); err != nil {
-			log.Printf("error closing file %s locally", fileName)
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
```
aravindhp: You are passing `localPath` which in this case is a directory, so won't a new local directory get created? Or am I not following the code correctly?

internal/test/framework/windowsvm.go
mansikulkarni96: the original method retrieveFiles only copies files, while it skipped all the directories whereas this can handle both the scenarios.

internal/test/framework/windowsvm.go
mansikulkarni96: I tried using the walk function initially but it takes the argument root string and tries to find this remote root directory path that we have using lstat on local and then applies the function we define inside it. So this was increasing the complexity for a case where we first establish a sftp connection and get a []os.fileInfo object in order to read this remote path(VM).

internal/test/framework/windowsvm.go
mansikulkarni96: mkdirall makes a new local directory using parameter localDir which is the original argument, I am modifying and sending the localPath and not the localDir here.

internal/test/framework/windowsvm.go
mansikulkarni96: This change is not required as os.MkdirAll returns nil(no err) if the directory already exists.

internal/test/framework/windowsvm.go
```diff
@@ -158,19 +157,18 @@ func (w *windowsVM) CopyFileToRemote(filePath, remoteDir string) error {
...
 	if w.sshClient == nil {
 		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
 	}
 
-	// Create a local dir
+	// creating a local directory to store the files and directories from remote directory.
 	err := os.MkdirAll(localDir, os.ModePerm)
 	if err != nil {
-		return fmt.Errorf("could not create %s: %v", localDir, err)
+		fmt.Errorf("could not create %s: %v", localDir, err)
```
sebsoto: This is creating a string but not doing anything with it.

internal/test/framework/windowsvm.go
```diff
@@ -181,46 +179,51 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
 		if remoteFile.IsDir() {
-			continue
-		}
-		fileName := remoteFile.Name()
-		dstFile, err := os.Create(filepath.Join(localDir, fileName))
-		if err != nil {
-			log.Printf("error creating file locally: %v", err)
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
+				log.Printf("error while retrieving %s directiory from Windows : %v", remotePath, err)
```
sebsoto: I'm not a fan of logging within the WIndowsVM methods. If we do this anything using this library will have log messages in its stdout. In my eyes, a better way to do this would be passing in a logger in when creating the WindowsVM struct which WindowsVM will then log with. I am not suggesting doing that within this PR.

Could use a second opinion, @aravindhp thoughts on this?

internal/test/framework/windowsvm.go
```diff
@@ -181,46 +179,51 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
+			}
+		} else {
+			if err = w.copyFileFromRemote(sftp, remotePath, localPath); err != nil {
+				log.Printf("error while retrieving %s file from Windows : %v", remotePath, err)
```
sebsoto: Same as above

aravindhp: I think @mansikulkarni96 is doing a best effort to retrieve the files.

@mansikulkarni96 please retain the comment.

internal/test/framework/windowsvm.go
```diff
@@ -190,49 +188,49 @@ func (w *windowsVM) RetrieveDirectories(remoteDir string, localDir string) error
...
 		}
 	}
 	return nil
 }
 
-// retrieveFiles retrieves list of files from remote directory to the local directory.
-// we're doing a best effort to collect every log possible. If a retrieval of file fails,
-// we would proceed with retrieval of other log files.
-func (w *windowsVM) retrieveFiles(sftp *sftp.Client, remotePath, localPath string) {
+// copyFileFromRemote copies a file from remote directory to the local directory.
```
aravindhp: Please rename to `copyFileFrom`

internal/test/framework/windowsvm.go
```diff
@@ -190,49 +188,49 @@ func (w *windowsVM) RetrieveDirectories(remoteDir string, localDir string) error
...
 	}
 	if err := remoteFile.Close(); err != nil {
-		log.Printf("error closing file on the remote host %s", localPath)
+		return fmt.Errorf("error closing file on the remote host %s", localPath)
 	}
 	if err := localFile.Close(); err != nil {
-		log.Printf("error closing file %s locally", localPath)
+		return fmt.Errorf("error closing file %s locally", localPath)
 	}
+	return nil
```
aravindhp: Why are you returning the errors here? There was a reason why they were logs in the first place. For example if the `Copy`fails you still need to `Close` the files.

mansikulkarni96: sorry for the confusion, but I thought we want to bubble up the errors as much as possible. But I get the point that we need to close the files. I ll change them to logs.

mansikulkarni96:
Addressed comments @aravindhp @sebsoto PTAL.

internal/test/framework/windowsvm.go
```diff
@@ -181,46 +177,53 @@ func (w *windowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
 		if remoteFile.IsDir() {
-			continue
-		}
-		fileName := remoteFile.Name()
-		dstFile, err := os.Create(filepath.Join(localDir, fileName))
-		if err != nil {
-			log.Printf("error creating file locally: %v", err)
-			continue
+			if err = w.RetrieveDirectories(remotePath, localPath); err != nil {
+				log.Printf("error while retrieving %s directiory from Windows : %v", remotePath, err)
```
aravindhp: Spelling error in `directory`. Please fix when squashing your commits.

internal/test/wsu/wsu_test.go
mansikulkarni96: yes because that file is not always present in the directory.

internal/test/framework/windowsvm.go
```diff
@@ -128,46 +124,53 @@ func (w *testWindowsVM) RetrieveFiles(remoteDir, localDir string) error {
...
-			log.Printf("error closing file on the remote host %s", fileName)
-			continue
-		}
-		if err := dstFile.Close(); err != nil {
-			log.Printf("error closing file %s locally", fileName)
-			continue
-		}
+// copyFileFrom copies a file from remote directory to the local directory.
+func (w *testWindowsVM) copyFileFrom(sftp *sftp.Client, remotePath, localPath string) error {
+
```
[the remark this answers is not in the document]
mansikulkarni96: removed blank line

internal/test/wsu/wsu_test.go
```diff
@@ -200,7 +203,6 @@ func captureWSULogs(wsuOut []byte, vm e2ef.TestWindowsVM, logFileName string) {
 // runWSU runs the WSU playbook against a VM. Returns WSU stdout
 func runWSU(vm e2ef.TestWindowsVM) ([]byte, error) {
 	var ansibleCmd *exec.Cmd
-
```
ravisantoshgudimetla: Why remove this line?

mansikulkarni96: sorry looks like a miss, updating it.

sebsoto:
/hold 
Waiting to see artifacts
/lgtm

sebsoto:
Thanks @mansikulkarni96 artifacts look good. Only issue is that we are overwriting the hybrid overlay output on the second test run. Something to keep in mind moving forward is that we should find a way to append the hybrid overlay output. I think its better to handle that in the operator/if the hybrid overlay becomes a Windows service.
/hold cancel

PR #163, opened 2020-03-03
[wsu] Download latest 0.8.x version of CNI plugins

akhil-rane:
Currently, we are hard coding the CNI plugins version to 0.8.2. This PR makes
wsu playbook dynamically pick the latest 0.8.x version of CNI plugins from repo's
release page.

### DIFF
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -69,7 +70,7 @@ type TestFramework struct {
 	ClusterVersion string
 	// latestRelease is the latest release of the wmcb
 	latestRelease *github.RepositoryRelease
-	// latestCniPluginsVersion is the latest 0.8.x version of CNI Plugins
+	// LatestCniPluginsVersion is the latest 0.8.x version of CNI Plugins
 	LatestCniPluginsVersion string
 }
 
@@ -173,8 +174,11 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
 	if err := f.getClusterVersion(); err != nil {
 		return fmt.Errorf("unable to get OpenShift cluster version: %v", err)
 	}
-	if err := f.getLatestGithubRelease(); err != nil {
-		return fmt.Errorf("unable to get latest github release: %v", err)
+	if err := f.getLatestWMCBRelease(); err != nil {
+		return fmt.Errorf("unable to get latest WMCB release: %v", err)
+	}
+	if err := f.getLatestCniPluginsVersion(); err != nil {
+		return fmt.Errorf("unable to get latest 0.8.x version of CNI Plugins: %v", err)
 	}
 	return nil
 }
@@ -238,8 +242,31 @@ func (f *TestFramework) getClusterVersion() error {
 	return nil
 }
 
-// GetGithubReleases gets all the github releases for a given owner and repo
-func (f *TestFramework) GetGithubReleases(owner string, repo string) ([]*github.RepositoryRelease, error) {
+// getLatestCniPluginsVersion returns the latest 0.8.x version of CNI plugins in 'v<major>.<minor>.<patch>' format
+func (f *TestFramework) getLatestCniPluginsVersion() error {
+	releases, err := f.getGithubReleases("containernetworking", "plugins")
+	if err != nil {
+		return err
+	}
+	// Iterating over releases to fetch versions from tag names
+	var cniPluginsVersions = []string{}
+	for _, release := range releases {
+		cniPluginsVersions = append(cniPluginsVersions, release.GetTagName())
+	}
+	// Sorting versions in reverse order so as to get latest version first
+	sort.Sort(sort.Reverse(sort.StringSlice(cniPluginsVersions)))
+	// Iterating over versions to find first 0.8.x version which is not a release candidate
+	for _, cniPluginsVersion := range cniPluginsVersions {
+		if strings.HasPrefix(cniPluginsVersion, "v0.8.") && !strings.Contains(cniPluginsVersion, "rc") {
+			f.LatestCniPluginsVersion = cniPluginsVersion
+			return nil
+		}
+	}
+	return fmt.Errorf("could not fetch latest 0.8.x version of CNI Plugins")
+}
+
+// getGithubReleases gets all the github releases for a given owner and repo
+func (f *TestFramework) getGithubReleases(owner string, repo string) ([]*github.RepositoryRelease, error) {
 	// Initializing a client for using the Github API
 	client := github.NewClient(nil)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
@@ -249,10 +276,10 @@ func (f *TestFramework) GetGithubReleases(owner string, repo string) ([]*github.
 	return releases, err
 }
 
-// getLatestGithubRelease gets the latest github release for the wmcb repo. This release is specific to the cluster
+// getLatestWMCBRelease gets the latest github release for the wmcb repo. This release is specific to the cluster
 // version
-func (f *TestFramework) getLatestGithubRelease() error {
-	releases, err := f.GetGithubReleases("openshift", "windows-machine-config-bootstrapper")
+func (f *TestFramework) getLatestWMCBRelease() error {
+	releases, err := f.getGithubReleases("openshift", "windows-machine-config-bootstrapper")
 	if err != nil {
 		return err
 	}
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -1,15 +1,12 @@
 package framework
 
 import (
-	"bufio"
 	"context"
 	"fmt"
 	"io/ioutil"
 	"log"
-	"net/http"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -40,8 +37,6 @@ const (
 	// remoteLogPath is the directory where all the log files related to components that we need are generated on the
 	// Windows VM
 	remoteLogPath = "C:\\k\\log\\"
-	// cniPluginsBaseURL is the base URL of the CNI Plugins location
-	cniPluginsBaseURL = "https://github.com/containernetworking/plugins/releases/download/"
 )
 
 var (
@@ -181,9 +176,6 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
 	if err := f.getLatestGithubRelease(); err != nil {
 		return fmt.Errorf("unable to get latest github release: %v", err)
 	}
-	if err := f.getLatestCniPluginsVersion(); err != nil {
-		return fmt.Errorf("unable to get latest 0.8.x version of CNI Plugins: %v", err)
-	}
 	return nil
 }
 
@@ -246,15 +238,21 @@ func (f *TestFramework) getClusterVersion() error {
 	return nil
 }
 
+// GetGithubReleases gets all the github releases for a given owner and repo
+func (f *TestFramework) GetGithubReleases(owner string, repo string) ([]*github.RepositoryRelease, error) {
+	// Initializing a client for using the Github API
+	client := github.NewClient(nil)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
+	defer cancel()
+	releases, _, err := client.Repositories.ListReleases(ctx, owner, repo,
+		&github.ListOptions{})
+	return releases, err
+}
+
 // getLatestGithubRelease gets the latest github release for the wmcb repo. This release is specific to the cluster
 // version
 func (f *TestFramework) getLatestGithubRelease() error {
-	client := github.NewClient(nil)
-
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
-	defer cancel()
-	releases, _, err := client.Repositories.ListReleases(ctx, "openshift", "windows-machine-config-bootstrapper",
-		&github.ListOptions{})
+	releases, err := f.GetGithubReleases("openshift", "windows-machine-config-bootstrapper")
 	if err != nil {
 		return err
 	}
@@ -292,67 +290,6 @@ func (f *TestFramework) GetReleaseArtifactSHA(artifactName string) (string, erro
 		}
 	}
 	return "", fmt.Errorf("no artifact with name %s", artifactName)
-}
-
-// getLatestCniPluginsVersion returns the latest 0.8.x version of CNI plugins in 'v<major>.<minor>.<patch>' format
-func (f *TestFramework) getLatestCniPluginsVersion() error {
-	client := github.NewClient(nil)
-
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
-	defer cancel()
-	releases, _, err := client.Repositories.ListReleases(ctx, "containernetworking", "plugins",
-		&github.ListOptions{})
-	if err != nil {
-		return err
-	}
-	// Iterating over releases to fetch versions from tag names
-	var cniPluginsVersions = []string{}
-	for _, release := range releases {
-		cniPluginsVersions = append(cniPluginsVersions, release.GetTagName())
-	}
-	// Sorting versions in reverse order so as to get latest version first
-	sort.Sort(sort.Reverse(sort.StringSlice(cniPluginsVersions)))
-	// Iterating over versions to find first 0.8.x version which is not a release candidate
-	for _, cniPluginsVersion := range cniPluginsVersions {
-		if strings.HasPrefix(cniPluginsVersion, "v0.8.") && !strings.Contains(cniPluginsVersion, "rc") {
-			f.LatestCniPluginsVersion = cniPluginsVersion
-			return nil
-		}
-	}
-	return fmt.Errorf("could not fetch latest 0.8.x version of CNI Plugins")
-}
-
-// GetLatestCniPluginsURL returns the URL of the latest 0.8.x version of CNI Plugins
-func (f *TestFramework) GetLatestCniPluginsURL() (string, error) {
-	latestCniPluginsURL := cniPluginsBaseURL + f.LatestCniPluginsVersion + "/cni-plugins-windows-amd64-" +
-		f.LatestCniPluginsVersion + ".tgz"
-	return latestCniPluginsURL, nil
-}
-
-// GetLatestCniPluginsSHA returns the SHA512 of the latest 0.8.x version of CNI Plugins
-func (f *TestFramework) GetLatestCniPluginsSHA() (string, error) {
-	latestCniPluginsURL, err := f.GetLatestCniPluginsURL()
-	if err != nil {
-		return "", fmt.Errorf("could not fetch latest CNI Plugins URL: %s", err)
-	}
-
-	latestCniPluginsChecksumFileURL := latestCniPluginsURL + ".sha512"
-	response, err := http.Get(latestCniPluginsChecksumFileURL)
-	if err != nil {
-		return "", fmt.Errorf("could not fetch CNI Plugins checksum file: %s", err)
-	}
-	defer response.Body.Close()
-
-	var checksumFileContent string
-	// Fetching checksum file content from the GET Response
-	scanner := bufio.NewScanner(response.Body)
-	for scanner.Scan() {
-		checksumFileContent = scanner.Text()
-	}
-	// The checksum file content is in the format "<sha> <filename>". So to get SHA we need to extract only the <sha>
-	// from the file
-	sha512 := strings.Split(checksumFileContent, " ")[0]
-	return sha512, nil
 }
 
 // GetNode returns a pointer to the node object associated with the external IP provided
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -1,12 +1,15 @@
 package framework
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -37,6 +40,8 @@ const (
 	// remoteLogPath is the directory where all the log files related to components that we need are generated on the
 	// Windows VM
 	remoteLogPath = "C:\\k\\log\\"
+	// cniPluginsBaseURL is the base URL of the CNI Plugins location
+	cniPluginsBaseURL = "https://github.com/containernetworking/plugins/releases/download/"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


tools/ansible/tasks/wsu/main.yaml
```diff
@@ -191,16 +206,36 @@
...
+        - name: Download checksum file for cni plugins
+          get_url:
+            url: "{{ cni_plugins_location }}.sha512"
+            dest: "{{ tmp_dir.path }}/cni_plugins_checksum_file"
+
+        # The checksum file content is in the format "<sha> <filename>". So to get SHA we need to extract only the <sha>
+        # from the file
+        - name: Get the SHA for cni plugins
+          shell: |
+            cat {{ tmp_dir.path }}/cni_plugins_checksum_file | awk -F ' ' '{ print $1 }'
```
sebsoto: Cant you just do awk '{print $1}'?
```
[ssoto@localhost aws]$ cat /tmp/sha | awk '{print $1}'
5583602180a99f7cb5ac66b5a80d97ae08845a2a71e7338c4941c55f383c69b2a63fd1b11feb585f155cf0b8507d3fb0de1b3acea39c38f5873944bf475cf9ec
```

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -191,16 +206,36 @@
...
+            dest: "{{ tmp_dir.path }}/cni_plugins.tgz"
+            checksum: "sha512:{{ cni_plugins_sha.stdout }}"
+
         - name: Create temporary cni directory
           file:
             path: "{{ tmp_dir.path }}/cni"
             state: directory
 
-        - name: Get cni plugins
+        - name: Extract cni plugins
```
sebsoto: I think we will want to get rid of the .tgz. Leaving it results in the .tgz being copied to the Windows instance, even though its not used.

aravindhp: Nice catch, @sebsoto

akhil-rane: Aren't we doing the same with the kube node binaries?

aravindhp: If that is case then we need to fix that and not repeat that bad pattern.

internal/test/framework/framework.go
```diff
@@ -284,6 +294,67 @@ func (f *TestFramework) GetReleaseArtifactSHA(artifactName string) (string, erro
 	return "", fmt.Errorf("no artifact with name %s", artifactName)
 }
 
+// getLatestCniPluginsVersion returns the latest 0.8.x version of CNI plugins in 'v<major>.<minor>.<patch>' format
+func (f *TestFramework) getLatestCniPluginsVersion() error {
+	client := github.NewClient(nil)
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I see some common code between this and `getLatestGitHubRelease()` method? Do you want to refactor this?

internal/test/framework/framework.go
```diff
@@ -284,6 +294,67 @@ func (f *TestFramework) GetReleaseArtifactSHA(artifactName string) (string, erro
...
+
+// GetLatestCniPluginsURL returns the URL of the latest 0.8.x version of CNI Plugins
+func (f *TestFramework) GetLatestCniPluginsURL() (string, error) {
+	latestCniPluginsURL := cniPluginsBaseURL + f.LatestCniPluginsVersion + "/cni-plugins-windows-amd64-" +
+		f.LatestCniPluginsVersion + ".tgz"
+	return latestCniPluginsURL, nil
+}
+
+// GetLatestCniPluginsSHA returns the SHA512 of the latest 0.8.x version of CNI Plugins
+func (f *TestFramework) GetLatestCniPluginsSHA() (string, error) {
```
ravisantoshgudimetla: Same as above? Can you refactor this if possible?

internal/test/wmcb/wmcb_test.go
```diff
@@ -88,9 +84,29 @@ type pkgInfo struct {
 	shaType string
 }
 
+func initializeCniPluginsPkgInfo() error {
+	cniPluginsURL, err := framework.GetLatestCniPluginsURL()
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I think, we need to get the package from the called function in framework instead of populating `pkg` structure here or atleast some method that takes the inputs from the called function and populates the pkg structure.

internal/test/wsu/wsu_test.go
```diff
@@ -328,6 +331,43 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.TestWindowsVM, ansibleTe
...
+	// Fetching names of CNI Plugins as a comma separated string
+	// Example: flannel,host-local,win-bridge,win-overlay
+	pluginNamesCommaSeparated, _, err := vm.Run("(gci -FILE "+cniBinaryDir+").basename -join ','", true)
+	pluginNames := strings.Split(pluginNamesCommaSeparated, ",")
+	// Executing each CNI Plugin to check if it is of latest 0.8.x version
+	for _, pluginName := range pluginNames {
+		// Example stderr after executing plugin executable: CNI win-overlay plugin v0.8.2
+		_, pluginExecutionStderr, err := vm.Run(cniBinaryDir+"\\"+pluginName, true)
+		require.NoError(t, err, "Could not execute CNI Plugin "+pluginName)
+		assert.Contains(t, pluginExecutionStderr, framework.LatestCniPluginsVersion, "CNI Plugin "+
```
ravisantoshgudimetla: So, we'll have this in stderr in case of failure?

internal/test/wsu/wsu_test.go
```diff
@@ -328,6 +331,43 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.TestWindowsVM, ansibleTe
...
 
+// testCNIPluginsVersion tests if the kubelet on the Windows VM is running with latest 0.8.x version of CNI Plugins
+func testCNIPluginsVersion(t *testing.T, vm e2ef.TestWindowsVM) {
+	// Fetching kubelet service image path which includes the  directory path of the CNI Plugin binaries
+	// Example kubelet service image path: c:\k\kubelet.exe --windows-service --bootstrap-kubeconfig=c:\k\bootstrap-kubeconfig
+	// --kubeconfig=c:\k\kubeconfig --network-plugin=cni --cni-bin-dir=c:\k\cni --log-file=c:\k\log\kubelet.log
+	// --node-labels=node.openshift.io/os_id=Windows --cloud-provider=aws --resolv-conf="" --config=c:\k\kubelet.conf
+	// --pod-infra-container-image=mcr.microsoft.com/k8s/core/pause:1.2.0 --logtostderr=false --v=3
+	// --cert-dir=c:\var\lib\kubelet\pki\ --register-with-taints=os=Windows:NoSchedule --cni-conf-dir=c:\k\cni\config
+	kubeletImagePath, _, err := vm.Run("$service=get-wmiobject -query \\\"select * from win32_service "+
```
ravisantoshgudimetla: Good job on getting the kubelet service from the remote  host

internal/test/wsu/wsu_test.go
```diff
@@ -328,6 +331,43 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.TestWindowsVM, ansibleTe
...
+	for _, kubeletArg := range kubeletArgs {
+		// if kubelet arg is of the form --<option>=<value> and option='--cni-bn-dir' then its value will be the required
+		// directory path of the CNI Plugin binaries
+		kubeletArgOptionAndValue := strings.Split(kubeletArg, "=")
+		if len(kubeletArgOptionAndValue) == 2 && kubeletArgOptionAndValue[0] == "--cni-bin-dir" {
+			cniBinaryDir = kubeletArgOptionAndValue[1]
+			break
+		}
+	}
+	// Fetching names of CNI Plugins as a comma separated string
```
ravisantoshgudimetla: Shouldn't we fail early when CNI plugin is not enabled?

internal/test/wsu/wsu_test.go
akhil-rane: When we execute CNI Plugin binary, its 'version' is returned in stderr. Nothing is returned in stdout.

internal/test/wmcb/wmcb_test.go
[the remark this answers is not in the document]
akhil-rane: @ravisantoshgudimetla I do not understand how the current method is different from what you are suggesting. The method initializeCniPluginsPkgInfo() takes inputs from framework.GetLatestCniPluginsURL() and framework.GetLatestCniPluginsSHA() functions and populates the pkg structure

internal/test/framework/framework.go
akhil-rane: I cannot see any code getting duplicated here

tools/ansible/tasks/wsu/main.yaml
akhil-rane: done

tools/ansible/tasks/wsu/main.yaml
akhil-rane: Created a playbook task 'Remove cni plugins tar' to address this

internal/test/framework/framework.go
```diff
@@ -284,6 +294,67 @@ func (f *TestFramework) GetReleaseArtifactSHA(artifactName string) (string, erro
...
+		}
+	}
+	return fmt.Errorf("could not fetch latest 0.8.x version of CNI Plugins")
+}
+
+// GetLatestCniPluginsURL returns the URL of the latest 0.8.x version of CNI Plugins
+func (f *TestFramework) GetLatestCniPluginsURL() (string, error) {
+	latestCniPluginsURL := cniPluginsBaseURL + f.LatestCniPluginsVersion + "/cni-plugins-windows-amd64-" +
+		f.LatestCniPluginsVersion + ".tgz"
+	return latestCniPluginsURL, nil
```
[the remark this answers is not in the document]
akhil-rane: Removed error type from function returns, This function is now only returning string

internal/test/framework/framework.go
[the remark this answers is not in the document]
akhil-rane: @mansikulkarni96 Added doc. @ravisantoshgudimetla Added function GetGithubReleases to avoid code duplication

internal/test/wsu/wsu_test.go
akhil-rane: Done

internal/test/wmcb/wmcb_test.go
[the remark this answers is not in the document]
akhil-rane: > This function will not return any errors and might set nil value for latestPluginURL, maybe call a function that can return error?
> Discussed this and would like another opinion @suhanime ?

latestCNIPluginURL value will never be nil so I don' really think we need a separate function to return error. @ravisantoshgudimetla any thoughts on this?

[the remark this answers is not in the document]
akhil-rane: > @ravisantoshgudimetla I do not understand how the current method is different from what you are suggesting. The method initializeCniPluginsPkgInfo() takes inputs from framework.GetLatestCniPluginsURL() and framework.GetLatestCniPluginsSHA() functions and populates the pkg structure

@ravisantoshgudimetla Addressed this as per our discussions

internal/test/wmcb/main_test.go
```diff
@@ -11,7 +11,8 @@ import (
 
 // framework holds the instantiation of test suite being executed. As of now, temp dir is hardcoded.
 var (
-	framework = &e2ef.TestFramework{}
+	// Initialize wmcbFramework which specializes TestFramework by adding some properties specific to WMCB tests
+	framework = wmcbFramework{}
```
ravisantoshgudimetla: Good job on the composition.

internal/test/wmcb/wmcb_test.go
```diff
@@ -84,28 +90,37 @@ type pkgInfo struct {
...
+		return fmt.Errorf("unable to initialize CNI Plugins package info: %v", err)
 	}
-	cniPlugins.url = cniPluginsURL
 
-	cniPluginsSHA, err := framework.GetLatestCniPluginsSHA()
+	// Set up the framework
+	err := f.TestFramework.Setup(vmCount, credentials, skipVMsetup)
 	if err != nil {
-		return fmt.Errorf("unable to get SHA for CNI Plugins: %v", err)
+		fmt.Errorf("framework setup failed: %v", err)
```
mansikulkarni96: any specific reason for not returning here?

internal/test/wmcb/wmcb_test.go
```diff
@@ -88,9 +90,38 @@ type pkgInfo struct {
 	shaType string
 }
 
+type wmcbFramework struct {
+	// TestFramework holds the instantiation of test suite being executed
+	*e2ef.TestFramework
+	// latestCniPluginsVersion is the latest 0.8.x version of CNI Plugins
+	LatestCniPluginsVersion string
```
ravisantoshgudimetla: why is this a public field?

internal/test/wmcb/wmcb_test.go
```diff
@@ -614,3 +645,69 @@ func testWMCBCluster(t *testing.T) {
...
+		if strings.HasPrefix(cniPluginsVersion, "v0.8.") && !strings.Contains(cniPluginsVersion, "rc") {
+			f.LatestCniPluginsVersion = cniPluginsVersion
+			return nil
+		}
+	}
+	return fmt.Errorf("could not fetch latest 0.8.x version of CNI Plugins")
+}
+
+// GetLatestCniPluginsURL returns the URL of the latest 0.8.x version of CNI Plugins
+func (f *wmcbFramework) GetLatestCniPluginsURL() string {
```
ravisantoshgudimetla: Why do we need a method for this and a public one

internal/test/wmcb/wmcb_test.go
ravisantoshgudimetla: nit: You could've directly added package slice here.

internal/test/wmcb/wmcb_test.go
ravisantoshgudimetla: nit and purely for fun comment: Please start using variables and we're not in a purely functional language like racket or scheme

internal/test/wsu/wsu_test.go
ravisantoshgudimetla: Interesting.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -191,16 +206,42 @@
             src: "{{ tmp_dir.path }}/kubernetes/node/bin/kube-proxy.exe"
             dest: "{{ tmp_dir.path }}/kube-proxy.exe"
 
+        - name: Download checksum file for cni plugins
+          get_url:
+            url: "{{ cni_plugins_location }}.sha512"
```
ravisantoshgudimetla: Why should we download the checksum url and then do integrity check? Why can we use checksum directly on url like this:
`sha256:http://example.com/path/sha256sum.txt` given here https://docs.ansible.com/ansible/latest/modules/get_url_module.html#parameter-checksum

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -191,16 +206,42 @@
...
 
-        - name: Get cni plugins
+        - name: Extract cni plugins
           unarchive:
-            src: "https://github.com/containernetworking/plugins/releases/download/v0.8.2/cni-plugins-windows-amd64-v0.8.2.tgz"
+            src: "{{ tmp_dir.path }}/cni_plugins.tgz"
             dest: "{{ tmp_dir.path }}/cni"
-            remote_src: yes
+
+        # Removing cni_plugins.tgz from temporary directory as we do not want to copy it to Windows VM
```
ravisantoshgudimetla: +1

internal/test/framework/framework.go
```diff
@@ -236,15 +238,21 @@ func (f *TestFramework) getClusterVersion() error {
...
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
 	defer cancel()
-	releases, _, err := client.Repositories.ListReleases(ctx, "openshift", "windows-machine-config-bootstrapper",
+	releases, _, err := client.Repositories.ListReleases(ctx, owner, repo,
 		&github.ListOptions{})
+	return releases, err
+}
+
+// getLatestGithubRelease gets the latest github release for the wmcb repo. This release is specific to the cluster
+// version
```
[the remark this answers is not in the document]
akhil-rane: This function was not introduced as part of this PR. I am not sure if I should change it.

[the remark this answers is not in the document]
sebsoto: You're changing how the function works. I agree that you should change it to `getLatestWMCBRelease`

internal/test/wmcb/wmcb_test.go
sebsoto: Yeah you want to do `return fmt.Errorf("framework setup failed: %v", err)`

internal/test/wsu/wsu_test.go
```diff
@@ -339,6 +342,51 @@ func testCNIConfig(t *testing.T, node *v1.Node, vm e2ef.TestWindowsVM, ansibleTe
...
+			err := fmt.Errorf("kubelet arg '--network-plugin' should have value 'cni', current value is '%s'",
+				kubeletArgOptionAndValue[1])
+			require.NoError(t, err, "CNI Plugins are not enabled")
+		}
+		// if kubelet arg is of the form --<option>=<value> and option='--cni-bin-dir' then its value will be the required
+		// directory path of the CNI Plugin binaries
+		if len(kubeletArgOptionAndValue) == 2 && kubeletArgOptionAndValue[0] == "--cni-bin-dir" {
+			cniBinaryDir = kubeletArgOptionAndValue[1]
+		}
+	}
```
sebsoto: Seems like you should require `cniBinaryDir` is not empty after this loop

internal/test/wmcb/wmcb_test.go
```diff
@@ -614,3 +645,69 @@ func testWMCBCluster(t *testing.T) {
...
+	latestCniPluginsURL := cniPluginsBaseURL + f.LatestCniPluginsVersion + "/cni-plugins-windows-amd64-" +
+		f.LatestCniPluginsVersion + ".tgz"
+	return latestCniPluginsURL
+}
+
+// GetLatestCniPluginsSHA returns the SHA512 of the latest 0.8.x version of CNI Plugins
+func (f *wmcbFramework) GetLatestCniPluginsSHA() (string, error) {
```
sebsoto: This method does not need to be public

sebsoto: It is only used within a private method

internal/test/wmcb/wmcb_test.go
```diff
@@ -614,3 +645,69 @@ func testWMCBCluster(t *testing.T) {
...
+
+	var checksumFileContent string
+	// Fetching checksum file content from the GET Response
+	scanner := bufio.NewScanner(response.Body)
+	for scanner.Scan() {
+		checksumFileContent = scanner.Text()
+	}
+	// The checksum file content is in the format "<sha> <filename>". So to get SHA we need to extract only the <sha>
+	// from the file
+	sha512 := strings.Split(checksumFileContent, " ")[0]
```
sebsoto: Return error if len(strings.Split(checksumFileContent, " ")) <2

akhil-rane: done

internal/test/wmcb/wmcb_test.go
akhil-rane: done

internal/test/wsu/wsu_test.go
akhil-rane: done

tools/ansible/tasks/wsu/main.yaml
akhil-rane: I did not know about this. Fixed it

internal/test/wmcb/wmcb_test.go
akhil-rane: done

internal/test/wmcb/wmcb_test.go
akhil-rane: Removed this function

internal/test/wmcb/wmcb_test.go
akhil-rane: Moved it to framework.go as it is used both in wmcb_test.go and wsu_test.go

internal/test/wmcb/main_test.go
akhil-rane: Thanx

internal/test/framework/framework.go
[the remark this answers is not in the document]
akhil-rane: done

internal/test/framework/framework.go
```diff
@@ -69,6 +69,8 @@ type TestFramework struct {
 	ClusterVersion string
 	// latestRelease is the latest release of the wmcb
 	latestRelease *github.RepositoryRelease
+	// latestCniPluginsVersion is the latest 0.8.x version of CNI Plugins
```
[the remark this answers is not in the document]
akhil-rane: It is used both in wmcb_test.go and wsu_test.go

internal/test/wmcb/wmcb_test.go
akhil-rane: Nice catch. Fixed it

internal/test/framework/framework.go
```diff
@@ -236,15 +242,44 @@ func (f *TestFramework) getClusterVersion() error {
...
+		if strings.HasPrefix(cniPluginsVersion, "v0.8.") && !strings.Contains(cniPluginsVersion, "rc") {
+			f.LatestCniPluginsVersion = cniPluginsVersion
+			return nil
+		}
+	}
+	return fmt.Errorf("could not fetch latest 0.8.x version of CNI Plugins")
+}
 
+// GetGithubReleases gets all the github releases for a given owner and repo
+func (f *TestFramework) GetGithubReleases(owner string, repo string) ([]*github.RepositoryRelease, error) {
```
ravisantoshgudimetla: Why is this method public? I see it being used in framework only..

internal/test/wmcb/wmcb_test.go
```diff
@@ -88,9 +87,31 @@ type pkgInfo struct {
...
+	if err := f.initializeCniPluginsPkgInfo(); err != nil {
+		return fmt.Errorf("unable to initialize CNI Plugins package info: %v", err)
+	}
+	return nil
+}
+
 // TestWMCB runs the unit and e2e tests for WMCB on the remote VMs
 func TestWMCB(t *testing.T) {
 	remoteDir := "C:\\Temp"
+
```
ravisantoshgudimetla: unncessary new line addition

internal/test/framework/framework.go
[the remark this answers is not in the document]
ravisantoshgudimetla: Ok then please make the comment to have `LatestCniPluginsVersion`

internal/test/framework/framework.go
akhil-rane: Made it private

internal/test/wmcb/wmcb_test.go
akhil-rane: removed it

ravisantoshgudimetla:
Forgot to ask squash before CI bot merged :(.

PR #164, opened 2020-03-09
Rename copyFile function

mansikulkarni96:
The purpose of this PR is to rename the copyFile function to copyFileTo which copies from local to remote. This change was necessary to make the WindowsVM interface clear for further implementations, with the addition of a new function copyFileFrom in PR #162 that copies file from remote to local.

### DIFF
--- a/internal/test/framework/windowsvm.go
+++ b/internal/test/framework/windowsvm.go
@@ -35,6 +35,9 @@ type testWindowsVM struct {
 // TestWindowsVM is the interface for interacting with a Windows VM in the test framework. This will hold the
 // specialized information related to test suite
 type TestWindowsVM interface {
+	// CopyFileTo copies the given file to the remote directory in the Windows VM. The remote directory is created if it
+	// does not exist
+	CopyFileTo(string, string) error
 	// RetrieveFiles retrieves the list of file from the directory in the remote Windows VM to the local host. As of
 	// now, we're limiting every file in the remote directory to be written to single directory on the local host
 	RetrieveFiles(string, string) error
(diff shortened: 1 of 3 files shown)


internal/test/framework/windowsvm.go
```diff
@@ -35,6 +35,9 @@ type testWindowsVM struct {
 // TestWindowsVM is the interface for interacting with a Windows VM in the test framework. This will hold the
 // specialized information related to test suite
 type TestWindowsVM interface {
+	// CopyFileTo copies the given file to the remote directory in the Windows VM. The remote directory is created if it
+	// does not exist
+	CopyFileTo(string, string) error
```
sebsoto: This method is already present  in types.WindowsVM, so you shouldn't need to add it here

suhanime: How is this func different from WNI's `CopyFile`?

internal/test/framework/windowsvm.go
```diff
@@ -104,6 +107,43 @@ func newWindowsVM(imageID, instanceType string, credentials *types.Credentials,
 	return w, nil
 }
 
+func (w *testWindowsVM) CopyFileTo(filePath, remoteDir string) error {
```
suhanime: Actually the implementation looks to be the same as `WindowsVM` interface's `CopyFile`. Why do we have 2 separate implementations of the same func?

sebsoto: Is this the same as windowsVM.CopyFileTo?
If so, we shouldn't be duplicating code. You should just call the `Windows` implementation of CopyFileTo.

sebsoto:
You will need 3 commits and 2 PRs minimum.
PR 1:
commit 1: Rename Windows.CopyFile() to CopyFileTo
PR 2:
commit 1: Update go.mod to point to new WNI. Update usage of CopyFile() to CopyFileTo()
commit 2: Capture hybrid-overlay and kube-proxy logs
PR 2 commit 1+2 could be broken down into multiple PR's, but I don't have too much of an opinion on that. Other people might.

mansikulkarni96:
closing this PR as copyFile rename is not required. It is a part of WindowsVM interface now and not a part of framework TestWindowsVM intercace which was the reason behind this rename.

PR #167, opened 2020-03-18
Bug 1809291: [wsu] Ensure kube-proxy service is not running.

mansikulkarni96:
This PR resolves the access to kube-proxy error encountered while running WSU
playbook multiple times accross the same set of VM's

Problem: WSU playbook fails to run in TASK [Copy kube-proxy to C:\k] with the error
"Access to the path 'C:\\k\\kube-proxy.exe' is denied".

Solution: This is happening because kube-proxy is in use. This is fixed by moving
the `Ensure kube-proxy Windows Service is not running` to execute before the
`Copy kube-proxy to C:\k` task.

Bug: [1809291](https://bugzilla.redhat.com/show_bug.cgi?id=1809291)

### DIFF
--- a/tools/ansible/tasks/wsu/main.yaml
+++ b/tools/ansible/tasks/wsu/main.yaml
@@ -430,6 +430,11 @@
         msg: "CNI Configuration error"
       when: '"CNI configuration completed successfully" not in bootstrap_out.stderr'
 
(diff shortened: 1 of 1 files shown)


suhanime:
lgtm

PR #174, opened 2020-04-08
Bug 1821429: [wsu] Fix getting kube node binary SHA

mansikulkarni96:
This PR fixes the shell command for extracting the kubernetes version without patch version for any
intermediate releases eg: v1.18.0-rc.1 should be extracted as 1.18 instead of 1.18.0-rc for forming
the changelog URL.
In addition it also makes the changes necessary in go.mod, verify-gofmt.sh and Dockerfile to update golang version from 1.12 to 1.13. 
Since the golang version has been updated in CI we need to unset the GOFLAGS environment variable where it is set to "-mod=vendor" value to avoid vendoring in CI environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,11 @@ GO_BUILD_ARGS=CGO_ENABLED=0 GO111MODULE=on
 # TODO (suhanime): Pin go versions and lint
 # TODO (suhanime): Enable linting when we don't have unimplemented methods
 
+# The golang 1.13 image used in CI enforces vendoring. Workaround that by unsetting it.
+ifeq ($(GOFLAGS), -mod=vendor)
+	unexport GOFLAGS
+endif
+
 .PHONY: build
 build:
 	$(GO_BUILD_ARGS) GOOS=windows go build -o wmcb.exe  $(MAIN_PACKAGE)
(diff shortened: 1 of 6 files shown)


mansikulkarni96:
@aravindhp this works for any intermediate as well as standard releases.

Makefile
```diff
@@ -10,6 +10,11 @@ GO_BUILD_ARGS=CGO_ENABLED=0 GO111MODULE=on
 # TODO (suhanime): Pin go versions and lint
 # TODO (suhanime): Enable linting when we don't have unimplemented methods
 
+# The golang 1.13 image used in CI enforces vendoring. Workaround that by unsetting it.
+ifeq ($(GOFLAGS), -mod=vendor)
```
sebsoto, line 14: Can goflags include other flags, not just the vendor flag? If so, I don't think that will  be caught here. In that case you will need to keep the other flags intact, removing only the vendor flag

aravindhp: @ravisantoshgudimetla and I had the same discussion with @mansikulkarni96 today. The ART team recommends doing it this way in CI. PS: we have done this in a few places now without any issues.

mansikulkarni96: Mentioning the discussion commit [here](https://github.com/openshift/release/pull/6337/commits/db916acf9afac0b1dd883a159a80696132f244e7) for future reference.

ravisantoshgudimetla:
I see that wsu-test has failed. @mansikulkarni96 ^

PR #176, opened 2020-04-15
Bug 1823919: [ci] React to upstream kube-ovn changes

ravisantoshgudimetla:
As of now, the ovn-kube network already comes with hybrid overlay enabled. So, we don't need to patch the network operator as the cluster comes up with hybrid overlay enabled. While patching we previously used to wait till all the nodes have hybrid-overlay specific annotations. In the new OVN world, the linux nodes won't have those annotations present, so the test was failing for us. This commit
- Removes the test which fails if we have don't those annotations
  on master nodes
- Removes patching the network operator object
- Removes the test which checks for the WSU failure when hybrid overlay
  is not enabled

### DIFF
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -16,11 +16,9 @@ import (
 	"github.com/google/go-github/v29/github"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	operatorv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
-	"github.com/openshift/windows-machine-config-bootstrapper/internal/test"
 	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/types"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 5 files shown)


ravisantoshgudimetla:
Looks like winwebserver  container is dying - `I0427 23:26:17.468115    2412 kubelet.go:1952] SyncLoop (PLEG): "win-curler-i-042c086b306cd560d-job-2hs4n_default(f631f142-8134-4824-be0b-79789c2b8fa2)", event: &pleg.PodLifecycleEvent{ID:"f631f142-8134-4824-be0b-79789c2b8fa2", Type:"ContainerDied", Data:"84dc2debebe0e66a5b40e8e21a6aae924001a7a59a2de9b958b157b4cf8bd6cd"}`

ravisantoshgudimetla:
/approve cancel

I am bringing a nightly to see if our changes are present. Looks like kubelet is failing to run the pods on the Windows node.

aravindhp:
@ravisantoshgudimetla please remove `testWithoutHybridOverlay` and we should be able to complete reacting to upstream hybrid overlay changes and step-registry (with #195)

aravindhp:
@ravisantoshgudimetla thinking a bit more, removing `testWithoutHybridOverlay` should be done in the #195 and I have taken care of it.

ravisantoshgudimetla:
Just to see if it's aa flake:

/retest

ravisantoshgudimetla:
Failed twice:

Let's see if it works at all. For sure this passed once yesterday :)

/retest

PR #177, opened 2020-04-22
[wni] put azure vm behind worker nodes' lb

PratikMahajan:
In Azure, for egress as well as for Loadbalancer service to work,
the windows vm is expected to be placed behind the worker nodes'
load balancer.

this commit puts the azure windows vm behind the worker node loadbalancer
and changes the sku for ip address from basic to standard in order to
successfully carry out the operation

We are using a Standard SKU in place of a Basic SKU as the loadbalancer
and IP SKU should match
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-ip-addresses-overview-arm#sku

Also, this PR prints the public and private ip address
for azure and aws VMs to the console

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
@@ -55,6 +55,10 @@ const (
 	sshRulePriority = 603
 	// sshRuleName is the security group rule name for the RDP rule
 	sshRuleName = "SSH"
+	// SKU for ip address
+	// We are using a Standard SKU in place of a Basic SKU as the Load balancer and IP SKU should match
+	// https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-ip-addresses-overview-arm#sku
+	ipSKU = "Standard"
 )
 
 var windowsWorker string = "winworker-"
@@ -128,8 +132,13 @@ func New(openShiftClient *client.OpenShift, credentialPath, subscriptionID,
 	if errorCheck(err) {
 		return nil, err
 	}
-
-	infraID, _ := openShiftClient.GetInfrastructureID()
+	if subscriptionID == "" {
+		return nil, fmt.Errorf("empty subscriptionID for azure")
+	}
+	infraID, err := openShiftClient.GetInfrastructureID()
+	if err != nil {
+		return nil, err
+	}
 	resourceAuthorizer, err := auth.NewAuthorizerFromFileWithResource(azure.PublicCloud.ResourceManagerEndpoint)
 	if errorCheck(err) {
 		return nil, err
@@ -323,6 +332,7 @@ func (az *AzureProvider) createPublicIP(ctx context.Context) (ip *network.Public
 		network.PublicIPAddress{
 			Name:     to.StringPtr(az.IpName),
 			Location: to.StringPtr(nodeLocation),
+			Sku:      &network.PublicIPAddressSku{Name: ipSKU},
 			PublicIPAddressPropertiesFormat: &network.PublicIPAddressPropertiesFormat{
 				PublicIPAddressVersion:   network.IPv4,
 				PublicIPAllocationMethod: network.Static,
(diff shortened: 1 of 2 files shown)


tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -323,6 +323,7 @@ func (az *AzureProvider) createPublicIP(ctx context.Context) (ip *network.Public
 		network.PublicIPAddress{
 			Name:     to.StringPtr(az.IpName),
 			Location: to.StringPtr(nodeLocation),
+			Sku:      &network.PublicIPAddressSku{Name: "Standard"},
```
aravindhp, line 326: Did you confirm that this is the only place where we need to set the SKU to `standard`?

Please add a link to the Azure docs that mention why SKUs need to match. Do this in the code, commit and PR messages.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -464,6 +470,13 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
 	return storageProfile
 }
 
+// getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
+func (az *AzureProvider) getWorkerBackendPoolID() *string {
+	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
+		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
```
aravindhp, line 476: I did not pick up on this when you showed us code yesterday. Are you assuming that index 2 will always be the worker LB? You cannot make that assumption. You need to make a call to get all the LBs in the RG and then pick the one being used by the Linux worker nodes.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -374,6 +375,11 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
 						Subnet:                    &subnet,
 						PrivateIPAllocationMethod: network.Dynamic,
 						PublicIPAddress:           &ip,
+						LoadBalancerBackendAddressPools: &[]network.BackendAddressPool{
```
aravindhp, line 378: Please mention why you need to place it behind the LB here and in the commit and PR message.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: yes. only place where sku needs to be changed.

ravisantoshgudimetla: nit: Make it a constant.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla, line 476: Do you have checks for `az.subscriptionID` and `az.infraID`, can they be nil? What would happen to Loadbalancer field when it has nil value?

PratikMahajan: in that case, we need to add a global check for infraID and subscriptionID. I assumed that they're being checked before assignment. Should i add a global check for infraID and subscriptionID ?

aravindhp:
/approve

@PratikMahajan please post the output of the Azure e2e tests as a comment.

This PR / story is not complete as such and we pushing it through so that the drop doc can be finalized. Please open a follow up PR where you are programmatically picking the worker LB. You should then add an e2e test that checks if the Windows VM created in the e2e test is associated with the worker LB.

PratikMahajan:
wni test output 
```
windows-node-installer wsu-internal-ip 🔒  ✏️ 1❔ 1 go test -run=TestCreateVM ./test/e2e/... -v
=== RUN   TestCreateVM
2020/04/24 15:57:04 kubeconfig source: /home/mahajan/projects/aws-infra/extras/pmahajan-az-01/auth/kubeconfig
=== RUN   TestCreateVM/check_if_required_security_rules_are_present
=== RUN   TestCreateVM/check_if_ansible_is_able_to_ping_on_the_WinRmHttps_port
=== RUN   TestCreateVM/check_if_container_logs_port_is_open_in_Windows_firewall
=== RUN   TestCreateVM/check_if_SSH_connection_is_available
--- PASS: TestCreateVM (17.69s)
    --- PASS: TestCreateVM/check_if_required_security_rules_are_present (0.68s)
    --- PASS: TestCreateVM/check_if_ansible_is_able_to_ping_on_the_WinRmHttps_port (7.60s)
    --- PASS: TestCreateVM/check_if_container_logs_port_is_open_in_Windows_firewall (7.91s)
    --- PASS: TestCreateVM/check_if_SSH_connection_is_available (0.64s)
PASS
ok      github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/test/e2e  17.702s
```

PR #182, opened 2020-04-27
[wni] Update Reinitialize() to handle WinRM client

aravindhp:
We are noticing that commands run using Run() that uses a WinRM client is taking more than 5 minutes to complete after the hybrid-overlay reconfigures the network. Reinitialize the WinRM client in the Reinitialize() method to mitigate this.

Add a "warning" pre-fix to the log message printed on errors encountered when reinitialzing the SSH client to call out that this is not a fatal error.

### DIFF
--- a/tools/windows-node-installer/pkg/types/types.go
+++ b/tools/windows-node-installer/pkg/types/types.go
@@ -157,6 +157,9 @@ func (w *Windows) Reinitialize() error {
 	if err := w.GetSSHClient(); err != nil {
 		return fmt.Errorf("failed to reinitialize ssh client: %v", err)
 	}
+	if err := w.SetupWinRMClient(); err != nil {
+		return fmt.Errorf("failed to reinitialize WinRM client: %v", err)
+	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


tools/windows-node-installer/pkg/types/types.go
```diff
@@ -157,6 +157,9 @@ func (w *Windows) Reinitialize() error {
 	if err := w.GetSSHClient(); err != nil {
 		return fmt.Errorf("failed to reinitialize ssh client: %v", err)
 	}
+	if err := w.SetupWinRMClient(); err != nil {
```
ravisantoshgudimetla, line 160: What happens to existing connection? Shouldn't we close it?

aravindhp: We do not need to close it. [NewClient()](https://godoc.org/github.com/masterzen/winrm#NewClient) that we use to initialize a client does not create a connection. A connection is created and closed with every `Run()` call.

ravisantoshgudimetla: Got it, thanks :)

aravindhp:
/override ci/prow/e2e-wmcb 
/override ci/prow/e2e-wsu 

Overriding the above tests as they need #176 to pass and this PR does not make any WSU and WMCB changes.

PR #186, opened 2020-04-28
[wni] Add programmatic retrieval of backendPoolID

PratikMahajan:
this PR changes the retrieval of backendPoolID to fetch the
backendPoolID in a programmatic way.
also, adds a method to disassociate ip address from NIC as the changes
introduced in the PR broke the deleteVM function

also, added tests to check the backendPoolID of the VMs created

[WINC-319]

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/2019-03-01/resources/mgmt/resources"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-04-01/network"
 	"github.com/Azure/go-autorest/autorest"
@@ -80,6 +81,8 @@ type AzureProvider struct {
 	nsgClient network.SecurityGroupsClient
 	// diskClient to query for disk related operations.
 	diskClient compute.DisksClient
+	// resourcesClient to query resources. Can query all resources
+	resourcesClient resources.Client
 	// a request authorization token to supply for clients
 	authorizer autorest.Authorizer
 	// resourceGroupName of the existing openshift cluster.
@@ -152,6 +155,7 @@ func New(openShiftClient *client.OpenShift, credentialPath, subscriptionID,
 	nsgClient := getNsgClient(resourceAuthorizer, subscriptionID)
 	diskClient := getDiskClient(resourceAuthorizer, subscriptionID)
 	rulesClient := getRulesClient(resourceAuthorizer, subscriptionID)
+	resourcesClient := getResourcesClient(resourceAuthorizer, subscriptionID)
 
 	requiredRules, err := constructRequiredRules(rulesClient, resourceGroupName)
 	if err != nil {
@@ -161,7 +165,7 @@ func New(openShiftClient *client.OpenShift, credentialPath, subscriptionID,
 	var IpName, NicName, NsgName string
 
 	return &AzureProvider{vnetClient, vmClient, ipClient,
-		subnetClient, nicClient, nsgClient, diskClient, resourceAuthorizer,
+		subnetClient, nicClient, nsgClient, diskClient, resourcesClient, resourceAuthorizer,
 		resourceGroupName, subscriptionID, infraID, IpName, NicName, NsgName,
 		imageID, instanceType, resourceTrackerDir, requiredRules}, nil
 }
@@ -241,6 +245,13 @@ func getDiskClient(authorizer autorest.Authorizer, subscriptionID string) comput
 	diskClient := compute.NewDisksClient(subscriptionID)
 	diskClient.Authorizer = authorizer
 	return diskClient
+}
+
+// getResourcesClient gets the resources client by passing the authorizer token.
+func getResourcesClient(authorizer autorest.Authorizer, subscriptionID string) resources.Client {
+	resourcesClient := resources.NewClient(subscriptionID)
+	resourcesClient.Authorizer = authorizer
+	return resourcesClient
 }
 
 // errorCheck checks if there exists an error and returns a bool response
@@ -373,6 +384,13 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
 		fmt.Errorf("failed to get ip address: %v", err)
 	}
 
+	// In Azure, for egress as well as for load balancer service to work on OpenShift, the
+	// Windows VM is expected to be placed behind the worker nodes' load balancer.
+	workerBackendAddressPools, err := az.getWorkerBackendAddressPools(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker node backend pool id: %v", err)
+	}
+
 	nicParams := network.Interface{
 		Name:     to.StringPtr(az.NicName),
 		Location: to.StringPtr(nodeLocation),
@@ -381,16 +399,10 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
 				{
 					Name: to.StringPtr(ipConfigName),
 					InterfaceIPConfigurationPropertiesFormat: &network.InterfaceIPConfigurationPropertiesFormat{
-						Subnet:                    &subnet,
-						PrivateIPAllocationMethod: network.Dynamic,
-						PublicIPAddress:           &ip,
-						// In Azure, for egress as well as for Load balancer service to work, the
-						// windows vm is expected to be placed behind the worker nodes' Load balancer.
-						LoadBalancerBackendAddressPools: &[]network.BackendAddressPool{
-							{
-								ID: az.getWorkerBackendPoolID(),
-							},
-						},
+						Subnet:                          &subnet,
+						PrivateIPAllocationMethod:       network.Dynamic,
+						PublicIPAddress:                 &ip,
+						LoadBalancerBackendAddressPools: &workerBackendAddressPools,
 					},
 				},
 			},
@@ -481,13 +493,145 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
 	return storageProfile
 }
 
-// getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
-func (az *AzureProvider) getWorkerBackendPoolID() *string {
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+// getWorkerBackendAddressPools gets the backend address pools of the worker node loadbalancer
+func (az *AzureProvider) getWorkerBackendAddressPools(ctx context.Context) ([]network.BackendAddressPool, error) {
+	// Create a filter to get a resource of type virtualMachine which contains 'worker' in its name
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name) and " +
+		"resourceGroup eq '" + az.resourceGroupName + "'"
+	// number of resources to retrieve
+	// https://learnk8s.io/kubernetes-node-size
+	// Kubernetes can technically accommodate 5000 nodes, considering that for optimal performance we limit it at 500 worker nodes
+	top := int32(500)
+	// Use getResource() function to get all the workerVMs by filtering VMs which have 'worker' in their name
+	// In the vmClient, they need full vmName. We can use list() but list would return all the VMs in the
+	// resource group, we will then have to filter it to get the worker VMs using go code iterating through
+	// all the VM names. getResource() does all that work for us. We just need to use the output in that case.
+	vmResources, err := az.getResource(ctx, filter, top)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker VMs: %v", err)
+	}
+	if vmResources == nil {
+		return nil, fmt.Errorf("failed to get worker VMs resource")
+	}
+
+	// keep track of workerVM load balancer
+	// using string as we are assuming all the worker VMs are behind a single load balancer
+	var backendAddressPool string
+
+	// Iterate over all the VM resources and get the worker backend pools attached to all the resources.
+	for _, vmResource := range vmResources {
+		if vmResource.ID == nil {
+			return nil, fmt.Errorf("id for VM resource cannot be nil")
+		}
+		vmName := extractResourceName(*vmResource.ID)
+		// get the backend address pool
+		workerVMBackendAddressPool, err := az.getVMBackendAddressPool(ctx, vmName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get backend address pools for vm %s", vmName)
+		}
+		if backendAddressPool == "" {
+			backendAddressPool = workerVMBackendAddressPool
+		} else if backendAddressPool != workerVMBackendAddressPool {
+			return nil, fmt.Errorf("multiple backend address pools for worker nodes is not supported")
+		}
+	}
+
+	// create an array of the backend address pools that are attached to worker nodes
+	// which will then be applied to the Windows node
+	lbBackendPoolAddressIDs := []network.BackendAddressPool{
+		{
+			ID: &backendAddressPool,
+		},
+	}
+
+	return lbBackendPoolAddressIDs, nil
+}
+
+// getVMBackendAddressPool gets the backend address pool associated with the VM
+// we are assuming that all the worker VMs have only one backend address pool associated.
+func (az *AzureProvider) getVMBackendAddressPool(ctx context.Context, vmName string) (string, error) {
+
+	err, nicName := az.getNICname(ctx, vmName)
+	if err != nil {
+		return "", fmt.Errorf("failed to get NIC name %v", err)
+	}
+
+	_, interfaceIPConfigs, err := az.getInfoFromNIC(ctx, nicName)
+	if err != nil {
+		return "", fmt.Errorf("failed to get ip configuration for NIC %s: %v", nicName, err)
+	}
+
+	// get the 1st IP config attached to the worker node
+	ipConf := interfaceIPConfigs[0]
+	if ipConf.InterfaceIPConfigurationPropertiesFormat == nil {
+		return "", fmt.Errorf("ip configuration properties for VM %s is nil", vmName)
+	}
+	ipConfigProp := *ipConf.InterfaceIPConfigurationPropertiesFormat
+	if ipConfigProp.LoadBalancerBackendAddressPools == nil {
+		return "", fmt.Errorf("load balancer backend address pools for VM %s cannot be nil", vmName)
+	}
+	backendPools := *ipConfigProp.LoadBalancerBackendAddressPools
+	if backendPools == nil {
+		return "", fmt.Errorf("backend address pool id for %s cannot be nil", vmName)
+	}
+	if len(backendPools) < 1 {
+		return "", fmt.Errorf("backend address pools for %s cannot be 0", vmName)
+	}
+	if len(backendPools) > 1 {
+		return "", fmt.Errorf("multiple backend address pools for %s are currently not supported", vmName)
+	}
+	// get the backendAddressPools attached to the IP config
+	backendAddressPool := backendPools[0]
+
+	return *backendAddressPool.ID, nil
+}
+
+// getResource gets a resource, the generic way.
+func (az *AzureProvider) getResource(ctx context.Context, filter string, num int32) ([]resources.GenericResource, error) {
+	list, err := az.resourcesClient.List(
+		ctx,
+		filter,
+		"",
+		&num,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get resources for filter %s: %v", filter, err)
+	}
+	return list.Values(), nil
+}
+
+// getInfoFromNIC gets the interface ip configuration from NIC name by calling the azure api
+func (az *AzureProvider) getInfoFromNIC(ctx context.Context, nicName string) (network.Interface, []network.InterfaceIPConfiguration, error) {
+	interfaceStruct, err := az.nicClient.Get(ctx, az.resourceGroupName, nicName, "")
+	if err != nil {
+		return network.Interface{}, []network.InterfaceIPConfiguration{},
+			fmt.Errorf("cannot fetch the network interface data of NIC %s: %v", nicName, err)
+	}
+
+	if interfaceStruct.InterfacePropertiesFormat == nil {
+		return network.Interface{}, []network.InterfaceIPConfiguration{},
+			fmt.Errorf("ip configuration for NIC %s cannot be nil", nicName)
+	}
+	interfacePropFormat := *interfaceStruct.InterfacePropertiesFormat
+
+	if interfacePropFormat.IPConfigurations == nil {
+		return network.Interface{}, []network.InterfaceIPConfiguration{},
+			fmt.Errorf("ip configuration for NIC %s cannot be nil", nicName)
+	}
+	interfaceIPConfigs := *interfacePropFormat.IPConfigurations
+
+	if len(interfaceIPConfigs) < 1 {
+		return network.Interface{}, []network.InterfaceIPConfiguration{},
+			fmt.Errorf("ip configuration properties for NIC %s cannot be 0", nicName)
+	}
+	// we assume that all worker nodes have only one IP config.
+	// if the worker nodes have more than one IP config, throw an error
+	if len(interfaceIPConfigs) > 1 {
+		return network.Interface{}, []network.InterfaceIPConfiguration{},
+			fmt.Errorf(" NIC %s has multiple ip configurations. This is not supported by WNI", nicName)
+	}
+
+	return interfaceStruct, interfaceIPConfigs, nil
 }
 
 // randomPasswordString generates random string with restrictions of given length.
@@ -910,6 +1054,30 @@ func (az *AzureProvider) getPrivateIP(ctx context.Context) (string, error) {
 	return *privateIp, nil
 }
 
+// disassociateIP disassociates public IP with the ipConfig.
+// This is needed to delete the public IP address as public IP address cannot be deleted without detaching it from
+// the ipConfig.
+func (az *AzureProvider) disassociateIP(ctx context.Context, nicName string) error {
+	interfaceStruct, interfaceIPConfigs, err := az.getInfoFromNIC(ctx, nicName)
+	if err != nil {
+		return fmt.Errorf("failed to get ip configuration for NIC %s: %v", nicName, err)
+	}
+	// remove the ip address association by setting it as 'nil'
+	interfaceIPConfigs[0].PublicIPAddress = nil
+
+	// nicClient.CreateOrUpdate() needs update 'parameter' of type network.interface
+	// here we are sending the edited interface struct with public ip address set to nil
+	future, err := az.nicClient.CreateOrUpdate(ctx, az.resourceGroupName, az.NicName, interfaceStruct)
+	if err != nil {
+		return fmt.Errorf("failed to disassociate/detach the Public IP address from nic %s: %v ", nicName, err)
+	}
+	err = future.WaitForCompletionRef(ctx, az.nicClient.Client)
+	if err != nil {
+		return fmt.Errorf("failed to disassociate/detach the Public IP address from nic %s: %v ", nicName, err)
+	}
+	return nil
+}
+
 // destroyIP deletes the IP address by taking it's name as argument.
 func (az *AzureProvider) destroyIP(ctx context.Context, ipName string) (err error) {
 	_, err = az.ipClient.Delete(ctx, az.resourceGroupName, ipName)
@@ -1011,6 +1179,11 @@ func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
 	err = az.destroyInstance(ctx, vmName)
 	if err != nil {
 		log.Printf("error destroying the instance %s: %v", vmName, err)
+	}
+
+	err = az.disassociateIP(ctx, nicName)
+	if err != nil {
+		log.Printf("error disassociating the Public IP of instance %s: %v", vmName, err)
 	}
 
 	err = az.destroyNIC(ctx, nicName)
--- a/tools/windows-node-installer/test/e2e/azure_test.go
+++ b/tools/windows-node-installer/test/e2e/azure_test.go
@@ -10,12 +10,15 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strings"
 	"testing"
 
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-04-01/network"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
@@ -71,8 +74,18 @@ type requiredRule struct {
 type azureProvider struct {
 	// resourceGroupName of the Windows node
 	resourceGroupName string
+	// subscriptionID of the corresponding azure service principal.
+	subscriptionID string
+	// infraID is the name of existing openshift infrastructure.
+	infraID string
 	// nsgClient to check if winRmHttps port is opened or not.
 	nsgClient network.SecurityGroupsClient
+	// vmClient to query for instance related operations.
+	vmClient compute.VirtualMachinesClient
+	// nicClient to query for nic related operations.
+	nicClient network.InterfacesClient
+	// Loadbalancer Backend Address Pool client
+	lbbapClient network.LoadBalancerBackendAddressPoolsClient
 	// requiredRules is the set of SG rules that need to be created or deleted
 	requiredRules map[string]*requiredRule
 }
@@ -102,6 +115,7 @@ var (
 func TestCreateVM(t *testing.T) {
 	err := setup()
 	require.NoErrorf(t, err, "failed at the setup with error: %v", err)
+	t.Run("check if Windows VM LB is same as Worker LB", testIfWindowsNodeIsBehindWorkerLB)
 	t.Run("check if required security rules are present", testRequiredRules)
 	t.Run("check if ansible is able to ping on the WinRmHttps port", testAnsiblePing)
 	t.Run("check if container logs port is open in Windows firewall", testAzureInstancesFirewallRule)
@@ -175,6 +189,13 @@ func populateAzureInfo() error {
 	if subscriptionId == "" {
 		return fmt.Errorf("failed to get the subscriptionId from AZURE_AUTH_LOCATION: %s", azureCredentials)
 	}
+	azureInfo.subscriptionID = subscriptionId
+
+	infraID, err := oc.GetInfrastructureID()
+	if err != nil {
+		return fmt.Errorf("failed to get the infraID from OpenShift client: %v", err)
+	}
+	azureInfo.infraID = infraID
 
 	// instantiate network security group client.
 	azureInfo.nsgClient = network.NewSecurityGroupsClient(subscriptionId)
@@ -185,6 +206,18 @@ func populateAzureInfo() error {
 		return fmt.Errorf("failed to get azure authorization token with error: %v", err)
 	}
 	azureInfo.nsgClient.Authorizer = resourceAuthorizer
+
+	// initiate the virtual machine client
+	vmClient := getVMClient(resourceAuthorizer, subscriptionId)
+	azureInfo.vmClient = vmClient
+
+	// initiate the network interface card client
+	nicClient := getNicClient(resourceAuthorizer, subscriptionId)
+	azureInfo.nicClient = nicClient
+
+	// initiate the loadbalancer backend address pool client
+	lbbapClient := getLBBAPClient(resourceAuthorizer, subscriptionId)
+	azureInfo.lbbapClient = lbbapClient
 
 	requiredRules, err := constructRequiredRules()
 	if err != nil {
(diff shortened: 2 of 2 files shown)


sebsoto:
Commit/ PR message should be `retrieval` not `retreival`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
 // getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
-func (az *AzureProvider) getWorkerBackendPoolID() *string {
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendPoolID(ctx context.Context) (*string, error) {
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
+	num := int32(1)
```
sebsoto, line 503: This is a magic number. Why this is 1 needs to be explained with both a comment and a more descriptive name

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
 
 // getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
-func (az *AzureProvider) getWorkerBackendPoolID() *string {
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendPoolID(ctx context.Context) (*string, error) {
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
```
sebsoto, line 502: This filter needs to be explained. What is it filtering for?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	num := int32(1)
+	vmResource, err := az.getResource(ctx, filter, num)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker node: %s", err)
+	}
+	vmName := extractResourceName(*vmResource[0].ID)
```
sebsoto, line 508: Why the zeroth entry?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
-func (az *AzureProvider) getWorkerBackendPoolID() *string {
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendPoolID(ctx context.Context) (*string, error) {
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
+	num := int32(1)
+	vmResource, err := az.getResource(ctx, filter, num)
```
sebsoto, line 504: What VM are you retrieving?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	interfaceIPConfigs := *(interfacePropFormat.IPConfigurations)
+	if interfaceIPConfigs[0].InterfaceIPConfigurationPropertiesFormat == nil {
+		return nil, fmt.Errorf("ip configuration properties for %s cannot be nil", vmName)
+	}
+	ipConfigProp := *(interfaceIPConfigs[0].InterfaceIPConfigurationPropertiesFormat)
+	if ipConfigProp.LoadBalancerBackendAddressPools == nil {
+		return nil, fmt.Errorf("load balancer backend address pools for %s cannot be nil", vmName)
+	}
+	backendPools := *(ipConfigProp.LoadBalancerBackendAddressPools)
+	if backendPools[0].ID == nil {
```
sebsoto, line 530: Why the zeroth entry?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,54 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	interfaceStruct, err := az.nicClient.Get(ctx, az.resourceGroupName, nicName, "")
+	if err != nil {
+		return nil, fmt.Errorf("cannot fetch the network interface data of %s: %s", vmName, err)
+	}
+	interfacePropFormat := *(interfaceStruct.InterfacePropertiesFormat)
+	if interfacePropFormat.IPConfigurations == nil {
+		return nil, fmt.Errorf("ip configuration for %s cannot be nil", vmName)
+	}
+	interfaceIPConfigs := *(interfacePropFormat.IPConfigurations)
+	if interfaceIPConfigs[0].InterfaceIPConfigurationPropertiesFormat == nil {
```
sebsoto, line 522: Why the zeroth entry?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
sebsoto: What if you retrieve more than one VM?

 What if you retrieve zero VMs? Is that possible? This will cause a panic if it is

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +498,63 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendPoolID(ctx context.Context) (*string, error) {
+	// Create a filter to get a resource of type virtualMachine which contains worker in its name
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
+	// number of resources to retrieve
+	// we are retrieving just 1 resource as all the worker nodes have same loadbalancers, any workerVM works for our case
```
ravisantoshgudimetla, line 505: This is not entirely true, what if the first node is behind another loadbalancer which was custom created. In general you're making assumptions which are not valid in all the cases.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +463,49 @@ func testAzureSSHConnection(t *testing.T) {
 	err = session.Run("dir")
 	require.NoErrorf(t, err, "failed to communicate vis SSH")
 }
+
+// test if the windows VM is behind the worker node load balancer
+func testAzureVMLoadbalancer(t *testing.T) {
+	// We assume that BackendPoolID for worker vm follows the following format
```
ravisantoshgudimetla, line 469: This is again an assumption. Shouldn't you do something like:

Get the loadbalancer for all the worker nodes(before the test starts) call the method which puts the Windows worker behind the loadbalancer  and in the test code verify if it front ends the Windows Node as well?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -373,6 +384,13 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
 		fmt.Errorf("failed to get ip address: %v", err)
 	}
 
+	// In Azure, for egress as well as for Load balancer service to work, the
+	// windows vm is expected to be placed behind the worker nodes' Load balancer.
```
aravindhp, line 388: `Windows VM`
`load balancer`

Please be consistent. Same issue in your commit messages.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -373,6 +384,13 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
 		fmt.Errorf("failed to get ip address: %v", err)
 	}
 
+	// In Azure, for egress as well as for Load balancer service to work, the
```
aravindhp, line 387: You need to mention that this is in an OpenShift context

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -373,6 +384,13 @@ func (az *AzureProvider) createNIC(ctx context.Context, vnetName, subnetName, ns
...
+	// windows vm is expected to be placed behind the worker nodes' Load balancer.
+	workerBackendPools, err := az.getWorkerBackendPools(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker node backend pool id: %s", err)
```
aravindhp, line 391: `%v` instead of `%s`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,108 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
 }
 
 // getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
-func (az *AzureProvider) getWorkerBackendPoolID() *string {
-	// We assume that BackendPoolID for worker vm follows the following format
-	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendPools(ctx context.Context) ([]network.BackendAddressPool, error) {
```
aravindhp, line 497: Where are you getting the term `BackendPools`? Isn't it `BackendAddressPools`?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,108 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	// Create a filter to get a resource of type virtualMachine which contains worker in its name
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
+	// number of resources to retrieve
+	// https://learnk8s.io/kubernetes-node-size
+	// Kubernetes can technically accommodate 5000 nodes, considering that for optimal performance organizations limit at 500 worker nodes
+	top := int32(500)
+	vmResources, err := az.getResource(ctx, filter, top)
```
aravindhp, line 504: Why do you need to introduce the resource client? Why can't you use the VM client? Which points to you should have added a comment explaining that here.

PratikMahajan: In the vmClient, they need full vmName. We can use list() but list would return all the VMs in the resource group, we will then have to filter it to get the worker VMs using go code iterating through all the VM names.
getResource() does all that work for us. We just need to use the output in that case.

aravindhp: Like I said you need to mention this in a comment.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	// Create a filter to get a resource of type virtualMachine which contains 'worker' in its name
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name)"
+	// number of resources to retrieve
+	// https://learnk8s.io/kubernetes-node-size
+	// Kubernetes can technically accommodate 5000 nodes, considering that for optimal performance organizations limit at 500 worker nodes
+	top := int32(500)
+	// Use getResource() function to get all the workerVMs by filtering VMs which have 'worker' in their name
+	vmResources, err := az.getResource(ctx, filter, top)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker node: %v", err)
```
aravindhp, line 507: You are querying worker VMs and not nodes here. `failed to get worker VMs`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	}
+	if vmResources == nil {
+		return nil, fmt.Errorf("failed to get worker node VM resource")
```
aravindhp, line 510: `worker VMs`

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -186,6 +207,18 @@ func populateAzureInfo() error {
...
+	// initiate the virtual machine client
+	vmClient := getVMClient(resourceAuthorizer, subscriptionId)
+	azureInfo.vmClient = vmClient
+
+	// initiate the network interface card client
+	nicClient := getNicClient(resourceAuthorizer, subscriptionId)
+	azureInfo.nicClient = nicClient
+
+	// initiate the loadbalancer backend address pool client
+	lbBapClient := getLBBAPClient(resourceAuthorizer, subscriptionId)
```
aravindhp, line 219: `lbbapClient`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+
+		interfaceStruct, err := az.nicClient.Get(ctx, az.resourceGroupName, nicName, "")
+		if err != nil {
+			return nil, fmt.Errorf("cannot fetch the network interface data of %s: %v", vmName, err)
+		}
+
+		if interfaceStruct.InterfacePropertiesFormat == nil {
+			return nil, fmt.Errorf("ip configuration for %s cannot be nil", vmName)
+		}
+		interfacePropFormat := *(interfaceStruct.InterfacePropertiesFormat)
```
aravindhp, line 534: Nit: you don't () when dereferencing a pointer.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+
+		if interfacePropFormat.IPConfigurations == nil {
+			return nil, fmt.Errorf("ip configuration for %s cannot be nil", vmName)
+		}
+		interfaceIPConfigs := *(interfacePropFormat.IPConfigurations)
+
+		if interfaceIPConfigs == nil {
```
aravindhp, line 541: Do you need to check if `len(interfaceIPConfigs) < 1`?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+		// we assume that all worker nodes have only one IP config.
+		// if the worker nodes have more than one IP config, throw an error
+		if len(interfaceIPConfigs) > 1 {
+			return nil, fmt.Errorf("multiple interface ip configs are not supported for VM %s", vmName)
+		}
+
+		// get the 1st IP config attached to the worker node
+		ipConf := interfaceIPConfigs[0]
+		if ipConf.InterfaceIPConfigurationPropertiesFormat == nil {
+			return nil, fmt.Errorf("unable to get ip configuration properties for VM %s ", vmName)
```
aravindhp, line 553: `IP configuration properties for VM is nil`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+		}
+		interfaceIPConfigs := *(interfacePropFormat.IPConfigurations)
+
+		if interfaceIPConfigs == nil {
+			return nil, fmt.Errorf("ip configuration properties for %s cannot be nil", vmName)
+		}
+		// we assume that all worker nodes have only one IP config.
+		// if the worker nodes have more than one IP config, throw an error
+		if len(interfaceIPConfigs) > 1 {
+			return nil, fmt.Errorf("multiple interface ip configs are not supported for VM %s", vmName)
```
aravindhp, line 547: `Existing worker `VM %s has multiple NICs. This is not supported by WNI`. Or something to do that effect.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+		return nil, fmt.Errorf("failed to get worker node: %v", err)
+	}
+	if vmResources == nil {
+		return nil, fmt.Errorf("failed to get worker node VM resource")
+	}
+
+	// create a map to keep track of all the loadbalancers associated with different worker nodes.
+	lbSet := make(map[string]struct{})
+	// filler value for map
+	var exists = struct{}{}
```
aravindhp, line 516: Why do you need a map when an array would suffice?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	}
+
+	if len(lbSet) > 1 {
+		return nil, fmt.Errorf("multiple backend address pools over multiple worker nodes is not supported")
+	}
+
+	// create an array of the backend address pools that are attached to worker nodes
+	// which will then be applied to the Windows node
+	var lbBackendPoolAddressIDs []network.BackendAddressPool
+	for workerBackendPoolAddressID := range lbSet {
```
aravindhp, line 578: Why do you need a loop when you know that the size is 1?

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
...
 }
+
+// test if the windows VM is behind the worker node load balancer
+func testAzureVMLoadbalancer(t *testing.T) {
+	ctx := context.Background()
+
+	// create a map to keep track of all the ip configurations associated with the backend pool .
+	ipCSet := make(map[string]struct{})
+	// filler value for map
+	var exists = struct{}{}
```
aravindhp, line 487: Same comment as before.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
...
+
+	// We assume that BackendPoolID for worker vm follows the following format
+	// /subscriptions/$SUBSCRIPTID/resourceGroups/$INFRAID-rg/providers/Microsoft.Network/loadBalancers/$INFRAID/backendAddressPools/$INFRAID
+	// We assume that loadbalancer and backendAddressPools both have $INFRAID as name
+	bapInfo, err := azureInfo.lbBapClient.Get(ctx, azureInfo.resourceGroupName, azureInfo.infraID, azureInfo.infraID)
+	if err != nil {
+		t.Fatalf("failed to get backendAddressPool information: %v", err)
```
aravindhp, line 494: Why do you need `t.Fatalf`? Won't `require.NoError()` suffice? Comment applies to all cases.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
...
+	if bapInfo.BackendIPConfigurations == nil {
+		t.Fatalf("failed to get backend address pool configuration")
+	}
+	backendIPC := *bapInfo.BackendIPConfigurations
+	if len(backendIPC) < 1 {
+		t.Fatalf("empty backend address pool configuration for %s", *bapInfo.ID)
+	}
+	// get all the ip configurations associated with the backend address pool
+	for _, ipC := range backendIPC {
+		ipCSet[extractResourceName(*ipC.ID)] = exists
```
aravindhp, line 505: You are again dereferencing a pointer without checking.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
...
+	}
+
+	// check if all the vm are present in the backend address pool associated with the worker node.
+	for _, vmName := range instanceIDs {
```
aravindhp, line 509: This should loop over all instances and not stop at the first error. You should log, collect the error in a slice and `continue`. In the end you can assert on the errors slice being empty.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: I'm using a `map` to avoid duplicates. 
because golang doesn't have a `set` data structure. i'm implementing a `map` as a `set`
consider a scenario, where we have 10VMs. 
according to the code, it'll go through all the 10VMs one by one and will append their backend pool address to the `Map`. [L568](https://github.com/openshift/windows-machine-config-bootstrapper/blob/f6013f4116d675ef5b228b9b014a2f06cc5a9d03/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go#L568)

Now, if we have backendpool `lb1` on VMs 1,2,4,5 
and backend piil `lb2` on VMs 3,6,7. 
in that case `map` helps us in avoid appending duplicates (using backend pool id as key). 
quick lookup of existing keys (used in test code). 
later i'm using the number of keys to determine if we have multiple worker load balancers attached [L571](https://github.com/openshift/windows-machine-config-bootstrapper/blob/f6013f4116d675ef5b228b9b014a2f06cc5a9d03/tools/windows-node-installer/pkg/cloudprovider/azure/vm.go#L571)

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: Used this as a provision for future, where we will be appending multiple LBs to the Windows VM. Will remove it from the code if needed, Let me know

tools/windows-node-installer/test/e2e/azure_test.go
PratikMahajan: As mentioned in [comment](https://github.com/openshift/windows-machine-config-bootstrapper/pull/186#discussion_r419704661). 
I am using `map` here as a `set` data structure. 
I am retrieving all the `ipConfig` attached to the backend pool and storing it as a `key` in `map`. 
And later i am getting all the `ipConfig` attached to the Windows VMs and checking if the backend address pool has that ip config attached. [L538](https://github.com/openshift/windows-machine-config-bootstrapper/blob/f6013f4116d675ef5b228b9b014a2f06cc5a9d03/tools/windows-node-installer/test/e2e/azure_test.go#L538)
`map` gives me o(1) loopkup for the keys. for array its o(n)

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
aravindhp: If that is the logic you are following then you should stop the minute you find the 2nd worker behind a different LB. So you can use a single variable to track that.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
aravindhp: Then please explain this in a comment

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	}
+
+	// create a map to keep track of all the loadbalancers associated with different worker nodes.
+	lbSet := make(map[string]struct{})
+	// filler value for map
+	var exists = struct{}{}
+
+	// Iterate over all the VM resources and get the worker backend pools attached to all the resources.
+	for _, vmResource := range vmResources {
+		vmName := extractResourceName(*vmResource.ID)
```
aravindhp, line 520: You are dereferecning a pointer without checking.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: Will add a comment here

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: Accepted. I will add an `if` condition to stop when we get more than one lb. 
but this type of implementation makes it future proof, i.e if and when we decide to implement multiple LBs on windows VM. we'll just need to remove/add few `if` conditions and add a loop. Will also add comments in the code regarding this.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: Will add the comment

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -80,6 +81,8 @@ type AzureProvider struct {
 	nsgClient network.SecurityGroupsClient
 	// diskClient to query for disk related operations.
 	diskClient compute.DisksClient
+	// resourcesClient to query resources
```
ravisantoshgudimetla, line 84: which resources specifically are you interested in, in this scenario?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	vmResources, err := az.getResource(ctx, filter, top)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worker node: %v", err)
+	}
+	if vmResources == nil {
+		return nil, fmt.Errorf("failed to get worker node VM resource")
+	}
+
+	// create a map to keep track of all the loadbalancers associated with different worker nodes.
+	lbSet := make(map[string]struct{})
```
ravisantoshgudimetla, line 514: You can just get away with a bool.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,109 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+		ipConf := interfaceIPConfigs[0]
+		if ipConf.InterfaceIPConfigurationPropertiesFormat == nil {
+			return nil, fmt.Errorf("unable to get ip configuration properties for VM %s ", vmName)
+		}
+		ipConfigProp := *(ipConf.InterfaceIPConfigurationPropertiesFormat)
+		if ipConfigProp.LoadBalancerBackendAddressPools == nil {
+			return nil, fmt.Errorf("load balancer backend address pools for %s cannot be nil", vmName)
+		}
+		backendPools := *(ipConfigProp.LoadBalancerBackendAddressPools)
+		if backendPools == nil || len(backendPools) < 1 {
```
ravisantoshgudimetla, line 560: splitting `||` makes it easier to be avoid nil pointers.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
...
+		err, nicName := azureInfo.getNICname(ctx, vmName)
+		if err != nil {
+			t.Fatalf("failed to get NIC name: %v", err)
```
ravisantoshgudimetla, line 512: Change these to asserts to make @aravindhp happy :). This applies everywhere.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,67 @@ func testAzureSSHConnection(t *testing.T) {
 	err = session.Run("dir")
 	require.NoErrorf(t, err, "failed to communicate vis SSH")
 }
+
+// test if the windows VM is behind the worker node load balancer
+func testAzureVMLoadbalancer(t *testing.T) {
```
ravisantoshgudimetla, line 481: perhaps changing the name to `testIfWindowsNodeIsBehindWorkerLB`?

tools/windows-node-installer/test/e2e/azure_test.go
ravisantoshgudimetla, line 509: Why do we need to loop over all the worker instances? The only instance we created is the Windows VM and we need to check if our Windows node is behind the LB. This way we can be more specific and targeted in our test.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: resources client is used to query all resources available. 
i'm querying vm resources using the resources client.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: `struct{}` takes 0 byte 
`bool` takes 1 byte

tools/windows-node-installer/test/e2e/azure_test.go
PratikMahajan: i thought we decided on testing all instances in design meet. i can change this to test single vm if needed to. 
Also, what happens in the scenario where the user created 2 VMs and started the test after that?

aravindhp: @ravisantoshgudimetla `instanceIDs` here is a global variable holding only the Windows VMs created as part of the e2e test and not all the worker nodes.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
ravisantoshgudimetla: But the usage here is confusing because are you're not using map's `val, ok` pattern.

tools/windows-node-installer/test/e2e/azure_test.go
ravisantoshgudimetla: > 

@ravisantoshgudimetla instanceIDs here is a global variable holding only the Windows VMs created as part of the e2e test and not all the worker nodes.



This is where I got confused, thanks for the  clarification.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,129 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
-	var wbpi = fmt.Sprintf("/subscriptions/%s/resourceGroups/%[2]s-rg/providers/"+
-		"Microsoft.Network/loadBalancers/%[2]s/backendAddressPools/%[2]s", az.subscriptionID, az.infraID)
-	return &wbpi
+func (az *AzureProvider) getWorkerBackendAddressPools(ctx context.Context) ([]network.BackendAddressPool, error) {
+	// Create a filter to get a resource of type virtualMachine which contains 'worker' in its name
+	filter := "resourceType eq 'Microsoft.Compute/virtualMachines' and substringof('worker', name) and " +
+		"resourceGroup eq '" + az.resourceGroupName + "'"
+	// number of resources to retrieve
+	// https://learnk8s.io/kubernetes-node-size
+	// Kubernetes can technically accommodate 5000 nodes, considering that for optimal performance organizations limit at 500 worker nodes
```
aravindhp, line 503: Not sure what `considering that for optimal performance organizations` means

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,129 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	}
+
+	// keep track of workerVM load balancer
+	// using string as we are assuming all the worker VMs are behind a single load balancer
+	var backendAddressPool string
+
+	// Iterate over all the VM resources and get the worker backend pools attached to all the resources.
+	for _, vmResource := range vmResources {
+		if vmResource.ID == nil {
+			return nil, fmt.Errorf("id for VM resource cannot be nil")
```
aravindhp, line 524: Will adding `vmResource` to the error message help with debugging?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,129 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	lbBackendPoolAddressIDs := []network.BackendAddressPool{
+		{
+			ID: &backendAddressPool,
+		},
+	}
+
+	return lbBackendPoolAddressIDs, nil
+}
+
+// get the backend address pool associated with the vm
```
aravindhp, line 550: Nit: `VM`

Please be consistent with such usages.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -910,6 +1039,45 @@ func (az *AzureProvider) getPrivateIP(ctx context.Context) (string, error) {
...
+	interfaceIPConfigs := *interfacePropFormat.IPConfigurations
+
+	if len(interfaceIPConfigs) < 1 {
+		return fmt.Errorf("ip configuration properties for nic %s cannot be 0", nicName)
+	}
+	// we assume that all worker nodes have only one IP config.
+	// if the worker nodes have more than one IP config, throw an error
+	if len(interfaceIPConfigs) > 1 {
+		return fmt.Errorf("existing worker VM %s has multiple ip Configs. This is not supported by WNI", nicName)
+	}
```
aravindhp, line 1067: Lines 1046-1067 seems to be a repeated pattern and can be made into a helper.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -910,6 +1039,45 @@ func (az *AzureProvider) getPrivateIP(ctx context.Context) (string, error) {
 	return *privateIp, nil
 }
 
+// disassociate public IP with the ipConfig.
+// This is needed to delete the public IP address as public IP address cannot be deleted without detaching it from
+// the ipConfig.
+func (az *AzureProvider) disassociateIP(ctx context.Context, nicName string) error {
+	interfaceStruct, err := az.nicClient.Get(ctx, az.resourceGroupName, nicName, "")
+	if err != nil {
+		log.Printf("cannot fetch the network interface data of nic %s: %s", nicName, err)
```
aravindhp, line 1048: You need to return an error here

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -1013,6 +1181,11 @@ func (az *AzureProvider) DestroyWindowsVM(vmName string) error {
 		log.Printf("error destroying the instance %s: %v", vmName, err)
 	}
 
+	err = az.disassociateIP(ctx, nicName)
+	if err != nil {
+		log.Printf("error disassociating the Public IP of instance %s: %v", vmName, err)
+	}
+
```
aravindhp, line 1188: Do we need to wait until we are sure the disassociation is complete as it not being complete can cause `destroyIP()` to fail?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: According to link,
`"If you plan to use more than 500 nodes, you can expect to hit some performance bottlenecks that require some effort to solve"`
Can increase it to 5000 if needed

aravindhp: Then please remove `organizations` from the sentence. That is what tripped me up.

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
PratikMahajan: API handles that and will return an error if it fails to disassociate it of if it is unable to disassociate it

aravindhp: Sure but would it help if we waited? Otherwise we are going to end up leaking resources.

PratikMahajan: added polling in disassociate function.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,94 @@ func testAzureSSHConnection(t *testing.T) {
...
+		require.NotNil(t, ipC.ID, "ip configuration id for backend address pool %s cannot be nil", *bapInfo.ID)
+		ipCSet[extractResourceName(*ipC.ID)] = exists
+	}
+
+	// check if all the VMs are present in the backend address pool associated with the worker node.
+	for _, vmName := range instanceIDs {
+		err, nicName := azureInfo.getNICname(ctx, vmName)
+		if err != nil {
+			errorSlice = append(errorSlice, fmt.Sprintf("failed to get NIC name for VM %s: %v", vmName, err))
+			continue // cannot execute rest of the loop without errors
```
aravindhp, line 521: Shouldn't this be `cannot execute rest of the loop on error`?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -910,6 +1054,30 @@ func (az *AzureProvider) getPrivateIP(ctx context.Context) (string, error) {
 	return *privateIp, nil
 }
 
+// disassociate public IP with the ipConfig.
```
aravindhp, line 1057: Doc has to start to with the function name

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,144 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+		&num,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get resources for filter %s: %v", filter, err)
+	}
+	return list.Values(), nil
+}
+
+// getNICInfoFromNIC() gets the interface ip configuration from NIC name by calling the azure api
+func (az *AzureProvider) getNICInfoFromNIC(ctx context.Context, nicName string) (network.Interface, []network.InterfaceIPConfiguration, error) {
```
aravindhp, line 604: Please rename to `getInfoFromNIC()`

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,144 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
...
+	lbBackendPoolAddressIDs := []network.BackendAddressPool{
+		{
+			ID: &backendAddressPool,
+		},
+	}
+
+	return lbBackendPoolAddressIDs, nil
+}
+
+// get the backend address pool associated with the VM
```
aravindhp, line 550: Doc has to start with function name. You have done this in multiple places. Please fix.

tools/windows-node-installer/test/e2e/azure_test.go
```diff
@@ -399,3 +476,94 @@ func testAzureSSHConnection(t *testing.T) {
 	err = session.Run("dir")
 	require.NoErrorf(t, err, "failed to communicate vis SSH")
 }
+
+// test if the windows VM is behind the worker node load balancer
```
aravindhp, line 480: Doc has to start with function name

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -482,12 +494,144 @@ func (az *AzureProvider) constructStorageProfile(imageId string) (storageProfile
 }
 
 // getWorkerBackendPoolID gets the backend pool id of the worker node loadbalancer
```
aravindhp, line 496: Please update the doc

PR #189, opened 2020-04-30
[docs] Updated docs to remove unwanted characters from pdf versions

mansikulkarni96:
Removed <200b> character from linux commands in the pdf version of AWS and Azure docs.


mansikulkarni96:
@VaishnaviHire thanks for the review. I verified again. I just found one at the start of oc get nodes -l kubernetes.io/os=windows which I rectified but I did not find the characters in other places you mentioned.

mansikulkarni96:
@VaishnaviHire I have addressed your changes. PTAL.

PR #192, opened 2020-05-01
[ci] Remove dependency on oc binary

aravindhp:
The oc binary is no longer injected into src image in the CI step registry world. There is no reason for us to use oc when we can achieve the same functionality using the API.
- Get the cluster address using API calls
- Removing patching of the network.operator CR as this needs to be  removed anyhow.

This is needed for https://github.com/openshift/release/pull/8623/

Based on openshift/windows-machine-config-operator#47
Co-authored-by: ravisantoshgudimetla <ravisantoshgudimetla@gmail.com>

### DIFF
--- a/internal/test/framework/framework.go
+++ b/internal/test/framework/framework.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"os"
 	"path/filepath"
 	"sort"
@@ -49,9 +50,6 @@ var (
 	artifactDir string
 	// privateKeyPath is the path to the key that will be used to retrieve the password of each Windows VM
 	privateKeyPath string
-	// clusterAddress is the address of the OpenShift cluster e.g. "foo.fah.com".
-	// This should not include "https://api-" or a port
-	ClusterAddress string
 )
 
 // TestFramework holds the info to run the test suite.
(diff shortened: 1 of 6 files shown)


internal/test/framework/framework.go
```diff
@@ -121,13 +123,48 @@ func initCIvars() error {
...
 
+// getK8sAPIServerEndpoint returns the API server endpoint of the cluster.
+// For example: https://api.abc.devcluster.openshift.com:6443
+func getK8sAPIServerEndpoint() (string, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return "", fmt.Errorf("unable to get config needed for API server interaction: %v", err)
+	}
+
+	if len(cfg.Host) == 0 {
```
sebsoto, line 137: note: interesting, usually see this check made by checking if its `== ""`

internal/test/framework/framework.go
```diff
@@ -138,13 +175,21 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
...
-	// the Intel 82599 network driver
-	instanceType := "m5a.large"
-	if err := initCIvars(); err != nil {
+
+	err := initCIvars()
+	if err != nil {
 		return fmt.Errorf("unable to initialize CI variables: %v", err)
 	}
+
+	ClusterAddress, err = getClusterAddress()
```
sebsoto, line 184: nit: this seems like it would belong in the TestFramework, and not as a global.

mansikulkarni96:
LGTM

aravindhp:
/override ci/prow/e2e-wmcb 
/override ci/prow/e2e-wsu 

Tests are failing as expected

PR #193, opened 2020-05-02
[wni] Don't panic on AWS session creation error

sebsoto:
This commit handles the AWS session creation error instead of panicing.
The previous implementation used `Must()` which asserts that the wrapped
function did not return an error and panics if it did. I removed that so
that we can bubble up the returned error and handle it gracefully.

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -91,7 +91,7 @@ func New(openShiftClient *client.OpenShift, imageID, instanceType, sshKey, crede
 	}
 	session, err := newSession(credentialPath, credentialAccountID, provider.AWS.Region)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not create new AWS session: %v", err)
 	}
 	return &AwsProvider{imageID, instanceType, sshKey,
 		ec2.New(session, aws.NewConfig()),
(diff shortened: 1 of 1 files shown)


tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -107,11 +107,10 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	if _, err := os.Stat(credentialPath); os.IsNotExist(err) {
 		return nil, fmt.Errorf("failed to find AWS credentials from path '%v'", credentialPath)
 	}
-	session := awssession.Must(awssession.NewSession(&aws.Config{
+	return awssession.NewSession(&aws.Config{
```
aravindhp, line 110: Please mention in the commit message how moving from `Must` to `NewSession` fixes the issue.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -107,11 +107,10 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	if _, err := os.Stat(credentialPath); os.IsNotExist(err) {
 		return nil, fmt.Errorf("failed to find AWS credentials from path '%v'", credentialPath)
 	}
-	session := awssession.Must(awssession.NewSession(&aws.Config{
```
ravisantoshgudimetla, line 110: Please update the commiit that this `Must` method is causing the panic here.

aravindhp:
/hold
until #176 merges

PR #194, opened 2020-05-04
[ci] Fix DEFAULT_LOCAL_TMP env var

ravisantoshgudimetla:
The CI's env var for ansible tmp is not
pointing to the user's home we create
in Dockerfile.tools but it is pointing to "/"
causing our CI to fail. This commit ensures
that we're using `/tmp` as the temporary
dir for our ansible playbook.

### DIFF
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -30,6 +30,6 @@ RUN chmod -R g=u+w /go
 ENV GOCACHE="/go/.cache"
 ENV PATH="${PATH}:/usr/local/go/bin"
 ENV GOPATH="/go"
-ENV DEFAULT_LOCAL_TMP="/tmp"
+ENV ANSIBLE_LOCAL_TEMP="/tmp"
 
 ENTRYPOINT [ "/bin/bash" ]
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -23,6 +23,11 @@ RUN pip2 install ansible pywinrm
 # Make Ansible happy with arbitrary UID/GID in OpenShift.
 RUN chmod g=u /etc/passwd /etc/group
 
+# Seems ansible-playbook doesn't work with ansible.cfg
+RUN mkdir -p /.ansible
+RUN chmod 777 -R /.ansible
+
+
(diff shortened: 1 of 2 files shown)


hack/run-wsu-ci-e2e-test.sh
```diff
@@ -28,6 +28,10 @@ TEST_DIR=$WMCO_ROOT/internal/test/wsu
 # Make gopath if doesnt exist
 mkdir -p $GOPATH
 
+# Seems ansible-playbook doesn't work with ansible.cfg
+mkdir -p /.ansible
+chmod 777 -R /.ansible
```
aravindhp, line 33: If this works, it should go in `Dockerfile.tools`. Otherwise we will run into issues running locally.

Dockerfile.tools
```diff
@@ -23,6 +23,11 @@ RUN pip2 install ansible pywinrm
 # Make Ansible happy with arbitrary UID/GID in OpenShift.
 RUN chmod g=u /etc/passwd /etc/group
 
+# Seems ansible-playbook doesn't work with ansible.cfg
+mkdir -p /.ansible
+chmod 777 -R /.ansible
```
aravindhp, line 28: Missing `RUN` in both commands.

I would use `g=u+w` instead of `777`.

Dockerfile.tools
```diff
@@ -23,12 +23,18 @@ RUN pip2 install ansible pywinrm
...
+
+
 # Allow building the WMCB and creation of /go/.cache directory
 RUN chmod -R g=u+w /go
 
 # Explicitly set GOCACHE environment variable
 ENV GOCACHE="/go/.cache"
 ENV PATH="${PATH}:/usr/local/go/bin"
 ENV GOPATH="/go"
+ENV ANSIBLE_LOCAL_TEMP="/tmp"
```
aravindhp, line 38: I wonder if we need to create an `ansible.cfg` file with this option and place it in the WSU playbook directory in the CI image. `ansible-playbook` will apparently look at that.

aravindhp:
Superseded by #195

PR #195, opened 2020-05-04
[ci] Step registry fixes

aravindhp:
- Bring back downloading and installing oc in Dockerfile.tools. This is   needed by the WSU playbook and was erroneously removed in #192.
- Create Ansible tmp directory with user accessible permissions.

### DIFF
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -17,11 +17,23 @@ RUN curl -L -s https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz > go1.13.8.l
     && mv go /usr/local \
     && rm -rf ./go*
 
+# Download and install oc
+RUN curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.2.2/openshift-client-linux-4.2.2.tar.gz -o openshift-origin-client-tools.tar.gz \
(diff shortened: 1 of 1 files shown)


Dockerfile.tools
```diff
@@ -17,18 +17,31 @@ RUN curl -L -s https://dl.google.com/go/go1.13.8.linux-amd64.tar.gz > go1.13.8.l
     && mv go /usr/local \
     && rm -rf ./go*
 
+# Download and install oc
+RUN curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.2.2/openshift-client-linux-4.2.2.tar.gz -o openshift-origin-client-tools.tar.gz \
```
sebsoto, line 21: Might want to use 4.4/4.5 for the client

aravindhp:
> I am fine with older version of client since this is used by our CI only.

I thought about updating it and then went with the approach of just reverting the change. Best to do it separately. I do think we need to update it to 4.4 as that is what a customer would be using.

ravisantoshgudimetla:
> I thought about updating it and then went with the approach of just reverting the change. Best to do it separately. I do think we need to update it to 4.4 as that is what a customer would be using.

Ok, then go ahead and force merge this so that we can get #176 in.

Dockerfile.tools
aravindhp: I prefer doing that in a separate PR. This is just a revert of what was wrongly removed in #192.

PR #201, opened 2020-05-20
[wni]Fix Windows VM leaks in CI

VaishnaviHire:
This PR modifies WNI to reflect following changes 

- Windows VMs are being tagged after creation. This commit ensures that, VMs are
being tagged with OpenShift tag during creation API call. This will make it
possible to detect VMs for cleanup even after tests are interrupted.

[WINC-406](https://issues.redhat.com/browse/WINC-406)

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -467,6 +467,10 @@ func (a *AwsProvider) getNetworkInterface(infraID string) (*ec2.InstanceNetworkI
 func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
 	networkInterface *ec2.InstanceNetworkInterfaceSpecification, iamProfile *ec2.IamInstanceProfileSpecification, userDataInput string) (
 	*ec2.Instance, error) {
+	tagSpec, err := a.createOpenShiftTagSpecification()
+	if err != nil {
+		return nil, fmt.Errorf("error creating tag specification %v", err)
+	}
 	runResult, err := a.EC2.RunInstances(&ec2.RunInstancesInput{
 		ImageId:            aws.String(imageID),
 		InstanceType:       aws.String(instanceType),
@@ -476,6 +480,7 @@ func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
 		NetworkInterfaces:  []*ec2.InstanceNetworkInterfaceSpecification{networkInterface},
 		IamInstanceProfile: iamProfile,
 		UserData:           aws.String(base64.StdEncoding.EncodeToString([]byte(userDataInput))),
+		TagSpecifications:  tagSpec,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -488,6 +493,46 @@ func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
...
 }
 
+// createOpenShiftTagSpecification creates TagSpecification object to add OpenShift tag while
+// launching an instance.
+func (a *AwsProvider) createOpenShiftTagSpecification() ([]*ec2.TagSpecification, error) {
+	infraID, err := a.GetInfraID()
+	if err != nil {
+		return nil, fmt.Errorf("error getting infrastructure ID for the OpenShift cluster, %v", err)
+	}
+	tagslist := map[string]string{
```
ravisantoshgudimetla, line 503: Perhaps a name change since it's a map not taglist

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -488,6 +493,46 @@ func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
...
+		ResourceType: aws.String(ec2.ResourceTypeInstance),
+		Tags:         awsTagslist,
+	}
+
+	return []*ec2.TagSpecification{tagspec}, nil
+}
+
+// createTagList is a generic function to create ec2.Tag object list given
+// a map of required tags
+func createTagList(taglist map[string]string) ([]*ec2.Tag, error) {
```
ravisantoshgudimetla, line 524: I believe you're having a map here for future expansion. As of now, to me it looks like we just pass a tuple(k, v) as we don't have too many tags being attached and this function is always called on a single tag.

VaishnaviHire: Yes. This function can take multiple tags. Even if we pass a tuple, we need to return list of form []*ec2.Tag. Hence, I kept it as a map.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: Please don't change to `tagMap` :-) Just go with `tags`. You don't need to express the datastructure being used in the variable name.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -488,6 +493,46 @@ func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
...
+		// - The kubelet can communicate with cloud provider
+		// - TearDown & Reaper job in OpenShift CI can delete the virtual machine as part of cluster
+		infraIDTagKeyPrefix + infraID: infraIDTagValue}
+
+	awsTagslist, err := createTagList(tagslist)
+	if err != nil {
+		return nil, fmt.Errorf("error creating tags %v", err)
+	}
+
+	tagspec := &ec2.TagSpecification{
```
aravindhp, line 514: `tagSpec`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: s/taglist/tags
s/tags/ec2Tags

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -488,6 +493,46 @@ func (a *AwsProvider) createInstance(imageID, instanceType, sshKey string,
...
+	}
+	tagslist := map[string]string{
+		// Add OpenShift tag, so that
+		// - The kubelet can communicate with cloud provider
+		// - TearDown & Reaper job in OpenShift CI can delete the virtual machine as part of cluster
+		infraIDTagKeyPrefix + infraID: infraIDTagValue}
+
+	awsTagslist, err := createTagList(tagslist)
+	if err != nil {
+		return nil, fmt.Errorf("error creating tags %v", err)
```
aravindhp, line 511: Please list the tags in the error message.

PR #205, opened 2020-06-02
Bug 1814989: [doc] Update rate limit section

aravindhp:
Update the rate limit section in the WSU README and dev preview doc to also include the "Get latest 0.8.x cni plugins version" task. This is so that the user is aware of all the tasks where they could run into GitHub rate limiting.


codyhoag:
@aravindhp do you think it would make sense to provide a link to the *API rate limit* section from the Azure/AWS docs? I assume it applies to both?

codyhoag:
I think the current relevancy for the generic doc is a bit weird. We refer to it for AWS/Azure prereqs, and then it's kind of its own general thing that really contains everything the AWS/Azure docs already cover. I assume the purpose of the general doc was to cover things shared by AWS/Azure in one place, to cut down on editing two docs. For the dev preview docs, that makes sense (although that will be structured slightly different for the official docs, since we have modular techniques to cover that). I can perhaps meet with @Anandnatraj sometime this week to discuss this further, or for him to let me know I'm way off base here :smile: .

aravindhp:
@codyhoag 

> do you think it would make sense to provide a link to the API rate limit section from the Azure/AWS docs? I assume it applies to both?

I added the API rate limiting section back to the AWS and Azure docs and updated them.

> I think the current relevancy for the generic doc is a bit weird. 

I agree with you but I will defer to @Anandnatraj on this.

PR #214, opened 2020-06-10
[wni] Expand Cloud interface

ravisantoshgudimetla:
As of now, we're using CreateWindowsVM which couples Windows VMs creation along
with attaching it to public subnet. This will not be ideal to us when we are
moving to machine-api in the WMCO world. Machine-api allows the VM to be created
only on the private interface. So, in order to mimick that functionality, this
commit expands the Cloud interface with a method called
CreateWindowsVMWithPrivateIP which allows the WindowsVMs to be created on the
private subnet and won't attach a public IP to the VM created.

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -113,6 +113,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	})
 }
 
+// CreateWindowsVMWithPrivateSubnet creates a WindowsVM within a private subnet of the OpenShift cluster. A public
+// IP will not be associated with the WindowsVM created by this method.
+func (a *AwsProvider) CreateWindowsVMWithPrivateSubnet() (windowsVM types.WindowsVM, err error) {
+	return a.createWindowsVM(true)
+}
+
 // CreateWindowsVM takes in imageId, instanceType, and sshKey name to create a Windows instance under the same VPC as
 // the existing OpenShift cluster with the following:
 // - attaches existing cloud-specific cluster worker security group and IAM to gain the same access as the linux
@@ -127,6 +133,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 // On success, the function outputs RDP access information in the commandline interface. It also returns the
 // the Windows VM Object to interact with using SSH, Winrm etc.
 func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
+	return a.createWindowsVM(false)
+}
+
+// createWindowsVM is a helper which creates the WindowsVM. It takes usePrivateSubnet as argument
+// to decide if the the created VM has to be associated with public or private subnet
+func (a *AwsProvider) createWindowsVM(usePrivateSubnet bool) (windowsVM types.WindowsVM, err error) {
 	w := &types.Windows{}
 	// If no AMI was provided, use the latest Windows AMI
 	if a.imageID == "" {
@@ -141,7 +153,7 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 	if err != nil {
 		return nil, err
 	}
-	networkInterface, err := a.getNetworkInterface(infraID)
+	networkInterface, err := a.getNetworkInterface(infraID, usePrivateSubnet)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get network interface, %v", err)
 	}
@@ -180,10 +192,13 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 		log.Printf("failed to assign name for instance: %s, %v", instanceID, err)
 	}
 
-	// Get the public IP
-	publicIPAddress, err := a.GetPublicIP(instanceID)
-	if err != nil {
-		return nil, err
+	var publicIPAddress string
+	if !usePrivateSubnet {
+		// Get the public IP
+		publicIPAddress, err = a.GetPublicIP(instanceID)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// get the private IP
@@ -200,8 +215,15 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 
 	// Build new credentials structure to be used by other actors. The actor is responsible for checking if
 	// the credentials are being generated properly. This method won't guarantee the existence of credentials
-	// if the VM is spun up
-	credentials := types.NewCredentials(instanceID, publicIPAddress, decryptedPassword, winUser)
+	// if the VM is spun up.
+	// Credentials can have both public or private IP Addresses.
+	ipAddress := ""
+	if usePrivateSubnet {
+		ipAddress = privateIPAddress
+	} else {
+		ipAddress = publicIPAddress
+	}
+	credentials := types.NewCredentials(instanceID, ipAddress, decryptedPassword, winUser)
 	w.Credentials = credentials
 
 	// Setup Winrm and SSH client so that we can interact with the Windows Object we created
@@ -435,7 +457,7 @@ func (a *AwsProvider) DestroyWindowsVMs() error {
 // getNetworkInterface is a wrapper function that includes all networking related work including getting OpenShift
 // cluster's VPC and its worker security group, a public subnet within the VPC, and a Windows security group.
 // It returns a valid ec2 network interface or an error.
-func (a *AwsProvider) getNetworkInterface(infraID string) (*ec2.InstanceNetworkInterfaceSpecification, error) {
+func (a *AwsProvider) getNetworkInterface(infraID string, usePrivateSubnet bool) (*ec2.InstanceNetworkInterfaceSpecification, error) {
 	vpc, err := a.getInfrastructureVPC(infraID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get VPC, %v", err)
@@ -444,7 +466,7 @@ func (a *AwsProvider) getNetworkInterface(infraID string) (*ec2.InstanceNetworkI
 	if err != nil {
 		return nil, fmt.Errorf("failed to get cluster worker security group, %v", err)
 	}
-	publicSubnetID, err := a.getPublicSubnetId(infraID, vpc)
+	subnetID, err := a.getSubnetID(infraID, vpc, usePrivateSubnet)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get a Public subnet, %v", err)
 	}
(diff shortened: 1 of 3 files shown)


tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +113,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	})
 }
 
+// CreateWindowsVMWithPrivate creates a WindowsVM with private IP
```
mansikulkarni96, line 116: function name does not match doc

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +113,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
+func (a *AwsProvider) CreateWindowsVMWithPrivateIP() (windowsVM types.WindowsVM, err error) {
+	usePrivateSubnet := true
```
mansikulkarni96, line 118: can rename to isPrivateSubnet?

tools/windows-node-installer/pkg/cloudprovider/azure/vm.go
```diff
@@ -900,6 +900,13 @@ func (az *AzureProvider) constructNetworkProfile(ctx context.Context,
 	return networkProfile, nil
 }
 
+// CreateWindowsVMWithPrivate creates a WindowsVM with private IP.
```
mansikulkarni96, line 903: function name does not match doc

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -127,6 +133,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 // On success, the function outputs RDP access information in the commandline interface. It also returns the
 // the Windows VM Object to interact with using SSH, Winrm etc.
 func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
+	usePrivateSubnet := false
+	return a.createWindowsVM(usePrivateSubnet)
+}
+
+// createWindowsVM is actually helper to create the Windows VM with public IP
```
mansikulkarni96, line 140: the doc could be rewritten

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -180,14 +192,17 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 		log.Printf("failed to assign name for instance: %s, %v", instanceID, err)
 	}
 
-	// Get the public IP
-	publicIPAddress, err := a.GetPublicIP(instanceID)
-	if err != nil {
-		return nil, err
+	var publicIPAddress, privateIPAddress string
```
mansikulkarni96, line 195: I maybe missing something, but is there a reason not to use short hand declaration in local scope?

tools/windows-node-installer/pkg/cloudprovider/factory.go
```diff
@@ -19,6 +19,8 @@ import (
 type Cloud interface {
 	// CreateWindowsVM creates a Windows VM for a given cloud provider
 	CreateWindowsVM() (types.WindowsVM, error)
+	// CreateWindowsVMWithPrivateIP creates a Windows VM for a given cloud provider with private IP
```
sebsoto, line 22: `with a private IP`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
sebsoto, line 118: Why create a variable for this instead of just passing true to `a.createWindowsVM()`?
The function doc for `a.createWindowsVM()` denotes what the parameter is for

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -127,6 +133,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 // On success, the function outputs RDP access information in the commandline interface. It also returns the
 // the Windows VM Object to interact with using SSH, Winrm etc.
 func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
+	usePrivateSubnet := false
```
sebsoto, line 136: same as above

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +113,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	})
 }
 
+// CreateWindowsVMWithPrivate creates a WindowsVM with private IP
+func (a *AwsProvider) CreateWindowsVMWithPrivateIP() (windowsVM types.WindowsVM, err error) {
```
sebsoto, line 117: Shouldn't this be `CreateWindowsVMWithoutPublicIP()` a private IP will be created no matter what

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -891,7 +919,7 @@ func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, e
 	}
 
 	err = fmt.Errorf("could not find a public subnet in VPC: %v", *vpc.VpcId)
-	if !foundPublicSubnet {
+	if !foundSubnet {
 		err = fmt.Errorf("could not find a public subnet in a zone that supports %s instance type",
```
VaishnaviHire, line 923: error message can be changed to `could not find the required subnet in a zone that supports %s instance type`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: It should be `CreateWindowVMWithPrivateSubnet()` :-)

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: How I abhor single use variables :-) And no, it does not really help with making the code readable ;-)

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: I understand declaring `publicIPAddress` here but why do you need to declare `privateIPAddress`?

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -200,8 +215,15 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 
 	// Build new credentials structure to be used by other actors. The actor is responsible for checking if
 	// the credentials are being generated properly. This method won't guarantee the existence of credentials
-	// if the VM is spun up
-	credentials := types.NewCredentials(instanceID, publicIPAddress, decryptedPassword, winUser)
+	// if the VM is spun up.
+	// Credentials can have both public and privateIP Addresses.
```
aravindhp, line 219: How? It can have either private or public IP address.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -832,9 +854,9 @@ func (a *AwsProvider) GetVPCByInfrastructure(infraID string) (*ec2.Vpc, error) {
 	return res.Vpcs[0], nil
 }
 
-// getPublicSubnetId tries to find a public subnet under the VPC and returns subnet id or an error.
-// These subnets belongs to the OpenShift cluster.
-func (a *AwsProvider) getPublicSubnetId(infraID string, vpc *ec2.Vpc) (string, error) {
+// getSubnetId tries to find a subnet under the VPC and returns subnet id or an error.
+// These subnets belongs to the OpenShift cluster. It can be either public or private
+func (a *AwsProvider) getSubnetId(infraID string, vpc *ec2.Vpc, usePrivateSubnet bool) (string, error) {
```
aravindhp, line 859: `getSubnetID`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
ravisantoshgudimetla: For the record, I disagree. I would have not created a single use var if it is a string variable which conveys the argument name that I want to pass to method. Since it is a bool, it will be hard for people to tell unless they go the definition of method. But I changed it

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
ravisantoshgudimetla: Done

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +113,12 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	})
 }
 
+// CreateWindowsVMWithPrivate creates a WindowsVM with private subnet of OpenShift cluster. A public
```
sebsoto, line 116: Doc name mismatch

sebsoto: `within a private subnet of the OpenShift cluster`?

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: If you feel strongly I fine if you leave it in. We are not in the kernel after all :-)

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -216,7 +216,7 @@ func (a *AwsProvider) createWindowsVM(usePrivateSubnet bool) (windowsVM types.Wi
 	// Build new credentials structure to be used by other actors. The actor is responsible for checking if
 	// the credentials are being generated properly. This method won't guarantee the existence of credentials
 	// if the VM is spun up.
-	// Credentials can have both public and privateIP Addresses.
+	// Credentials can have both public or privateIP Addresses.
```
aravindhp, line 219: Sorry to nitpick :-) but it has to be `either public or private IP addresses`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -854,9 +854,9 @@ func (a *AwsProvider) GetVPCByInfrastructure(infraID string) (*ec2.Vpc, error) {
 	return res.Vpcs[0], nil
 }
 
-// getSubnetId tries to find a subnet under the VPC and returns subnet id or an error.
+// getSubnetID tries to find a subnet under the VPC and returns subnet id or an error.
```
aravindhp, line 857: s/id/ID

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
ravisantoshgudimetla: Thank you, addressed it.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
ravisantoshgudimetla: No problem

ravisantoshgudimetla:
/hold

to avoid retests

PR #215, opened 2020-06-16
[wni] Use SSH key based authentication 

mansikulkarni96:
Moving towards Machine API, we want to test if we are able to use
public key based ssh authentication to connect and run ssh commands
over windows vm. This PR adds this ssh configuration and sends public key
as a part of userData powershell script while intance creation.

### DIFF
--- a/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
+++ b/tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/client"
 	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/resource"
 	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/types"
+	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
@@ -136,6 +137,46 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
 	return a.createWindowsVM(false)
 }
 
+// retrievePrivateKey retrieves private key in format required for generating
+// public key using the privateKeyPath
+func (a *AwsProvider) retrievePrivateKey() (*rsa.PrivateKey, error) {
+	privateKeyBytes, err := ioutil.ReadFile(a.privateKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find private key from path: %v with: %v", a.privateKeyPath, err)
+	}
+
+	privateKeyBlock, _ := pem.Decode(privateKeyBytes)
+	if privateKeyBlock == nil {
+		return nil, fmt.Errorf("failed to decode retrieved private key: %v", a.privateKeyPath)
+	}
+
+	var parsedKey interface{}
+	if parsedKey, err = x509.ParsePKCS1PrivateKey(privateKeyBlock.Bytes); err != nil {
+		return nil, fmt.Errorf("failed to parse private key with: %v", err)
+	}
+
+	privateKey, ok := parsedKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("failed to cast private key as rsa key type: %v", err)
+	}
+	return privateKey, nil
+}
+
+// generatePublicKey generates a public key and returns it in the ssh-rsa..
+// string format using the retrieved private key
+func (a *AwsProvider) generatePublicKey() (string, error) {
+	privateKey, err := a.retrievePrivateKey()
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve private key in rsa format: %v", err)
+	}
+	publicKey, err := ssh.NewPublicKey(&privateKey.PublicKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate public key: %v", err)
+	}
+	pubKeyBytes := ssh.MarshalAuthorizedKey(publicKey)
+	return string(pubKeyBytes[:]), nil
+}
+
 // createWindowsVM is a helper which creates the WindowsVM. It takes usePrivateSubnet as argument
 // to decide if the the created VM has to be associated with public or private subnet
 func (a *AwsProvider) createWindowsVM(usePrivateSubnet bool) (windowsVM types.WindowsVM, err error) {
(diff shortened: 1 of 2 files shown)


tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
+// retrievePrivateKey retrieves private key in format required for generating
+// public key using the privateKeyPath
+func (a *AwsProvider) retrievePrivateKey() (*rsa.PrivateKey, error) {
+	privateKeyBytes, err := ioutil.ReadFile(a.privateKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find private key from path: %v", a.privateKeyPath)
+	}
+
+	privateKeyBlock, _ := pem.Decode(privateKeyBytes)
+
```
PratikMahajan, line 126: need to catch errors for privateKeyBlock here, i.e. if its nil

tools/windows-node-installer/test/e2e/aws_test.go
```diff
@@ -465,3 +468,40 @@ func testInstallerJsonFileIsDeleted(t *testing.T) {
...
+			pubKeyAuth,
+		},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+	sshClient, err := ssh.Dial("tcp", windowsVM.GetCredentials().GetIPAddress()+":22", config)
+	require.NoErrorf(t, err, "failed to connect via SSH")
+	session, err := sshClient.NewSession()
+	require.NoErrorf(t, err, "failed to create SSH session")
+	err = session.Run("dir")
+	require.NoErrorf(t, err, "failed to communicate vis SSH")
```
sebsoto, line 491: this should be an assert as it is not required for any other test logic

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
 
+// retrievePrivateKey retrieves private key in format required for generating
+// public key using the privateKeyPath
+func (a *AwsProvider) retrievePrivateKey() (*rsa.PrivateKey, error) {
+	privateKeyBytes, err := ioutil.ReadFile(a.privateKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find private key from path: %v", a.privateKeyPath)
+	}
+
+	privateKeyBlock, _ := pem.Decode(privateKeyBytes)
```
sebsoto, line 125: This needs to be checked for not nil:
```
 Decode will find the next PEM formatted block (certificate, private key etc) in the input. It returns that block and the remainder of the input. If no PEM data is found, p is nil and the whole of the input is returned in rest. 
```

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
+
+	var parsedKey interface{}
```
sebsoto, line 127: nit:  variable definition not needed

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
+	if parsedKey, err = x509.ParsePKCS1PrivateKey(privateKeyBlock.Bytes); err != nil {
+		return nil, fmt.Errorf("failed to parse private key with: %v", err)
+	}
+
+	privateKey, ok := parsedKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("failed to convert private key to rsa key type: %v", err)
```
sebsoto, line 134: nit: it might be more appropriate to say `cast as` instead of `convert to`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
...
+}
+
+// generatePublicKey generates a public key and returns it in the ssh-rsa..
+// string format using the retrieved private key
+func (a *AwsProvider) generatePublicKey() (string, error) {
+	privateKey, err := a.retrievePrivateKey()
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve private key in rsa format: %v", err)
+	}
+	publicRsaKey, err := ssh.NewPublicKey(&privateKey.PublicKey)
```
sebsoto, line 146: `publicRsaKey` -> `publicRSAKey`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -113,6 +114,43 @@ func newSession(credentialPath, credentialAccountID, region string) (*awssession
 	})
 }
 
+// retrievePrivateKey retrieves private key in format required for generating
+// public key using the privateKeyPath
+func (a *AwsProvider) retrievePrivateKey() (*rsa.PrivateKey, error) {
+	privateKeyBytes, err := ioutil.ReadFile(a.privateKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find private key from path: %v", a.privateKeyPath)
```
sebsoto, line 122: Please include the actual error, there can be other reasons instead of the file not being missing that readfile will fail

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
mansikulkarni96: we are not supporting this on azure through wni as per the ticket, as we just wanted to ensure public key authentication works and use it in wmco.

tools/windows-node-installer/test/e2e/aws_test.go
```diff
@@ -465,3 +468,40 @@ func testInstallerJsonFileIsDeleted(t *testing.T) {
...
+}
+
+// getPubKeyAuth parses the private key file to return an AuthMethod
+// that uses the given key pairs in creating ssh config
+func getPubKeyAuth(privateKeyPath string) (ssh.AuthMethod, error) {
```
sebsoto, line 496: `new` might be more appropriate than `get`

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: I would suggest `publicKey`. The key needn't use RSA.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -150,20 +188,40 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
...
+	Start-Service sshd
+	$pubKeyConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PubkeyAuthentication yes','PubkeyAuthentication yes'
+	$pubKeyConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+ 	$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication yes'
+	$passwordConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+	$pubKeyLocationConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'Match Group administrators','#Match Group administrators'
+	$pubKeyLocationConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+	Restart-Service sshd
+	New-item -Path $env:USERPROFILE -Name .ssh -ItemType Directory -force
+	echo "` + publicKey + `"| Out-File $env:USERPROFILE\.ssh\authorized_keys -Encoding ascii
```
aravindhp, line 220: Please align

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -150,20 +188,40 @@ func (a *AwsProvider) CreateWindowsVM() (windowsVM types.WindowsVM, err error) {
...
+	}
+
+	// PowerShell script to setup WinRM for Ansible, installing OpenSSH server, configuring
+	// public key authentication and open firewall port number 10250 on the Windows node created
+	userData := `<powershell>
+	$url = "https://raw.githubusercontent.com/ansible/ansible/devel/examples/scripts/ConfigureRemotingForAnsible.ps1"
+	$file = "$env:temp\ConfigureRemotingForAnsible.ps1"
+	(New-Object -TypeName System.Net.WebClient).DownloadFile($url,  $file)
+	& $file
+	Add-WindowsCapability -Online -Name OpenSSH.Server~~~~0.0.1.0
```
aravindhp, line 203: Do we still need to stick to version `0.0.1.0`?

aravindhp: I guess we have to as per the Microsoft [doc](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse#installing-openssh-with-powershell).

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
mansikulkarni96: done

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
mansikulkarni96: done error check

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
```diff
@@ -162,20 +203,40 @@ func (a *AwsProvider) createWindowsVM(usePrivateSubnet bool) (windowsVM types.Wi
...
+	Set-Service -Name sshd -StartupType ‘Automatic’
+	Start-Service ssh-agent
+	Start-Service sshd
+	$pubKeyConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PubkeyAuthentication yes','PubkeyAuthentication yes'
+	$pubKeyConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+ 	$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication yes'
+	$passwordConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+	$pubKeyLocationConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'Match Group administrators','#Match Group administrators'
+	$pubKeyLocationConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+	Restart-Service sshd
```
aravindhp, line 233: @mansikulkarni96, sorry I did not catch this before. Can't you just start the service here instead of restarting? I would look into doing that when moving this to WMCO.

tools/windows-node-installer/pkg/cloudprovider/aws/ec2.go
aravindhp: The alignment I was asking for is similar to what was there before. Please do that when moving this to WMCO.

PR #216, opened 2020-06-22
Bug 1814433: [wmcb] Generate Windows specific kubelet config

VaishnaviHire:
This commit ensures that kubelet is configured specific
to windows environment.

Following changes are made : 

- wmcb creates kubelet.conf file using a config template
- config template can be updated using `bindata`  by running hack script

Bug tickets that can be closed after this PR

- [WINC-298](https://issues.redhat.com/browse/WINC-298)
- [WINC-297](https://issues.redhat.com/browse/WINC-299)

 [WINC-127](https://issues.redhat.com/browse/WINC-127), [WINC-215](https://issues.redhat.com/browse/WINC-215), [WINC-301](https://issues.redhat.com/browse/WINC-301)  require upstream changes


pkg/bootstrapper/bindata.go
```diff
@@ -0,0 +1,244 @@
+// Code generated for package bootstrapper by go-bindata DO NOT EDIT. (@generated)
```
sebsoto, line 1: This file is adding a lot of exposed functions to the bootstrapper package, which i feel should not be exposed. Its also adding additional practices of code generation.

Why use this `go-bindata` approach instead of forming a KubeletConfiguration object? https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/config/types.go#L74. I feel like if you take that approach the code will be a lot simpler, and we wont have to deal with the extra hack script, generated code, the template file, etc.

VaishnaviHire: Yes. I did look into `KubeletConfiguration` object. As we are not using the ignition file anymore, we would have to manually define most of the fields in the struct. So I am not sure, if that would be a cleaner approach.

We can also opt to not expose the `bindata.go` file and just make its generation as part of the build. So every time we build wmcb, it can be regenerated. This would reduce the extra files.

pkg/bootstrapper/kubelet_config.json
```diff
@@ -0,0 +1 @@
...cPodPath":"/etc/kubernetes/manifests","syncFrequency":"0s","fileCheckFrequency":"0s","httpCheckFrequency":"0s","rotateCertificates":true,"serverTLSBootstrap":true,"resolvConf":"","authentication":{"x509":{"clientCAFile":"{{.ClientCAFile}} "},"webhook":{"cacheTTL":"0s"},"anonymous":{"enabled":false}},"authorization":{"webhook":{"cacheAuthorizedTTL":"0s","cacheUnauthorizedTTL":"0s"}},"clusterDomain":"cluster.local","clusterDNS":["172.30.0.10"],"streamingConnectionIdleTimeout":"0s","nodeStatusUpdateFrequency":"0s","nodeStatusReportFrequency":"0s","imageMinimumGCAge":"0s","volumeStatsAggPeriod":"0s","systemCgroups":"/system.slice","cgroupRoot":"/","cgroupsPerQOS":false,"cgroupDriver":"cgroupfs","cpuManagerReconcilePeriod":"0s","runtimeRequestTimeout":"0s","maxPods":250,"kubeAPIQPS":50,"kubeAPIBurst":100,"serializeImagePulls":false,"evictionPressureTransitionPeriod":"0s","featureGates":{"LegacyNodeRoleBehavior":false,"NodeDisruptionExclusion":true,"RotateKubeletServerCertificate":true,"SCTPSupport":true,"ServiceNodeExclusion":true,"SupportPodPidsLimit":true},"containerLogMaxSize":"50Mi","systemReserved":{"cpu":"500m","ephemeral-storage":"1Gi","memory":"1Gi"},"enforceNodeAllocatable":[]}
```
ravisantoshgudimetla, line 1: I'd put this in `pkg/bootstrapper/manifests` or `pkg/bootstrapper/templates`?

pkg/bootstrapper/bindata.go
ravisantoshgudimetla: That sounds reasonable, the downside to that approach is what will happen if we want to just test the generated code. I think it's better to expose the endpoint but the build can always opt to generate

pkg/bootstrapper/kubelet_config.json
ravisantoshgudimetla: Removed the eof line, please add it back

pkg/bootstrapper/bootstrapper.go
```diff
@@ -211,51 +203,44 @@ type fileTranslation struct {
...
 
-// prepKubeletConfForWindows adds all Windows specific configuration options we need to the kubelet configuration
-// specifically, we change the cgroup driver, CA path, resolv.conf path, and enforce node allocatable.
-func prepKubeletConfForWindows(wmcb *winNodeBootstrapper, initialConfig []byte) ([]byte, error) {
-	var out []byte
-	// Here we parse the initial configuration, which was yaml, into a KubeletConfiguration struct
-	b := bufio.NewReader(bytes.NewReader(initialConfig))
-	r := yaml.NewYAMLReader(b)
-	doc, err := r.Read()
+// kubeletConf defines fileds of kubelet.conf file that are defined by wmcb variables
```
aravindhp, line 206: `WMCB`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -211,51 +203,44 @@ type fileTranslation struct {
...
+type kubeletConf struct {
+	// ClientCAFile specifies location to client certificate
+	ClientCAFile string
+}
+
+// prepKubeletConfForWindows creates config file for kubelet, with windows specific configuration
+// Add values in kubelet_config.json files, for additional static fields.
+// Add fields in kubeletConf struct for variable fields
+func prepKubeletConfForWindows(wmcb *winNodeBootstrapper) ([]byte, error) {
```
aravindhp, line 215: I suggest `createKubeletConf` instead.

pkg/bootstrapper/kubelet_config.json
aravindhp: Please use `templates`

pkg/bootstrapper/bindata.go
VaishnaviHire: `openshift/origin` exposes `bindata.go`  in test and has a hack script to update it, as @ravisantoshgudimetla  suggested. 
https://github.com/openshift/origin/blob/d6084de4c1cfb6dfb00c1d746debbfba628822ad/test/extended/testdata/bindata.go

pkg/bootstrapper/bootstrapper.go
```diff
@@ -1,27 +1,19 @@
...
 
 import (
-	"bufio"
-	"bytes"
-	"encoding/json"
 	"fmt"
 	ignitionv2 "github.com/coreos/ignition/config/v2_2"
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
```
mansikulkarni96, line 8: please sort the packages according to inbuilt and imported

pkg/bootstrapper/bootstrapper.go
```diff
@@ -211,51 +203,44 @@ type fileTranslation struct {
...
+	err = kubeletConfTmpl.Execute(kubeletConfigPath, variableFields)
 	if err != nil {
-		return out, err
+		return nil, fmt.Errorf("error writing data to kubelet.conf file: %v", err)
 	}
-
-	// replacing EnforceNodeAllocatable with an empty slice,
-	outString := strings.Replace(string(out), "\"THIS_MUST_BE_EMPTY\"", "", -1)
-	return []byte(outString), err
+	data, _ := ioutil.ReadFile(kubeletConfigPath.Name())
```
mansikulkarni96, line 242: please add an error check for the read , data can be renamed to something specific

pkg/bootstrapper/bindata.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, fmt.Errorf("Read %q: %v", name, err)
+	}
+	if clErr != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+type asset struct {
```
mansikulkarni96, line 38: please add a doc for the structs

pkg/bootstrapper/bindata.go
aravindhp: `openshift/origin` needn't follow the best practices around this :-)  Anyhow please articulate what your plan is.

pkg/bootstrapper/bindata.go
VaishnaviHire: This is an auto generated file

pkg/bootstrapper/bindata.go
VaishnaviHire: **Approach:** Didnt expose the `bindata.go` file. It can be generated as part of `build`, `build-wmcb-unit-test` and `build-wmcb-e2e-test`.

**Reasoning:**  Keeping `bindata` as a part of `build-wmcb-unit-test` lets us test only the generated code. `bindata.go` is no longer exposed, resulting in a cleaner codebase.

.gitignore
```diff
@@ -22,3 +22,6 @@ vendor/
 tools/windows-node-installer/wni
 windows-node-installer.json
 winworker*
+
+# bindata files
+pkg/bootstrapper/bindata.go
```
ravisantoshgudimetla, line 27: Why are we ignoring the bindata files? Once we run `go bin-data`, this file will be updated too.

Makefile
```diff
@@ -58,3 +58,7 @@ run-wsu-ci-e2e-test:
 # TODO: Add other verifications
 verify-all:
 	hack/verify-gofmt.sh
+
+.PHONY: bindata
+bindata:
+	hack/generate-conf-files.sh
```
ravisantoshgudimetla, line 64: new line at EOF.

.gitignore
VaishnaviHire: According to conversation [here](https://github.com/openshift/windows-machine-config-bootstrapper/pull/216#discussion_r447305876), `bindata.go` is generated as part of the build. Will exposing the `bindata` file be useful ?

pkg/bootstrapper/templates/kubelet_config.json
```diff
@@ -0,0 +1 @@
...dPath":"/etc/kubernetes/manifests","syncFrequency":"0s","fileCheckFrequency":"0s","httpCheckFrequency":"0s","rotateCertificates":true,"serverTLSBootstrap":true,"resolvConf":"","authentication":{"x509":{"clientCAFile":"{{.ClientCAFile}} "},"webhook":{"cacheTTL":"0s"},"anonymous":{"enabled":false}},"authorization":{"webhook":{"cacheAuthorizedTTL":"0s","cacheUnauthorizedTTL":"0s"}},"clusterDomain":"cluster.local","clusterDNS":["172.30.0.10"],"streamingConnectionIdleTimeout":"0s","nodeStatusUpdateFrequency":"0s","nodeStatusReportFrequency":"0s","imageMinimumGCAge":"0s","volumeStatsAggPeriod":"0s","systemCgroups":"/system.slice","cgroupRoot":"/","cgroupsPerQOS":false,"cgroupDriver":"cgroupfs","cpuManagerReconcilePeriod":"0s","runtimeRequestTimeout":"10m0s","maxPods":250,"kubeAPIQPS":50,"kubeAPIBurst":100,"serializeImagePulls":false,"evictionPressureTransitionPeriod":"0s","featureGates":{"LegacyNodeRoleBehavior":false,"NodeDisruptionExclusion":true,"RotateKubeletServerCertificate":true,"SCTPSupport":true,"ServiceNodeExclusion":true,"SupportPodPidsLimit":true},"containerLogMaxSize":"50Mi","systemReserved":{"cpu":"500m","ephemeral-storage":"1Gi","memory":"1Gi"},"enforceNodeAllocatable":[]}
```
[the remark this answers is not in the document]
VaishnaviHire: Adding a newline is breaking the unit test here, as we are comparing the bytes.

pkg/bootstrapper/templates/kubelet_config.json
```diff
@@ -0,0 +1 @@
+{"kind":"KubeletConfiguration","apiVersion":"kubelet.config.k8s.io/v1beta1","rotateCertificates":true,"serverTLSBootstrap":true,"resolvConf":"","authentication":{"x509":{"clientCAFile":"{{.ClientCAFile}} "},"anonymous":{"enabled":false}},"clusterDomain":"cluster.local","clusterDNS":["172.30.0.10"],"cgroupsPerQOS":false,"runtimeRequestTimeout":"10m0s","maxPods":250,"kubeAPIQPS":50,"kubeAPIBurst":100,"serializeImagePulls":false,"featureGates":{"LegacyNodeRoleBehavior":false,"NodeDisruptionExclusion":true,"RotateKubeletServerCertificate":true,"SCTPSupport":true,"ServiceNodeExclusion":true,"SupportPodPidsLimit":true},"containerLogMaxSize":"50Mi","systemReserved":{"cpu":"500m","ephemeral-storage":"1Gi","memory":"1Gi"},"enforceNodeAllocatable":[]}
```
ravisantoshgudimetla, line 1: Why do we need resolvConf?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -211,51 +204,49 @@ type fileTranslation struct {
...
+	err = kubeletConfTmpl.Execute(kubeletConfFile, variableFields)
 	if err != nil {
-		return out, err
+		return nil, fmt.Errorf("error writing data to %v file: %v", kubeletConfPath, err)
 	}
 
-	// replacing EnforceNodeAllocatable with an empty slice,
-	outString := strings.Replace(string(out), "\"THIS_MUST_BE_EMPTY\"", "", -1)
-	return []byte(outString), err
+	kubeletConfData, err := ioutil.ReadFile(kubeletConfFile.Name())
```
ravisantoshgudimetla, line 245: Are you returning value to be used to in tests? I don't see it being used elsewhere.

pkg/bootstrapper/templates/kubelet_config.json
ravisantoshgudimetla: New line.

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: yes and it would be easier to debug

pkg/bootstrapper/templates/kubelet_config.json
VaishnaviHire: The default value for `resolvConf` is `/etc/resolv.conf` which is not required for Windows.

ravisantoshgudimetla: So, we if you make it "", would it point to nothing. IIRC, in the command line flags, it shows up as `/etc/resolv.conf` if give it empty

pkg/bootstrapper/bootstrapper.go
ravisantoshgudimetla: Hmm, not a huge fan of that, the method is generating a file and we should use the file as source of truth but I can live with it.

VaishnaviHire:
hold cancel

PR #217, opened 2020-06-26
[ci] Change key pair

ravisantoshgudimetla:
Test if the key pair change is working properly


aravindhp:
/approve

If the e2e tests pass

PR #220, opened 2020-07-22
[wmcb] Change kubelet log directory

sebsoto:
This commit changes the kubelet log directory to C:\var\log\kubelet.
This change is being done to allow `oc adm node-logs` to access the log
file. The --path argument given to that command requires the logs to be
in a sub-directory of /var/log, which resolves to C:\var\log on windows.


VaishnaviHire:
Mostly LGTM. Just one quick question , do we also need to make changes in `bootstrapper/bootstrapper_test.go` file ?

https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/pkg/bootstrapper/bootstrapper_test.go#L522

sebsoto:
Those changes do not need to be made, as that test is just testing that the kubelet args can be changed

aravindhp:
I would suggest making them for consistency if it is not too difficult.

PR #221, opened 2020-07-22
Bug 1859537: Add retries for SSH config

aravindhp:
Add retries when attempting to configure the SSH client as the VM might still be processing the user data.

### DIFF
--- a/tools/windows-node-installer/pkg/types/types.go
+++ b/tools/windows-node-installer/pkg/types/types.go
@@ -231,7 +231,19 @@ func (w *Windows) GetSSHClient() error {
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 	}
 
-	sshClient, err := ssh.Dial("tcp", w.Credentials.GetIPAddress()+":22", config)
+	var err error
+	var sshClient *ssh.Client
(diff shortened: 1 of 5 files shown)


tools/windows-node-installer/pkg/types/types.go
```diff
@@ -231,7 +231,19 @@ func (w *Windows) GetSSHClient() error {
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 	}
 
-	sshClient, err := ssh.Dial("tcp", w.Credentials.GetIPAddress()+":22", config)
+	var err error
+	var sshClient *ssh.Client
+	// Retry if we are unable to create a client as the VM could still be executing the steps in its user data. We
+	// cannot reuse the entries in the retry package as they are too granular.
+	for retries := 0; retries < 5; retries++ {
```
gmarkley-VI, line 238: Does this provide us enough safety moving forward? Windows is slow to boot... And Env dependent. I would retry every min for 10 min to try and cover other use cases like VMware in the future.

aravindhp: I am fine increasing the time to 10 minutes. We will have to do the same in WMCO.

PR #222, opened 2020-07-23
[internal] Bump WNI version

aravindhp:
Bumped WNI version to the latest by removing `github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer` from `internal/test/go.mod` and ran `go mod tidy`. This resulted in conflicting packages being used by the internal test and WNI. The easiest way around seems to be matching what is being used in theinternal package and WNI. This was done by not pinning any versions. A side effect was that many client calls needed to be updated.


internal/test/go.mod
```diff
@@ -3,31 +3,45 @@ module github.com/openshift/windows-machine-config-bootstrapper/internal/test
 go 1.12
 
 replace (
-	github.com/openshift/api => github.com/openshift/api v0.0.0-20200205145930-e9d93e317dd1 // OpenShift 4.3
-	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191125132246-f6563a70e19a // OpenShift 4.3
-	k8s.io/api => k8s.io/api v0.16.7
-	k8s.io/apimachinery => k8s.io/apimachinery v0.16.7
-	k8s.io/client-go => k8s.io/client-go v0.16.7
+	github.com/Azure/azure-sdk-for-go => github.com/Azure/azure-sdk-for-go v34.1.0+incompatible
```
ravisantoshgudimetla, line 6: > Bumped WNI version to the latest by running go get -u github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer

Shouldn't we run `go mod tidy` which won't touch the rest of the packages, if they're already pinned. If there is a transitive dependency, WNI will fail to update and we should update those packages? I am curious why we had to pin the `azure-sdk` considering we are not directly using it.

aravindhp: I did run `go mod tidy` too. I should have mentioned that in my comments. So I am not sure what you want me to do. Just run `go mod tidy` and not `get -u github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer`? How would WNI get updated? Please clarify.

>I am curious why we had to pin the azure-sdk considering we are not directly using it.

If I don't pin it the latest version the Azure SDK is getting pulled and it is conflicting with what WNI is using.

PR #223, opened 2020-07-30
bootstrapper: Add support for Ignition config spec v3.1

LorbusChris:
```
commit 89b1698d5d9b38fd7f56f1ff343057418ae0316a
Author: Christian Glombek <cglombek@redhat.com>
Date:   Tue Aug 18 20:58:13 2020 +0200

    [bootstrapper] Switch to Ignition config spec v3.1
    
    The bootstrapper now supports parsing Ignition config spec v3.1.
    
    With this commit, the bootstrapper uses spec v3.1 as the canonical
    internal representation of the Ignition configuration it processes.

commit 44ce4e8f8297ba0a3f36a7576c1764982b2bda6b
Author: Christian Glombek <cglombek@redhat.com>
Date:   Tue Aug 25 18:23:41 2020 +0200

    [bootstrapper] Add back Ignition config spec v2.x parsing
    
    With this commit the bootstrapper is now again able to parse
    Ignition spec v2 config, and to convert it to spec v3.1 before
    further processing.
    
    This dual parsing support is necessary until the windows-mco is
    changed to pull Ignition spec v3.1 config from the
    machine-config-server (MCS), and can be removed afterwards.

commit 7033a1dc26ad8a2f8dd0d4d098794207c89a88ca
Author: Christian Glombek <cglombek@redhat.com>
Date:   Mon Aug 31 18:58:25 2020 +0200

    Update vendor
    
    by running `go mod vendor`.
```


LorbusChris:
```
Detected go version: go version go1.13.8 linux/amd64.
Kubernetes requires go1.14.4 or greater.
Please install go1.14.4 or later. 
```
hm..
/retest

LorbusChris:
This is now ready for review. Not sure if a BZ for this exists already. If not, I'll create one.

aravindhp:
> This is now ready for review. Not sure if a BZ for this exists already. If not, I'll create one.

Thanks for working on this @LorbusChris. There is a [Jira story](https://issues.redhat.com/browse/WINC-458) tracking this so a BZ is not required.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -267,13 +272,41 @@ func (wmcb *winNodeBootstrapper) translateFile(ignitionSource string, fileTransl
 	return newContents, err
 }
 
+// convertIgnition2to3 takes an ignition spec v2.3 config and returns a v3.1 config
```
LorbusChris, line 275: this should say `spec v2.4`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -323,7 +328,11 @@ func (wmcb *winNodeBootstrapper) parseIgnitionFileContents(ignitionFileContents
 	// and write it to the destination path
 	for _, ignFile := range configuration.Storage.Files {
 		if filePair, ok := filesToTranslate[ignFile.Node.Path]; ok {
-			newContents, err := wmcb.translateFile(ignFile.Contents.Source, filePair.translationFunc)
+			if ignFile.Contents.Source == nil {
```
sebsoto, line 364: Should this be done directly below the `for _, ignFile := range configuration.Storage.Files {` line?
No point in doing `if filePair, ok := filesToTranslate[ignFile.Node.Path]; ok {` if the file is empty

LorbusChris: Having empty file contents is not invalid per se and some files like that could be included in the Ignition config we get from the MCS.
My reasoning here was that we really just care about the files in the `filesToTranslate` map, which must not be empty, and we can fail in that case.

tools/ansible/tasks/wsu/main.yaml
```diff
@@ -257,7 +257,8 @@
       win_get_url:
         url: "https://api-int.{{ cluster_address }}:22623/config/worker"
         dest: "{{ win_temp_dir.path }}\\worker.ign"
-        http_agent: "Ignition/0.35.0"
+        headers:
+          Accept: "application/vnd.coreos.ignition+json`;version=3.1.0"
```
VaishnaviHire, line 261: Is there an extra single quote after `ignition+json` ?

go.mod
```diff
@@ -11,20 +11,20 @@ replace (
...
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.3
-	github.com/stretchr/testify v1.3.0
+	github.com/stretchr/testify v1.5.1
 	github.com/vincent-petithory/dataurl v0.0.0-20160330182126-9a301d65acbb
 	go.uber.org/atomic v1.4.0 // indirect
 	go.uber.org/zap v1.10.0
-	go4.org v0.0.0-20190919214946-0cfe6e5be80f // indirect
-	golang.org/x/sys v0.0.0-20190403152447-81d4e9dc473e
+	golang.org/x/sys v0.0.0-20200610111108-226ff32320da
```
mansikulkarni96, line 27: are these updates required as a part of this change?

LorbusChris: They aren't pinned via a replace directive, so `go mod tidy` has updated them here to a version it deems compatible.

tools/ansible/tasks/wsu/main.yaml
LorbusChris: \` is the escape character in PowerShell (I had to use it above at https://github.com/openshift/windows-machine-config-bootstrapper/pull/223/files#diff-631087f953268c655ad0d05d71b2f6e2R206 to escape `;`

That said, I'm not sure how or whether this ansible module might do some sanitation on input strings - it might not be necessary. Is there a way to test this code? (IIRC the wsu e2e test was removed a while back)

LorbusChris: I've removed this part since the playbook is going to be deprecated anyway (as per yesterday's decision in the design meeting)

LorbusChris:
@aravindhp thanks for the pointer! As the e2e tests were still running anyway, I've changed the messages to adhere to this repo's common guidelines
(and also used the opportunity to reword the messages a bit)

test/e2e/bootstrapper_test.go
```diff
@@ -23,7 +23,7 @@ var installDir string
...
...ldM7GlBnGdzzD0z78aqq54KVP2HTvMx8Qte5hPi7UqsV3Xigk4YwQpxwV4I_GdalNHzTJZxRjuA6-A%0A"},"mode":420}]},"systemd":{"units":[{"contents":"[Unit]\nDescription=Kubernetes Kubelet\nWants=rpc-statd.service crio.service\nAfter=crio.service\n\n[Service]\nType=notify\nExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests\nExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state\nEnvironmentFile=/etc/os-release\nEnvironmentFile=-/etc/kubernetes/kubelet-workaround\nEnvironmentFile=-/etc/kubernetes/kubelet-env\n\nExecStart=/usr/bin/hyperkube \\\n    kubelet \\\n      --config=/etc/kubernetes/kubelet.conf \\\n      --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig \\\n      --kubeconfig=/var/lib/kubelet/kubeconfig \\\n      --container-runtime=remote \\\n      --container-runtime-endpoint=/var/run/crio/crio.sock \\\n      --node-labels=node-role.kubernetes.io/worker,node.openshift.io/os_id=${ID} \\\n      --minimum-container-ttl-duration=6m0s \\\n      --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \\\n      --cloud-provider=aws \\\n       \\\n      --v=3\n\nRestart=always\nRestartSec=10\n\n[Install]\nWantedBy=multi-user.target\n","enabled":true,"name":"kubelet.service"}]}}`
```
ravisantoshgudimetla, line 26: If possible, please keep the 2.x config as well to exercise the code path. It's fine to keep if you think this is going to be a throw away work

internal/test/wmcb/wmcb_test.go
```diff
@@ -202,11 +202,11 @@ func (vm *wmcbVM) initializeTestBootstrapperFiles() error {
 		return fmt.Errorf("unable to copy kubelet.exe to %s", winTemp)
 	}
 
-	// The 0.35.0 maps to ignition spec v2. This should be modified when we switch to v3
-	ignitionUserAgentSpec := "Ignition/0.35.0"
+	// Ignition v2.3.0 maps to Ignition config spec v3.1.0.
+	ignitionAcceptHeaderSpec := "application/vnd.coreos.ignition+json`;version=3.1.0"
```
ravisantoshgudimetla, line 206: If possible, please keep the 2.x config as well to exercise the code path. It's fine to keep if you think this is going to be a throw away work

pkg/bootstrapper/bootstrapper.go
```diff
@@ -283,15 +317,19 @@ func (wmcb *winNodeBootstrapper) parseIgnitionFileContents(ignitionFileContents
 			continue
 		}
 
-		results := cloudProviderRegex.FindStringSubmatch(unit.Contents)
+		if unit.Contents == nil {
```
ravisantoshgudimetla, line 320: AFAIK, it's good to have this guard but for Kubelet this will never be nil

LorbusChris: Contents is a pointer field in spec v3 -- I think we definitely want to guard against this here to avoid panics when it is nil for some unexpected reason.

internal/test/wmcb/wmcb_test.go
LorbusChris: Agree, we should still test it as long as the functionality is there. I'll add back a spec v2 test in the second commit.

LorbusChris:
@ravisantoshgudimetla I've added a third commit that restores the spec v2 e2e tests. PTAL

LorbusChris:
Note before un-holding: I'll have to update the vendor dir too now.

ravisantoshgudimetla:
What happened bot??

/lgtm 

again

PR #224, opened 2020-08-05
[e2e] Remove hybrid-overlay/kubelet release deps

aravindhp:
Remove the dependency on having an hybrid-overlay-node.exe and kubelet.exe being in a GitHub release by:
- Building it the hack script
- Copying the binaries to the Windows VM instead of remote downloading it from the release URL
- Move GetClusterVersion() and GetLatestWMCBRelease() from framework.Setup() into the wsuFramework.Setup()

### DIFF
--- a/internal/test/wsu/wsu_test.go
+++ b/internal/test/wsu/wsu_test.go
@@ -68,6 +68,13 @@ func (f *wsuFramework) Setup(vmCount int, credentials []*types.Credentials, skip
 		fmt.Errorf("framework setup failed: %v", err)
 		return err
 	}
+	if err := f.GetLatestWMCBRelease(); err != nil {
+		return fmt.Errorf("unable to get latest WMCB release: %v", err)
+	}
+
+	if err := f.GetClusterVersion(); err != nil {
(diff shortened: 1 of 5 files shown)


internal/test/wsu/wsu_test.go
```diff
@@ -68,6 +68,13 @@ func (f *wsuFramework) Setup(vmCount int, credentials []*types.Credentials, skip
 		fmt.Errorf("framework setup failed: %v", err)
 		return err
 	}
+	if err := f.GetLatestWMCBRelease(); err != nil {
```
ravisantoshgudimetla, line 71: Why do we need them in WSU?

aravindhp: The WSU e2e tests runs WSU twice, once with a released WMCB binary and then with a WMCB binary built from source.

mansikulkarni96:
Thanks for working on this @aravindhp 
/lgtm

aravindhp:
Thanks for the review @mansikulkarni96 
/hold cancel
/retest

PR #225, opened 2020-08-12
*: Update to Go 1.14

LorbusChris:
plus various dep wrangling


internal/test/go.mod
```diff
@@ -18,3 +18,5 @@ require (
 	k8s.io/apimachinery v0.18.6
 	k8s.io/client-go v0.18.2
 )
+
+replace github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer => ../../tools/windows-node-installer
```
ravisantoshgudimetla, line 22: IIRC, we wanted to pin to the old version of WNI. However it doesn't matter as long as the tests pass for now as @PratikMahajan  is working on removing it.

LorbusChris: Lmk if I should drop this commit :)

ravisantoshgudimetla: To be on the safe side, yes.

pkg/bootstrapper/bindata.go
```diff
@@ -1,5 +1,6 @@
-// Code generated for package bootstrapper by go-bindata DO NOT EDIT. (@generated)
-// sources:
+// Code generated by go-bindata. (@generated) DO NOT EDIT.
+
+ //Package bootstrapper generated by go-bindata.// sources:
```
LorbusChris, line 3: In case you're wondering: Yes this is indeed the header generated by go-bindata v3.1.3 (note the space at the beginning and the missing newline before `// sources`). I've PRed a fix upstream and hopefully it'll be pretty again with the next version

LorbusChris:
@aravindhp e2e-wmcb won't pass until https://github.com/openshift/release/pull/10955 is in, which in turn won't pass without the change in `hack/verify-gofmt.sh` from this PR here.. :chicken: :egg: 

Can you override?

internal/test/go.mod
LorbusChris: done

LorbusChris:
@aravindhp looks like it need the override again

PR #229, opened 2020-08-21
[WMCB] Replace WNI with MachineSet in WMCB tests

PratikMahajan:
This PR replaces the usage of WNI from WMCB tests to use MachineSets. 

As a part of this PR, WSU tests are being removed from master and future branches as we are
not planning on supporting it anymore. 

We will no longer be associating Public IP with Windows VMs and all operations will be performed using Private IP. 

We are following the following workflow for WMCB tests,
1. Build the image using Dockerfile.tools in images step
1. Use this built image to deploy test pod on host network in the test cluster and use private IP to interact with VMs
1. Create Machine Set
1. Identify machines in Provisioned State
1. Run the tests on host network 
1. Destroy machines after tests are completed

We have updated `dockerfile.tools` to build the new image for WMCB and use this image in tests.
Some additional changes include, building the kubelet and hybrid-overlay-node in the modified 
dockerfile.tools(we are currently doing this in hack script)

We have also changes the tests to use only SSH and move away from winRM. 

Also, changing the tests to use in-cluster pod permissions instead of mounting the kubeconfig, as we will no longer 
be mounting kubeconfig for the test pod. 

Modified the hack script to work with new modified workflow for tests. 

This is needed as we are moving away from WNI and no longer plan to support it. WNI should be removed in 
subsequent PR.


internal/test/framework/framework.go
```diff
@@ -73,91 +79,43 @@ type TestFramework struct {
...
-	if kubeconfig == "" {
-		return fmt.Errorf("KUBECONFIG environment variable not set")
-	}
-	awsCredentials = os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
-	if awsCredentials == "" {
-		return fmt.Errorf("AWS_SHARED_CREDENTIALS_FILE environment variable not set")
-	}
 	artifactDir = os.Getenv("ARTIFACT_DIR")
-	if awsCredentials == "" {
+	if artifactDir == "" {
```
mansikulkarni96, line 94: Is this check required, as we create. a temp ARTIFACT_DIR if the env variable is not set.

internal/test/clusterinfo/openshift.go
```diff
@@ -18,19 +16,10 @@ type OpenShift struct {
...
-	var rc *rest.Config
-	var err error
-
-	if kubeConfigPath == "" {
-		// InClusterConfig uses default service account or service account provided by the pod to obtain config.
-		rc, err = rest.InClusterConfig()
-	} else {
-		rc, err = clientcmd.BuildConfigFromFlags("", kubeConfigPath)
-	}
+// GetOpenShift creates client for the current OpenShift cluster. If KUBECONFIG env var is set, it is used to
```
mansikulkarni96, line 19: looks like doc needs to be updated since we are not checking for kubeconfig env var here.

internal/test/framework/framework.go
```diff
@@ -176,6 +134,35 @@ func (f *TestFramework) Setup(vmCount int, credentials []*types.Credentials, ski
 	if err := f.getLatestCniPluginsVersion(); err != nil {
 		return fmt.Errorf("unable to get latest 0.8.x version of CNI Plugins: %v", err)
 	}
+	if err := f.getKubeAPIClient(); err != nil {
```
mansikulkarni96, line 137: is there a difference between the client here and the one at line 122?

internal/test/framework/framework.go
```diff
@@ -215,22 +202,29 @@ func (f *TestFramework) getOpenShiftOperatorClient(config *restclient.Config) er
...
 func (f *TestFramework) getClusterAddress(config *restclient.Config) error {
-	if config.Host == "" {
-		return fmt.Errorf("API server has empty host name")
+	host, err := f.OSConfigClient.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to get cluster infrastructure resource: %v", err)
+	}
+	// get API server internal url of format https://api-int.abc.devcluster.openshift.com:6443
+	if host.Status.APIServerInternalURL == "" {
+		return fmt.Errorf("could not get host name for the kubernetes api server")
```
mansikulkarni96, line 211: could not get host name -> host url

internal/test/framework/framework.go
```diff
@@ -215,22 +202,29 @@ func (f *TestFramework) getOpenShiftOperatorClient(config *restclient.Config) er
...
 	}
 
-	clusterEndPoint, err := url.Parse(config.Host)
+	kubeAPIServerEndpoint := host.Status.APIServerInternalURL
```
mansikulkarni96, line 214: we are not using this var anywhere else can directly use the value in clusterEndPoint, err := url.Parse(host.Status.APIServerInternalURL)

internal/test/framework/framework.go
```diff
@@ -215,22 +202,29 @@ func (f *TestFramework) getOpenShiftOperatorClient(config *restclient.Config) er
...
+
+	clusterEndPoint, err := url.Parse(kubeAPIServerEndpoint)
 	if err != nil {
 		return fmt.Errorf("unable to parse the API server endpoint: %v", err)
 	}
 
 	hostName := clusterEndPoint.Hostname()
-	if !strings.HasPrefix(hostName, "api.") {
-		return fmt.Errorf("API server has invalid format: expected hostname to start with `api.`")
+	log.Print(hostName)
```
mansikulkarni96, line 222: can remove log or add additional info

internal/test/framework/framework.go
```diff
@@ -413,13 +407,13 @@ func (f *TestFramework) RetrieveArtifacts() {
 			continue
 		}
 
-		externalIP := vm.GetCredentials().GetIPAddress()
-		if len(externalIP) == 0 {
+		internalIP := vm.GetCredentials().GetIPAddress()
+		if len(internalIP) == 0 {
 			log.Printf("no external ip address found for the vm with instance ID %s", instanceID)
```
mansikulkarni96, line 412: no "internal" ip address found for the vm with instance ID

internal/test/framework/framework.go
```diff
@@ -492,17 +486,10 @@ func (f *TestFramework) TearDown() {
...
-			continue
-		}
-		if err := vm.Destroy(); err != nil {
-			log.Printf("failed tearing down the Windows VM %v with error: %v", vm, err)
-		} else {
-			// WNI will delete all the VMs in windows-node-installer.json so we need this to succeed only once
-			return
-		}
+	if err := f.Destroy(); err != nil {
+		log.Printf("failed delete MachineSets with error: %v", err)
```
mansikulkarni96, line 490: failed to delete

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,114 @@ type TestWindowsVM interface {
...
-
-	cloudProvider, err = cloudprovider.CloudProviderFactory(kubeconfig, awsCredentials, "default", artifactDir,
-		imageID, instanceType, sshKey, privateKeyPath)
+//createMachineSet() gets the generated MachineSet configuration from cloudprovider package and creates a MachineSet
+func (f *TestFramework) createMachineSet() error {
+	cloudProvider, err := providers.NewCloudProvider(sshKey)
+	if err != nil {
+		return fmt.Errorf("error instantiating cloud provider %v", err)
+	}
+	machineSet, err := cloudProvider.GenerateMachineSet(true, 1)
```
mansikulkarni96, line 61: why not initialize f.machineSet here?

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,114 @@ type TestWindowsVM interface {
...
-	// Wait for some time before starting configuring of ssh server. This is to let sshd service be available
-	// in the list of services
-	// TODO: Parse the output of the `Get-Service sshd, ssh-agent` on the Windows node to check if the windows nodes
-	// has those services present
-	if !skipSetup {
-		time.Sleep(time.Minute)
-		if err := w.ConfigureOpenSSHServer(); err != nil {
-			return w, fmt.Errorf("failed to configure OpenSSHServer on the Windows VM: %v", err)
-		}
+	return nil, fmt.Errorf("expected event count %d but got %d", vmCount, len(provisionedMachines))
```
mansikulkarni96, line 110: event count -> vm count

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,114 @@ type TestWindowsVM interface {
...
+		// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
+		providerTokens := strings.Split(providerID, "/")
+		instanceID := providerTokens[len(providerTokens)-1]
+		if len(instanceID) == 0 {
+			return nil, fmt.Errorf("empty instance id in provider id")
+		}
+		creds := types.NewCredentials(instanceID, ipAddress, types.Username)
+		winVM.Credentials = creds
+		log.Print("setting up ssh")
+		winVM.Credentials.SetSSHKey(f.Signer)
```
mansikulkarni96, line 157: we should have a nil or error check for the credentials we are setting up in this code block

internal/test/providers/cloudprovider.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func NewCloudProvider(sshKeyPair string) (CloudProvider, error) {
+	openshift, err := oc.GetOpenShift()
+	if err != nil {
+		return nil, errors.Wrap(err, "Getting OpenShift client failed")
+	}
+	cloudProvider, err := openshift.GetCloudProvider()
+	if err != nil {
+		return nil, errors.Wrap(err, "Getting cloud provider type")
```
mansikulkarni96, line 24: "failed to get cloud provider type"

internal/test/types/types.go
```diff
@@ -38,7 +33,7 @@ const (
 //		jira: https://issues.redhat.com/browse/WINC-239. Remove the fields related to SSHClient and WinrmClient and put
 // 		them in connectivity struct
 type Windows struct {
-	// Credentials to access the Windows VM created
+	// Creds is used for parsing the vmCreds command line argument
```
mansikulkarni96, line 36: Creds-> Credentials

internal/test/types/types.go
```diff
@@ -227,56 +132,56 @@ func (w *Windows) GetSSHClient() error {
...
-	for retries := 0; retries < 10; retries++ {
-		sshClient, err = ssh.Dial("tcp", w.Credentials.GetIPAddress()+":22", config)
-		if err == nil {
-			break
-		}
-		log.Printf("warning - SSH dial %s error: %v", w.Credentials.GetIPAddress(), err)
-		time.Sleep(1 * time.Minute)
-	}
-
+	sshClient, err := ssh.Dial("tcp", w.Credentials.GetIPAddress()+":22", config)
```
mansikulkarni96, line 139: w.SSHClient can be assigned here

internal/test/types/types.go
```diff
@@ -227,56 +132,56 @@ func (w *Windows) GetSSHClient() error {
...
 
 // GetIPAddress returns the ip address of the given node
 func (cred *Credentials) GetIPAddress() string {
 	return cred.ipAddress
 }
 
-// GetPassword returns the password associated with the given node
-func (cred *Credentials) GetPassword() string {
-	return cred.password
+// GetSSHKey returns the password associated with the given node
```
mansikulkarni96, line 177: doc needs updation

internal/test/providers/cloudprovider.go
```diff
@@ -0,0 +1,33 @@
...
+	oc "github.com/openshift/windows-machine-config-bootstrapper/internal/test/clusterinfo"
+	awsProvider "github.com/openshift/windows-machine-config-bootstrapper/internal/test/providers/aws"
+	"github.com/pkg/errors"
+)
+
+type CloudProvider interface {
+	GenerateMachineSet(bool, int32) (*mapi.MachineSet, error)
+}
+
+func NewCloudProvider(sshKeyPair string) (CloudProvider, error) {
```
aravindhp, line 17: You typically want to add methods in the commit where they get used. Yes, it might make the commit large but it implies that how you organize the commits needs to change. But given this is for e2e testing, I won't block on this change.

internal/test/framework/framework.go
```diff
@@ -31,23 +39,21 @@ const (
...
 	// WindowsLabel represents the node label that need to be applied to the Windows node created
 	WindowsLabel = "node.openshift.io/os_id=Windows"
-
-	// awsUsername is the default windows username on AWS
-	awsUsername = "Administrator"
 	// remoteLogPath is the directory where all the log files related to components that we need are generated on the
 	// Windows VM
 	remoteLogPath = "C:\\var\\log\\"
+	// PrivateKeyPath contains the path to the private key which is used in decrypting password in case of AWS
+	// cloud provider. This would have been mounted as a secret by user
```
aravindhp, line 46: We no longer use the key to decrypt the password. Please update the docs.

internal/test/framework/framework.go
```diff
@@ -31,23 +39,21 @@ const (
...
+	PrivateKeyPath = "/etc/private-key/private-key.pem"
+	// AWSCredentialsPath contains the path to the aws credentials to interact with AWS cloud provider.
```
aravindhp, line 48: Please be consistent. `AWS`.

internal/test/framework/framework.go
```diff
@@ -73,91 +79,43 @@ type TestFramework struct {
...
-		if err != nil {
-			return fmt.Errorf("unable to instantiate Windows VM: %v", err)
-		}
+	var config *rest.Config
+	var err error
+	if kubeconfig == "" {
+		// use in-cluster pod permissions
+		config, err = rest.InClusterConfig()
+	} else {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
```
aravindhp, line 116: Why is this needed? Won't be always using the in cluster config?

internal/test/framework/framework.go
aravindhp: I have the same question. It is unclear why you need both clients.

internal/test/framework/framework.go
```diff
@@ -229,8 +216,7 @@ func (f *TestFramework) getClusterAddress(config *restclient.Config) error {
 		return fmt.Errorf("API server has invalid format: expected hostname to start with `api.`")
 	}
 
-	// Replace `api.` with empty string for the first occurrence.
-	f.ClusterAddress = strings.Replace(hostName, "api.", "", 1)
```
aravindhp, line 233: Why is this no longer needed?

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,117 @@ type TestWindowsVM interface {
...
-	if credentials == nil {
-		windowsVM, err := cloudProvider.CreateWindowsVM()
+// waitForMachines() waits until all the machines required are in Provisioned state. It returns an array of all
+// the machines created. All the machines are created concurrently.
+func (f *TestFramework) waitForMachines(vmCount int) ([]mapi.Machine, error) {
+	log.Print("Waiting for Machine Sets ")
+	windowsOSLabel := "machine.openshift.io/os-id"
+	var provisionedMachines []mapi.Machine
+	// it takes approximately 12 minutes in the CI for all the machines to appear.
+	timeOut := 12 * time.Minute
```
aravindhp, line 82: Shouldn't the VM count be factored into this?

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,117 @@ type TestWindowsVM interface {
...
-		if err := w.ConfigureOpenSSHServer(); err != nil {
-			return w, fmt.Errorf("failed to configure OpenSSHServer on the Windows VM: %v", err)
-		}
+	return nil, fmt.Errorf("expected event count %d but got %d", vmCount, len(provisionedMachines))
+}
+
+// newWindowsVM creates and sets up a Windows VM in the cloud and returns the WindowsVM interface that can be used to
+// interact with the VM. If no error is returned then it is guaranteed that the VM was
+// created and can be interacted with.
+func (f *TestFramework) newWindowsVM(vmCount int) ([]TestWindowsVM, error) {
```
aravindhp, line 116: `newWindowsVM` implies one VM while you are creating `vmCount` number of them. This should rather be `newWindowsMachineSet`.

internal/test/framework/windowsvm.go
```diff
@@ -51,57 +52,117 @@ type TestWindowsVM interface {
...
+		creds := types.NewCredentials(instanceID, ipAddress, types.Username)
+		winVM.Credentials = creds
+		log.Print("setting up ssh and winrm")
+		winVM.Credentials.SetSSHKey(f.Signer)
+		if err := winVM.GetSSHClient(); err != nil {
+			return nil, fmt.Errorf("unable to get ssh client for vm %s : %v", instanceID, err)
+		}
+		if err := winVM.SetupWinRMClient(); err != nil {
+			return nil, fmt.Errorf("unable to setup winRM client for vm %s : %v", instanceID, err)
+		}
```
aravindhp: Why do we still need WinRM?

internal/test/framework/windowsvm.go
```diff
@@ -175,12 +236,15 @@ func (w *testWindowsVM) copyFileFrom(sftp *sftp.Client, remotePath, localPath st
 	return nil
 }
 
-func (w *testWindowsVM) Destroy() error {
-	// There is no VM to destroy
-	if cloudProvider == nil || w.Windows == nil || w.GetCredentials() == nil {
-		return nil
+// Destroy() deletes the MachineSet which in turn deletes all the Machines created by the MachineSet
+func (f *TestFramework) Destroy() error {
```
aravindhp, line 237: Please rename this to `DestroyMachineSet`.

internal/test/types/types.go
```diff
@@ -0,0 +1,271 @@
+package types
```
aravindhp, line 1: Why is this called types? This looks like the new `Windows` package and it implies that the older `Windows` package has been repurposed and needs to be renamed.

internal/test/types/types.go
```diff
@@ -0,0 +1,271 @@
...
+	"time"
+
+	"github.com/masterzen/winrm"
+	"github.com/pkg/sftp"
+	"golang.org/x/crypto/ssh"
+)
+
+// This package should have the types that will be used by component. For example, aws should have it's own
+// sub-package
+// TODO: Move every cloud provider types here
```
aravindhp: Why does the Windows package have to be cloud provider specific?

internal/test/types/types.go
```diff
@@ -0,0 +1,271 @@
...
+//		getters and setters for it so that it can be exposed as a public method
+//		jira: https://issues.redhat.com/browse/WINC-239. Remove the fields related to SSHClient and WinrmClient and put
+// 		them in connectivity struct
+type Windows struct {
+	// Creds is used for parsing the vmCreds command line argument
+	Credentials *Credentials
+	// SSHClient contains the ssh client information to access the Windows VM via ssh
+	SSHClient *ssh.Client
+	// WinrmClient to access the Windows VM created
+	WinrmClient *winrm.Client
```
aravindhp: Again, why is WinRM required?

internal/test/types/types.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return fmt.Errorf("error copying %s to the Windows VM: %v", filePath, err)
+	}
+
+	// Forcefully close it so that we can execute the binary later
+	dstFile.Close()
+	return nil
+}
+
+func (w *Windows) Run(cmd string, psCmd bool) (string, string, error) {
```
aravindhp: Now I see why WinRM is required. Please have a single `Run` method that uses SSH like we did in WMCO.

internal/test/types/types.go
```diff
@@ -0,0 +1,271 @@
...
+	winrmClient, err := winrm.NewClientWithParameters(endpoint, user, "", params)
+	if err != nil {
+		return fmt.Errorf("failed to set up winrm client with error: %v", err)
+	}
+	w.WinrmClient = winrmClient
+	return nil
+}
+
+// Credentials holds the information to access the Windows instance created.
+type Credentials struct {
```
aravindhp: Please make this a separate package.

internal/test/framework/windowsvm.go
```diff
@@ -153,14 +153,11 @@ func (f *TestFramework) newWindowsVM(vmCount int) ([]TestWindowsVM, error) {
...
 		winVM.Credentials = creds
-		log.Print("setting up ssh and winrm")
+		log.Print("setting up ssh")
 		winVM.Credentials.SetSSHKey(f.Signer)
 		if err := winVM.GetSSHClient(); err != nil {
 			return nil, fmt.Errorf("unable to get ssh client for vm %s : %v", instanceID, err)
 		}
-		if err := winVM.SetupWinRMClient(); err != nil {
-			return nil, fmt.Errorf("unable to setup winRM client for vm %s : %v", instanceID, err)
-		}
```
aravindhp: Removing WinRM should not be a separate commit. This will cause reviewers to ask questions like I did for the previous commit where you are moving the WinRM code around which  makes reviewing more difficult.

aravindhp (changes requested):
Thanks for working on this,  @PratikMahajan. Please address my comments.

internal/test/go.mod
```diff
@@ -2,19 +2,33 @@ module github.com/openshift/windows-machine-config-bootstrapper/internal/test
 
 go 1.14
 
+replace (
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200422081840-fdd1b0c14c88 // OpenShift 4.5
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200422192633-6f6c07fc2a70 // OpenShift 4.5
```
aravindhp, line 8: Shouldn't we be pinning to 4.6?

internal/test/go.mod
```diff
@@ -2,19 +2,33 @@ module github.com/openshift/windows-machine-config-bootstrapper/internal/test
 
 go 1.14
 
+replace (
```
aravindhp, line 5: Please make the commit, `[build] update dependencies`. Add the commands you ran to perform the update.

internal/test/go.mod
```diff
@@ -2,19 +2,33 @@ module github.com/openshift/windows-machine-config-bootstrapper/internal/test
...
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200422081840-fdd1b0c14c88 // OpenShift 4.5
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200422192633-6f6c07fc2a70 // OpenShift 4.5
+	k8s.io/api => k8s.io/api v0.18.2
+	k8s.io/apimachinery => k8s.io/apimachinery v0.18.2
+	k8s.io/client-go => k8s.io/client-go v0.18.2
```
aravindhp, line 11: Given that `v.19.0` has been released, shouldn't we pin to that? Otherwise we have to revisit and update later.

aravindhp (changes requested):

internal/test/framework/framework.go
```diff
@@ -202,18 +202,26 @@ func (f *TestFramework) getOpenShiftOperatorClient(config *restclient.Config) er
 // are running against. For example: the kubernetes API server endpoint https://api.abc.devcluster.openshift.com:6443
 // gets converted to abc.devcluster.openshift.com
 func (f *TestFramework) getClusterAddress(config *restclient.Config) error {
```
aravindhp, line 204: I don't follow what you mean by `this commit changes the WMCB to use internal ip instead of external ip
for retrieving the node.` We are using the API server internal endpoint and the private IP of the VM.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,79 @@ set -o errexit
...
+get_kubectl(){
+  # Download the kubectl binary only if it is not already available
+  # We do not validate the version of kubectl if it is available already
+  if type kubectl >/dev/null 2>&1; then
+    which kubectl
+    return
+  fi
+
+  KUBECTL_DIR=/tmp/kubectl
+  curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.2.2/openshift-client-linux-4.2.2.tar.gz -o openshift-origin-client-tools.tar.gz \
```
aravindhp, line 20: Any reason to pick an old version? We should at least use `4.5`.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,79 @@ set -o errexit
...
-    \? )
-      echo "Usage: $0 [-v] [-s]"
-      exit 0
-      ;;
-  esac
-done
-
-# TODO: Add steps to execute code that will kick off the e2e tests in CI
+# downloads the oc binary but we are just using kubectl as we dont need oc specific commands, oc binary depends on kubectl
+# using only kubectl binary reduces the hassle of moving both oc and kubectl to path. We can use kubectl directly from /tmp
```
aravindhp, line 10: I don't understand the reasoning to use `kubectl`. There is no reason to use `kubectl` just because we don't use `oc` features. I also don't follow the hassle of moving both binaries when only is being used. Moreover can't you use `oc` directly from `/tmp`.

PS: I am fine using `kubectl`. I just don't follow your reasoning for doing so.

internal/test/wmcb/deploy/deploy.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: windows-machine-config-bootstrapper
```
aravindhp, line 4: This should reflect we are running the tests. `wmcb-e2e-test` or something along this lines. Please reflect this elsewhere too.

internal/test/wmcb/deploy/deploy.yaml
```diff
@@ -0,0 +1,39 @@
...
+        name: windows-machine-config-bootstrapper
+    spec:
+      hostNetwork: true
+      volumes:
+      - name: cloud-private-key
+        secret:
+          secretName: cloud-private-key
+      - name: aws-creds
+        secret:
+          secretName: aws-creds
```
aravindhp, line 22: Why do we still need the creds?

Dockerfile.tools
```diff
@@ -1,47 +1,63 @@
...
 
-# Make Ansible happy with arbitrary UID/GID in OpenShift.
-RUN chmod g=u /etc/passwd /etc/group
+# Build WMCB unit and e2e tests
+RUN make build-wmcb-unit-test
+RUN make build-wmcb-e2e-test
 
-RUN mkdir -p /.ansible
-RUN chmod -R g=u+w /.ansible
+FROM golang:1.14.4 as testing
```
aravindhp, line 52: Why are we using a golang image for the final stage?

internal/test/wmcb/runTest.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+cd wmcb
```
aravindhp, line 3: You need to call out that  this cannot be executed locally and this script can only be run within the container.

internal/test/wmcb/runTest.sh
```diff
@@ -0,0 +1,10 @@
...
+
+# Transfer the files and run the unit and e2e tests
+if ! CGO_ENABLED=0 GO111MODULE=on go test -v -run=TestWMCB \
```
aravindhp, line 6: Is it not possible to build a test binary in the Dockerfile and transfer it? You can the specify the command in the deployment and remove this script.

aravindhp (changes requested):
Please update the README if the testing workflow has changed.

hack/run-wmcb-ci-e2e-test.sh
PratikMahajan: My bad, my assumption was that we need kubectl binary for oc binary to work. Just checked that oc works perfectly without kubectl as well.

hack/run-wmcb-ci-e2e-test.sh
PratikMahajan: took this snippet from previous dockerfile. Will update to use latest

internal/test/framework/windowsvm.go
PratikMahajan: in case we get an error machineSet will be set to nil, so assigning it after check. IMO, just noticed, i should be assigning it after creation is successful.

internal/test/framework/windowsvm.go
PratikMahajan: need to check that. As all VMs are created simultaneously..

internal/test/types/types.go
PratikMahajan: checking for error before setting the value.

internal/test/wmcb/deploy/deploy.yaml
PratikMahajan: We need aws creds to create machineSets. We are fetching the Subnet, etc

internal/test/wmcb/runTest.sh
PratikMahajan: will look into this and update.

Dockerfile.tools
```diff
@@ -1,47 +1,63 @@
...
+# Build hybrid-overlay-node.exe
+WORKDIR /build/
+RUN git clone https://github.com/openshift/ovn-kubernetes.git
+WORKDIR /build/ovn-kubernetes/go-controller/
+RUN make windows
+
+# Different step for building binaries other then kubelet and hybrid-overlay-node so as to make use of caching
+# while building the container. Building kubelet and hybrid-overlay-node takes considerable time and its not required
+# to build them for every run. We can force building kubelet and hybrid-overlay-node by not using using cache while
+# building the container image. Reduces the container build time considerably.
```
aravindhp, line 40: This sort of optimization is overkill. Moreover it is safer to always build kubelet and hybrid-overlay-node on every run to ensure we pick up changes.

Dockerfile.tools
aravindhp: Yes that is preferable

internal/test/clusterinfo/openshift.go
```diff
@@ -0,0 +1,66 @@
...
+
+// OpenShift is an Client struct which will be used for all OpenShift related functions to interact with the existing
+// Cluster.
+type OpenShift struct {
+	Client *clientset.Clientset
+}
+
+// GetOpenShift creates client for the current OpenShift cluster. If KUBECONFIG env var is set, it is used to
+// create client, otherwise it uses in-cluster config.
+func GetOpenShift() (*OpenShift, error) {
```
sebsoto, line 21: Shouldnt this be named something like `NewOpenShift`? Since it is constructing an instance of the `OpenShift` struct.

internal/test/clusterinfo/openshift.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if infra.Status == (v1.InfrastructureStatus{}) || infra.Status.PlatformStatus == nil {
+		return nil, fmt.Errorf("error getting infrastructure status")
+	}
+	return infra.Status.PlatformStatus, nil
+}
+
+// getInfrastructure returns the information of current Infrastructure referred by the OpenShift client or an error.
```
sebsoto: Comment doesnt make too much sense, this is also not great but i think it works better:
`getInfrastructure returns the Infrastructure object defining the configuration of the cluster's infrastructure`

internal/test/clusterinfo/openshift.go
```diff
@@ -0,0 +1,66 @@
...
+func (o *OpenShift) getInfrastructure() (*v1.Infrastructure, error) {
+	infra, err := o.Client.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return infra, nil
```
sebsoto: This can just be
```
return o.Client.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
```
as you are not wrapping any additional information in the error, and I dont think there is anything to add

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	"k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	awsprovider "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1beta1"
+)
+
+const (
+	infraIDTagKeyPrefix = "kubernetes.io/cluster/"
+	infraIDTagValue     = "owned"
+	windowsLabel        = "machine.openshift.io/os-id"
```
sebsoto, line 29: These need comments

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	// instanceType is the AWS specific instance type to create the VM with
+	instanceType = "m5a.large"
+)
+
+type awsProvider struct {
+	// imageID is the AMI image-id to be used for creating Virtual Machine
```
sebsoto, line 35: `imageID is the AMI image-id that is used to create new Virtual Machines`

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	// region in which the Machine needs to be created
+	region string
+	// sshKeyPair is the key pair associated with the Windows VM
+	sshKeyPair string
+}
+
+// newSession uses AWS credentials to create and returns a session for interacting with EC2.
+func newSession(credentialPath, credentialAccountID, region string) (*awssession.Session, error) {
+	if _, err := os.Stat(credentialPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to find AWS credentials from path '%v'", credentialPath)
```
sebsoto, line 54: You should be checking for errors in general. permission errors, etc.
Checking specifically if not exist is useful for if an entity can create the file in question if its missing, in this case this is not possible

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	credentialAccountID, instanceType, region, sshKeyPair string) (*awsProvider, error) {
+	session, err := newSession(credentialPath, credentialAccountID, region)
+	if err != nil {
+		return nil, fmt.Errorf("could not create new AWS session: %v", err)
+	}
+
+	ec2Client := ec2.New(session, aws.NewConfig())
+
+	iamClient := iam.New(session, aws.NewConfig())
+
```
sebsoto, line 76: unecessary newlines

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+		iamClient,
+		ec2Client,
+		openShiftClient,
+		region,
+		sshKeyPair,
+	}, nil
+}
+
+// SetupAWSCloudProvider creates AWS provider using the give OpenShift client
+// This is the first step of the e2e test and fails the test upon error.
```
sebsoto, line 92: I dont think `This is the first step of the e2e test and fails the test upon error.` should be said here, instead if it is not clear it should be commented at the call site

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+func SetupAWSCloudProvider(region, sshKeyPair string) (*awsProvider, error) {
+	oc, err := clusterinfo.GetOpenShift()
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize OpenShift client with error: %v", err)
+	}
+	// awsCredentials is set by OpenShift CI
+	awsCredentials := os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
+	if len(awsCredentials) == 0 {
```
sebsoto, line 100: nit: can just check if `== ""` for consistency with rest of these checks

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil, fmt.Errorf("error obtaining aws interface object: %v", err)
+	}
+	return awsProvider, nil
+}
+
+// getInfraID returns the infrastructure ID associated with the OpenShift cluster.
+func (a *awsProvider) getInfraID() (string, error) {
+	infraID, err := a.openShiftClient.GetInfrastructureID()
+	if err != nil {
+		return "", fmt.Errorf("erroring getting OpenShift infrastructure ID associated with the cluster")
```
sebsoto, line 114: Please wrap the error
also `erroring` -> `error`

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	searchFilter := ec2.Filter{Name: &windowsAMIFilterName, Values: []*string{&windowsAMIFilterValue}}
+
+	describedImages, err := ec2Client.DescribeImages(&ec2.DescribeImagesInput{
+		Filters: []*ec2.Filter{&searchFilter},
+		Owners:  []*string{&windowsAMIOwner},
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(describedImages.Images) < 1 {
```
sebsoto, line 140: nit: Could just check if len == 0

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil, fmt.Errorf("no instance offerings returned for %s", a.instanceType)
+	}
+
+	// Finding required subnet within the vpc.
+	foundSubnet := false
+	requiredSubnet := "-private-"
+
+	for _, subnet := range subnets.Subnets {
+		for _, tag := range subnet.Tags {
+			// TODO: find required subnet by checking igw gateway in routing.
```
sebsoto, line 211: Please create a ticket for this

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+	err = fmt.Errorf("could not find the required subnet in VPC: %v", *vpc.VpcId)
+	if !foundSubnet {
+		err = fmt.Errorf("could not find the required subnet in a zone that supports %s instance type",
+			a.instanceType)
+	}
+	return nil, err
+}
+
+// getClusterWorkerSGID gets worker security group id from the existing cluster or returns an error.
+// This function is exposed for testing purpose.
```
sebsoto, line 236: I think this line needs to be removed

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+				Values: aws.StringSlice([]string{infraIDTagValue}),
+			},
+			{
+				Name:   aws.String("state"),
+				Values: aws.StringSlice([]string{"available"}),
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
```
sebsoto, line 274: Please wrap this error with context

sebsoto (changes requested):

internal/test/wmcb/wmcb_test.go
```diff
@@ -191,35 +182,36 @@ func (vm *wmcbVM) runTestConfigureCNI(t *testing.T) {
...
 
 	return nil
 }
 
 // remoteDownload downloads the tar file in url to the remoteDownloadFile location and checks if the SHA matches
 func (vm *wmcbVM) remoteDownload(pkg PkgInfo, remoteDownloadFile string) error {
-	_, stderr, err := vm.Run("if (!(Test-Path "+remoteDownloadFile+")) { wget "+pkg.getUrl()+" -o "+remoteDownloadFile+" }",
-		true)
+	useragent := "Wget/1.13.4"
+	cmd := wgetIgnoreCertCmd + " -server \"" + pkg.getUrl() + "\" -output \"" + remoteDownloadFile + "\" -useragent \"" + useragent + "\""
```
LorbusChris, line 211: the `-useragent` flag of `wgetIgnoreCertCmd` has been replaced with an `-acceptHeader` flag.

You can probably just pass it `*/*` as accept header here.

PratikMahajan: thanks @LorbusChris  ! Noticed this after failed test. Changing it !

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,404 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	awssession "github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+	"github.com/aws/aws-sdk-go/service/iam"
+	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	"github.com/openshift/windows-machine-config-bootstrapper/internal/test/clusterinfo"
+	"k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
LorbusChris: I'd suggest `corev1`, `configv1` and `metav1` for these common api types

internal/test/providers/cloudprovider.go
LorbusChris: All of this looks like boilerplate code that also exists in the operator repo. There is a lot of duplication going on here.

internal/test/providers/aws/aws.go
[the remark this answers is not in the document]
aravindhp: We have been using `core`, unless we need to use both v1 and v2 APIs at the same time. Doing this reduces code churn if we want to move from v1 to v2 minus the breaking API spots.

internal/test/framework/framework.go
PratikMahajan: because we are getting the hostname in the format that we intend to use. so there is no more `api.` at the start.

internal/test/framework/windowsvm.go
PratikMahajan: i don't think we need to check error for getters and setters.

Dockerfile.tools
```diff
@@ -1,47 +1,63 @@
...
+RUN git clone https://github.com/openshift/kubernetes.git
+WORKDIR /build/kubernetes/
+RUN KUBE_BUILD_PLATFORMS=windows/amd64 make WHAT=cmd/kubelet
+
+# Build hybrid-overlay-node.exe
+WORKDIR /build/
+RUN git clone https://github.com/openshift/ovn-kubernetes.git
+WORKDIR /build/ovn-kubernetes/go-controller/
+RUN make windows
+
```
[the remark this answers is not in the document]
PratikMahajan: i'd prefer a follow up story

internal/test/framework/framework.go
PratikMahajan: One is kubernetes clientset of type `kubernetes.Clientset` to query the cluster status and other is kubernetes client of type  `client.Client` to interact with kubernetes API servers

internal/test/providers/cloudprovider.go
PratikMahajan: We decided to copy the code in design meet as we dont want to import code from operator repo in WMCB test. That explains a different commit.

internal/test/framework/framework.go
PratikMahajan: changed the commit message to make it more clear

LorbusChris (changes requested):
just a few minor things.

internal/test/credentials/credentials.go
```diff
@@ -0,0 +1,80 @@
+package credentials
+
+import (
+	"fmt"
+	"golang.org/x/crypto/ssh"
+	"strings"
```
aravindhp, line 6: Let us go with @LorbusChris suggestion to group imports with new files. Note: I am not asking you to make this change in existing files.

internal/test/framework/framework.go
aravindhp: I still don't follow. Aren't you talking to the API server in both cases?

internal/test/providers/cloudprovider.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/openshift/api/config/v1"
+	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	oc "github.com/openshift/windows-machine-config-bootstrapper/internal/test/clusterinfo"
+	awsProvider "github.com/openshift/windows-machine-config-bootstrapper/internal/test/providers/aws"
+	"github.com/pkg/errors"
+)
+
+type CloudProvider interface {
```
aravindhp, line 13: I just noticed this stutters, i.e. `providers.CloudProvider`. Did you do this to maintain parity with WMCO? If not, `Cloud` should suffice here.

internal/test/framework/windowsvm.go
aravindhp: Did you get a chance to check on this?

internal/test/credentials/credentials.go
```diff
@@ -0,0 +1,80 @@
...
+
+// GetInstanceID returns the instanceId associated with the given node
+func (cred *Credentials) GetInstanceId() string {
+	return cred.instanceID
+}
+
+// GetUserName returns the user name associated with the given node
+func (cred *Credentials) GetUserName() string {
+	return cred.user
+}
```
aravindhp, line 55: You can remove the `Get` from the getters as per Go guidelines.

internal/test/framework/windowsvm.go
```diff
@@ -38,70 +44,128 @@ type TestWindowsVM interface {
...
-	//		jira: https://issues.redhat.com/browse/WINC-243
-	Destroy() error
 	// BuildWMCB returns the value of buildWMCB. It can be used by WSU to decide if it should build WMCB before using it
 	BuildWMCB() bool
 	// SetBuildWMCB sets the value of buildWMCB. Setting buildWMCB to true would indicate WSU will build WMCB instead of
 	// downloading the latest as per the cluster version. False by default
 	SetBuildWMCB(bool)
 	// Compose the Windows VM we have from WNI
-	types.WindowsVM
+	windows.WindowsVM
```
aravindhp, line 53: This is a bit confusing now. You have both  the struct and the interface in here. Why is that?

internal/test/windows/windows.go
```diff
@@ -0,0 +1,137 @@
...
+	// of stdout and stderr. If the bool is set, it implies that the cmd is to be execute in PowerShell. This function
+	// should be used in scenarios where you want to execute a command that runs in the background. In these cases we
+	// have observed that Run() returns before the command completes and as a result killing the process.
+	Run(string, bool) (string, error)
+	// GetCredentials returns the interface for accessing the VM credentials. It is up to the caller to check if non-nil
+	// Credentials are returned before usage.
+	GetCredentials() *credentials.Credentials
+	// Reinitialize re-initializes the Windows VM. Presently only the ssh client is reinitialized.
+	Reinitialize() error
+}
```
aravindhp, line 44: This design pattern does not look right and needs to be revisited.

Dockerfile.tools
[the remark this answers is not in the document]
aravindhp: Sounds good. Please create the story and link it here.

Dockerfile.tools
```diff
@@ -1,47 +1,74 @@
...
 
-# Install ansible and required packages
-RUN pip2 install ansible pywinrm
+RUN mkdir /build/
+WORKDIR /build/
 
-# Make Ansible happy with arbitrary UID/GID in OpenShift.
-RUN chmod g=u /etc/passwd /etc/group
+# Build kubelet.exe
+RUN git clone https://github.com/openshift/kubernetes.git
```
aravindhp, line 27: Please use `git clone --branch release-4.6 --single-branch`. Applies to all repos you are cloning.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,78 @@ set -o errexit
...
+sed -i "s~ARTIFACT_DIR_VALUE~${ARTIFACT_DIR}~g" internal/test/wmcb/deploy/deploy.yaml
+sed -i "s~REPLACE_IMAGE~${WMCB_IMAGE}~g" internal/test/wmcb/deploy/deploy.yaml
+
+# deploy the test pod on test cluster
+if ! $OC apply -f internal/test/wmcb/deploy/deploy.yaml -n default; then
+    echo "application already deployed"
+fi
+
+# Sleep for 1 minute while the deployment is being created.
+sleep 1m
```
aravindhp, line 69: Please wait for the deployment to be ready rather than a sleep.

internal/test/wmcb/runTest.sh
aravindhp: Please update.

aravindhp (changes requested):

internal/test/credentials/credentials.go
```diff
@@ -0,0 +1,80 @@
...
+// Set populates the list of credentials from the vmCreds command line argument
+func (c *Credentials) Set(value string) error {
+	if value == "" {
+		return nil
+	}
+
+	splitValue := strings.Split(value, ",")
+	// Credentials consists of three elements, so this has to be
+	if len(splitValue) != 2 {
+		return fmt.Errorf("incomplete VM credentials provided")
```
VaishnaviHire, line 71: `incorrect value for VM credentials` ?

internal/test/framework/windowsvm.go
```diff
@@ -38,70 +44,128 @@ type TestWindowsVM interface {
...
+	}
+	machineSet, err := cloudProvider.GenerateMachineSet(true, 1)
+	if err != nil {
+		return fmt.Errorf("error creating Windows MachineSet: %v", err)
+	}
+	log.Print("Creating Machine Sets")
+	err = f.client.Create(context.TODO(), machineSet)
 	if err != nil {
-		return nil, fmt.Errorf("error instantiating cloud provider %v", err)
+		return fmt.Errorf("unable to create MachineSet %v", err)
```
VaishnaviHire, line 69: `error creating MachineSet`

internal/test/framework/windowsvm.go
```diff
@@ -38,70 +44,128 @@ type TestWindowsVM interface {
...
-	} else {
-		//TODO: Add username as well, as it will change depending on cloud provider
-		if credentials.GetIPAddress() == "" || credentials.GetPassword() == "" {
-			return nil, fmt.Errorf("password or IP address not specified in credentials")
+		provisionedMachines = []mapi.Machine{}
+
+		phaseProvisioned := "Provisioned"
+
+		for _, machine := range allMachines.Items {
+			instanceStatus := machine.Status
```
VaishnaviHire, line 97: Can we combine these conditions to one -
`if instanceStatus.Phase != nil && *instanceStatus.Phase == phaseProvisioned{
    append()
}`

internal/test/framework/windowsvm.go
```diff
@@ -38,70 +44,128 @@ type TestWindowsVM interface {
...
-	if err := w.SetupWinRMClient(); err != nil {
-		return w, fmt.Errorf("failed to setup winRM client for the Windows VM: %v", err)
+// newWindowsMachineSet creates and sets up Windows VMs in the cloud and returns the WindowsVM interface that can be used to
+// interact with the VM. If no error is returned then it is guaranteed that the VM was
+// created and can be interacted with.
+func (f *TestFramework) newWindowsMachineSet(vmCount int) ([]TestWindowsVM, error) {
+	w := make([]TestWindowsVM, vmCount)
+	err := f.createMachineSet()
+	if err != nil {
+		return nil, fmt.Errorf("unable to create Windows MachineSet: %v", err)
```
VaishnaviHire, line 121: `error creating Windows MachineSet`

VaishnaviHire (commented):
Thanks for working on this @PratikMahajan . PTAL at my comments

ravisantoshgudimetla (commented):
Mostly LGTM. I have some questions around implementation, please address them

internal/test/framework/windowsvm.go
PratikMahajan: yes, actually 12 min should be enough for all VMS to come up. 
They actually come up in parallel.

internal/test/credentials/credentials.go
PratikMahajan: actually the condition checks if the credentials are in a pair of 2. So incomplete will better describe this error. incorrect might mean that the credentials provided are not correct (and not incomplete). This might cause confusion.

internal/test/providers/cloudprovider.go
PratikMahajan: Actually this is coming directly from the WMCO repo, kept this as it is on purpose. Should i change it to cloud ?

aravindhp: No, given you are doing it to maintain parity with WMCO.

internal/test/providers/aws/aws.go
PratikMahajan: basically what we are doing here is that , we are checking if the file exists and in case it does not exist it'll throw a `*PathError`. With `os.IsNotExist()` takes the error and returns a boolean indicating whether the error is known to report that a file or directory does not exist. Our intention here is to find if the credentials file is mounted. I am not sure about a different way to implement this.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,78 @@ set -o errexit
...
+get_oc(){
+  # Download the oc binary only if it is not already available
+  # We do not validate the version of oc if it is available already
+  if type oc >/dev/null 2>&1; then
+    which oc
+    return
+  fi
+
+  OC_DIR=/tmp/oc
+  curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.5.8/openshift-client-linux-4.5.8.tar.gz -o openshift-origin-client-tools.tar.gz \
```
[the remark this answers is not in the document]
PratikMahajan: actually this needs to be done from the hack script as we are directly deploying the pod on the host network. 
basically, no go code is run when we are doing this and in order to deploy we need to apply it using oc command.

internal/test/framework/framework.go
```diff
@@ -196,22 +196,27 @@ func (f *TestFramework) getOpenShiftOperatorClient(config *restclient.Config) er
...
-	if config.Host == "" {
-		return fmt.Errorf("API server has empty host name")
+	host, err := f.OSConfigClient.ConfigV1().Infrastructures().Get(context.TODO(), "cluster", metav1.GetOptions{})
```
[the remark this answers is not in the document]
PratikMahajan: right. Adding a comment here

internal/test/wmcb/runTest.sh
PratikMahajan: sorry, missed this. 
actually i am not sure how that will work. 
you want me to create a binary using `-c` and then use this binary to run the tests on the pod? 
in that case, i am not sure if it will transfer the required files to the VM

hack/run-wmcb-ci-e2e-test.sh
PratikMahajan: replaced the 1 minute sleep by a while loop which waits for the pod to enter running state and starts streaming logs once it enters a running state.

internal/test/wmcb/runTest.sh
aravindhp: Yes create the binary like other test binaries. And you will need to transfer the binary as part of building the test container image. But I am willing to let it go at this point.

internal/test/providers/aws/aws.go
aravindhp: So what happens if you get a different error? You will just proceed and fail later?

internal/test/providers/aws/aws.go
aravindhp: We need a ticket to address this separate from creating a test framework. That ticket can be blocked by the test framework ticket, so that we have fix it in only one spot.

internal/test/credentials/credentials.go
aravindhp: I would then say that is a parsing error.

internal/test/framework/framework.go
aravindhp: This should have been brought up at the stand up then

internal/test/framework/windowsvm.go
```diff
@@ -21,172 +23,135 @@ const (
...
-// interact with the VM. If credentials are passed then it is assumed that VM already exists in the cloud and those
-// credentials will be used to interact with the VM. If no error is returned then it is guaranteed that the VM was
-// created and can be interacted with. If skipSetup is true, then configuration steps are skipped.
-func newWindowsVM(imageID, instanceType string, credentials *types.Credentials, skipSetup bool) (TestWindowsVM, error) {
-	w := &testWindowsVM{}
-	var err error
-
-	cloudProvider, err = cloudprovider.CloudProviderFactory(kubeconfig, awsCredentials, "default", artifactDir,
-		imageID, instanceType, sshKey, privateKeyPath)
+//createMachineSet() gets the generated MachineSet configuration from cloudprovider package and creates a MachineSet
```
aravindhp, line 38: Missing space after `//`

internal/test/wmcb/main_test.go
```diff
@@ -19,14 +19,14 @@ var (
...
 func TestMain(m *testing.M) {
-	var vmCreds e2ef.Creds
+	var vmCreds credentials.Credentials
 	var skipVMSetup bool
 
 	flag.Var(&vmCreds, "vmCreds", "List of VM credentials")
 	flag.BoolVar(&skipVMSetup, "skipVMSetup", false, "Option to disable setup in the VMs")
 	flag.Parse()
 
-	err := framework.Setup(vmCount, vmCreds, skipVMSetup)
```
aravindhp, line 29: Why are you removing `skipVMSetup`?

internal/test/wmcb/powershell/wget-ignore-cert.ps1
```diff
@@ -24,4 +24,5 @@ public static class Dummy {
 }
 [System.Net.ServicePointManager]::ServerCertificateValidationCallback = [dummy]::GetDelegate()
 
+$ProgressPreference = "SilentlyContinue"
```
aravindhp, line 27: Why are you making this change?

internal/test/wmcb/runTest.sh
aravindhp: This too should have been brought up during the standup.

aravindhp (changes requested):

internal/test/wmcb/main_test.go
PratikMahajan: added skipVMSetup back. removed vmCreds as we can no more take vm creds. Instead using the existing private key to access the VMs created.

internal/test/wmcb/powershell/wget-ignore-cert.ps1
PratikMahajan: After switching to SSH, the powershell script stopped working. as mentioned in the commit message made this change to make sure the script runs successfully. Was getting the following error before this change:
```
wget : Win32 internal error "Access is denied" 0x5 occurred while reading the console output buffer. Contact Microsoft 
Customer Support Services.
At C:\Temp\wget-ignore-cert.ps1:27 char:1
+ wget -UserAgent $useragent $server -o $output
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ReadError: (:) [Invoke-WebRequest], HostException
    + FullyQualifiedErrorId : ReadConsoleOutput,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```

internal/test/providers/aws/aws.go
PratikMahajan: if there is some other error that causes the requests to AWS to error out. It'll fail later.

internal/test/wmcb/powershell/wget-ignore-cert.ps1
aravindhp: If this change is actually needed, it cannot go in as part of this PR. Anyhow it looks like #232 will make this irrelevant.

Dockerfile.tools
```diff
@@ -1,47 +1,82 @@
...
+COPY . .
+
+# Build WMCB unit and e2e tests
+RUN make build-wmcb-unit-test
+RUN make build-wmcb-e2e-test
 
-# Explicitly set GOCACHE environment variable
+# Build TestWMCB binary to run the tests on the Windows VM created
+WORKDIR /build/internal/test/wmcb/
+RUN CGO_ENABLED=0 GO111MODULE=on go test -c -v -run=TestWMCB -timeout=30m . -o TestWMCB.test
```
aravindhp, line 46: Please rename `TestWMCB.test` to `test-wmcb`

docs/README.md
```diff
@@ -60,27 +60,23 @@ The following environment variables need to be set for running the end to end te
...
-   to `Administrator`. This allows you to run the tests against existing set of VMs.
-   ```shell script
-   $ hack/run-wmcb-ci-e2e-test.sh -v"aws-instance-id-1,3.135.234.23,password,aws-instance-id-2,3.135.234.23,password"
-   ```
-
-- `-s` option allows you to skip the framework setup. The assumption here is that the framework setup has already been
-  run on the VM.
-  ```shell script
-  $ hack/run-wmcb-ci-e2e-test.sh -v"aws-instance-id,1.2.34.23,password" -s
-  ```
```
aravindhp, line 82: Is this no longer needed or skippable?

hack/run-wmcb-ci-e2e-test.sh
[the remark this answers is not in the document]
aravindhp: We should eventually move to SteveK's [recommendation](https://issues.redhat.com/browse/DPTP-1467?focusedCommentId=14424357&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14424357)

Dockerfile.tools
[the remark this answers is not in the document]
aravindhp: This is not addressed

aravindhp (changes requested):
Partial review.

Dockerfile.tools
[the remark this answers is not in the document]
PratikMahajan: https://issues.redhat.com/browse/WINC-490

docs/README.md
PratikMahajan: We can no longer accept credentials in the form of username or password. So thats no more needed. 
as far as skipping framework setup goes, we can do that through `job.yaml`, and in that case, we will be using the mounted private key to access the VM.

internal/test/wmcb/powershell/wget-ignore-cert.ps1
PratikMahajan: As #232 is closed, should i create another PR for this change ?

internal/test/providers/aws/aws.go
PratikMahajan: https://issues.redhat.com/browse/WINC-491
https://issues.redhat.com/browse/WINC-492

internal/test/wmcb/powershell/wget-ignore-cert.ps1
aravindhp: Yes

docs/README.md
aravindhp: I was asking about skipping the framework setup. How are we going to do this through `job.yaml`? I am assuming we have to do it manually. Moreover, can we still reuse an existing `Machine`? I think not so this point becomes moot.

internal/test/wmcb/main_test.go
aravindhp: So given `vmCreds` cannot be used what is the use of  `skipVMSetup`? Or is there a way for us to use an existing `VM` / `Machine` that is not immediately obvious to me?

And please note, I asked `Why are you removing skipVMSetup`. I did not request you to add it back. Please think things through before making changes to a question.

internal/test/wmcb/powershell/wget-ignore-cert.ps1
aravindhp: Does this block your PR from merging? If yes, please leave it as it is. Ideally this should have been part of a commit that migrated from WinRM to SSH.

Dockerfile.tools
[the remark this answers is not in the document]
aravindhp: Please link to the story here as a comment.

docs/README.md
aravindhp: Please correct me if there is a way to use an existing VM / `Machine`.

internal/test/wmcb/deploy/role.yaml
```diff
@@ -0,0 +1,46 @@
...
+rules:
+  - apiGroups:
+    - ""
+    resources:
+    - nodes
+    verbs:
+    - "*"
+  # The infrastructure endpoint is used within WNI
+  - apiGroups:
+    - "config.openshift.io"
```
aravindhp, line 15: No need for quotes. Applies to all `apiGroups`.

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,96 @@ set -o errexit
...
+    && tar -xzf openshift-origin-client-tools.tar.gz \
+    && mv oc $OC_DIR \
+    && rm -rf ./openshift*
+
+  echo $OC_DIR
+}
+
+# delete_deployment deletes the deployed test pod
+delete_deployment(){
```
aravindhp, line 25: Please replace with `delete_job`

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,96 @@ set -o errexit
...
+  if ! $1 delete -f internal/test/wmcb/deploy/job.yaml -n default; then
+    echo "no deployment found"
```
aravindhp, line 27: `No job found`

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -3,57 +3,96 @@ set -o errexit
...
+if ! $OC apply -f internal/test/wmcb/deploy/rolebinding.yaml -n default; then
+    echo "rolebinding already present"
+fi
+
+sed -i "s~ARTIFACT_DIR_VALUE~${ARTIFACT_DIR}~g" internal/test/wmcb/deploy/job.yaml
+sed -i "s~REPLACE_IMAGE~${WMCB_IMAGE}~g" internal/test/wmcb/deploy/job.yaml
+
+# deploy the test pod on test cluster
+if ! $OC apply -f internal/test/wmcb/deploy/job.yaml -n default; then
+    echo "application already deployed"
```
aravindhp, line 62: `job already deployed`

aravindhp (changes requested):

docs/README.md
PratikMahajan: actually, if the required machine has all the labels that we add, and if it is in provisioned state. We can actually use that machine to run tests.. 
after i removed skipVMSetup option, every time i had to run the tests the wait would be atleast 12 min. 
we can create a machineset and machine before hand, and we can just delete existing VM attached as a node and reuse the new VM created. This will reduce the wait time to just 1 minute. 
Will bring this up in stamdup

internal/test/wmcb/powershell/wget-ignore-cert.ps1
PratikMahajan: yes, 
this should go in first in order for this PR to merge. 
this was originally part of the commit that migrated from WinRM to SSH. But, as that commit was after this commit. i had to squash it to reduce confusion while reviewing.

aravindhp: Yes, you fist migrated to MachineSets and then migrated to WinRM to SSH. You should have opened a separate PR to first move from WinRM to SSH or at least made it your first commit in this PR.

Dockerfile.tools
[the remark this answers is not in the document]
PratikMahajan: linked.

docs/README.md
PratikMahajan: added detailed information to Readme.

internal/test/providers/aws/aws.go
PratikMahajan: linked

aravindhp (approved):
Thanks for working on this, @PratikMahajan 

/approve

aravindhp:
@PratikMahajan the e2e cluster tests have been consistently failing. Please look into that.

internal/test/framework/windowsvm.go
```diff
@@ -1,16 +1,18 @@
...
-	"os"
-	"path/filepath"
+	"strings"
 	"time"
 
-	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/cloudprovider"
-	"github.com/openshift/windows-machine-config-bootstrapper/tools/windows-node-installer/pkg/types"
-	"github.com/pkg/sftp"
+	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	"github.com/openshift/windows-machine-config-bootstrapper/internal/test/credentials"
```
mansikulkarni96, line 11: internal imports can be separated here

internal/test/framework/windowsvm.go
```diff
@@ -21,172 +23,157 @@ const (
...
+	f.machineSet = machineSet
+	log.Printf("Created Machine Set %v", machineSet.Name)
+	return nil
+}
 
-	if credentials == nil {
-		windowsVM, err := cloudProvider.CreateWindowsVM()
+// waitForMachines() waits until all the machines required are in Provisioned state. It returns an array of all
+// the machines created. All the machines are created concurrently.
+func (f *TestFramework) waitForMachines(vmCount int, skipVMSetup bool) ([]mapi.Machine, error) {
```
mansikulkarni96, line 60: nit: can be renamed along the lines of getMachineSets as it does more than just waiting for them to be provisioned

internal/test/framework/windowsvm.go
```diff
@@ -21,172 +23,157 @@ const (
...
-	// in the list of services
-	// TODO: Parse the output of the `Get-Service sshd, ssh-agent` on the Windows node to check if the windows nodes
-	// has those services present
-	if !skipSetup {
-		time.Sleep(time.Minute)
-		if err := w.ConfigureOpenSSHServer(); err != nil {
-			return w, fmt.Errorf("failed to configure OpenSSHServer on the Windows VM: %v", err)
+	if skipVMSetup {
+		if vmCount == 0 {
+			return nil, fmt.Errorf("no VMs found")
```
mansikulkarni96, line 90: no Windows* VMs found

internal/test/providers/aws/aws.go
```diff
@@ -0,0 +1,403 @@
...
+// getInfraID returns the infrastructure ID associated with the OpenShift cluster.
+func (a *awsProvider) getInfraID() (string, error) {
+	infraID, err := a.openShiftClient.GetInfrastructureID()
+	if err != nil {
+		return "", fmt.Errorf("error getting OpenShift infrastructure ID associated with the cluster")
+	}
+	return infraID, nil
+}
+
+// getLatestWindowsAMI returns the imageid of the latest released "Windows Server with Containers" image
```
mansikulkarni96, line 117: imageID*

internal/test/providers/cloudprovider.go
```diff
@@ -0,0 +1,33 @@
+package providers
+
+import (
+	"fmt"
+
+	"github.com/openshift/api/config/v1"
+	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	oc "github.com/openshift/windows-machine-config-bootstrapper/internal/test/clusterinfo"
+	awsProvider "github.com/openshift/windows-machine-config-bootstrapper/internal/test/providers/aws"
```
mansikulkarni96, line 9: internal imports can be seperated

mansikulkarni96 (commented):
Thanks @PratikMahajan, mostly LGTM with some minor comments.

ravisantoshgudimetla (approved):
/lgtm

Thank you for working on this @PratikMahajan

internal/test/framework/windowsvm.go
PratikMahajan: changed it to `getWindowsMachines()`

internal/test/framework/windowsvm.go
PratikMahajan: done

internal/test/providers/cloudprovider.go
PratikMahajan: done

PR #232, opened 2020-09-14
[bootstrapper] Add ignition-url flag to initialize-kubelet cmd

LorbusChris:
The flag can be used to specify the URL of the Machine Config Server,
which serves Ignition config.

Adds code to the WMCB for requesting and receiving ignition config from
the server, obsoleting the `wget-ignore-cert.ps1` script.


pkg/bootstrapper/bootstrapper.go
```diff
@@ -114,6 +116,8 @@ type winNodeBootstrapper struct {
 	// ignitionFilePath is the path to the ignition file which is used to set up worker nodes
 	// https://github.com/coreos/ignition/blob/spec2x/doc/getting-started.md
 	ignitionFilePath string
```
aravindhp, line 118: Do we need to maintain both options? Or will this be removed once we update the WMCO to use the newer option?

aravindhp (approved):

/approve

ravisantoshgudimetla (commented):
Thank you for working on this @LorbusChris. While the change LGTM, I am wondering what benefit we get out this. To me, bootstrapper shouldn't have anything to do with downloading ignition file, I've always seen bootstrapper as an entity which looks at the input files given to it and act upon them. The main reason I am sort of nack on this is we're assuming bootstrapper would always access to hostNetwork access. The downloading should be part of another entity which has access to MCS. Having this separation would also help us run bootstrapper in situations where MCS blocks serving ignition files on hostNetwork like in case of UPI or `Bring Your Own Hosts`.

pkg/bootstrapper/bootstrapper.go
LorbusChris: We have to keep `--ignition-file-path` at least until we change WMCO to use the new one.
After that I think we could remove it safely.

LorbusChris:
@ravisantoshgudimetla the main reason for this is to get rid of the powershell script, and do everything in go code, with just one binary (the wmcb executable) and with just one vm.Run SSH command, instead of two as before. Other than that, the behavior shouldn't change with this.

Instead of adding this as a flag, this could also be a separate sub-command in the same binary - however I don't really see a benefit in splitting the Ignition download out into a separate step:
One still has to pass the MCS endpoint either way, whether it be used by the wget script or the wmcb binary. I don't think that would be any different in the UPI case.

ravisantoshgudimetla:
> the main reason for this is to get rid of the powershell script, and do everything in go code, with just one binary (the wmcb executable) and with just one vm.Run SSH command, instead of two as before. Other than that, the behavior shouldn't change with this.

To be clear, that functionality is outside the scope of WMCB, the functionality of downloading ignition file as of now is part of test code not wmcb

LorbusChris:
@ravisantoshgudimetla yes, right now it is not part of WMCB itself, but rather lives in the `wget-ignore-cert` powershell script, which is co-located with the bootstrapper binary, and which we have duplicated between here and the WMCO repo.

This PR also doesn't remove the WMCB's functionality to read the Ignition config from a file - it just adds functionality that will allow us to skip the powershell script invocation.

aravindhp:
/approve cancel

Based on the earlier discussion, we decided to download the ignition file from MCS in WMCO. I suggest we remove `wget-ignore-cert.ps1` from the WMCB test code once #229 goes in and have the test pod download the ignition file and transfer it over.

LorbusChris:
Closing in favor of https://github.com/openshift/windows-machine-config-operator/pull/112

/close

PR #233, opened 2020-09-17
[wmcb] add service dependency and service dependent restart

mansikulkarni96:
This PR adds service dependent stopping and starting in Windows services. It also adds a dependency from kubelet on underlying docker service.

### DIFF
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -489,10 +489,11 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 		StartType:    mgr.StartAutomatic,
 		ErrorControl: 0,
 		// Path to kubelet.exe
-		BinaryPathName:   filepath.Join(wmcb.installDir, "kubelet.exe"),
-		LoadOrderGroup:   "",
-		TagId:            0,
-		Dependencies:     nil,
+		BinaryPathName: filepath.Join(wmcb.installDir, "kubelet.exe"),
+		LoadOrderGroup: "",
+		TagId:          0,
+		// set dependency on docker
+		Dependencies:     []string{"docker"},
 		ServiceStartName: "",
 		DisplayName:      "",
 		Password:         "",
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -19,7 +19,6 @@ import (
 	ignitionCfgv3Types "github.com/coreos/ignition/v2/config/v3_1/types"
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
-	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 )
 
@@ -37,13 +36,15 @@ const (
 	// KubeletServiceName is the name will we run the kubelet Windows service under. It is required to be named "kubelet":
 	// https://github.com/kubernetes/kubernetes/blob/v1.16.0/cmd/kubelet/app/init_windows.go#L26
 	KubeletServiceName = "kubelet"
+	// kubeletDependentSvc is the name of the service dependent on kubelet Windows service
+	kubeletDependentSvc = "hybrid-overlay-node"
 	// kubeletSystemdName is the name of the systemd service that the kubelet runs under,
 	// this is used to parse the kubelet args
 	kubeletSystemdName = "kubelet.service"
 	// kubeletPauseContainerImage is the location of the image we will use for the kubelet pause container
 	kubeletPauseContainerImage = "mcr.microsoft.com/oss/kubernetes/pause:1.3.0"
-	// serviceWaitTime is an arbitrary amount of time to wait for the Windows service API to complete requests
-	serviceWaitTime = time.Second * 10
+	// serviceWaitTime is amount of wait time required for the Windows service API to complete stop requests
+	serviceWaitTime = time.Second * 20
 	// certDirectory is where the kubelet will look for certificates
 	certDirectory = "c:\\var\\lib\\kubelet\\pki\\"
 	// cloudConfigOption is kubelet CLI option for cloud configuration
@@ -117,8 +118,8 @@ type winNodeBootstrapper struct {
 	//initialKubeletPath is the path to the kubelet that we'll be using to bootstrap this node
 	initialKubeletPath string
 	// TODO: When more services are added consider decomposing the services to a separate Service struct with common functions
-	// kubeletSVC is a pointer to the kubelet Windows service object
-	kubeletSVC *mgr.Service
+	// kubeletSVC is a pointer to the kubeletService struct
+	kubeletSVC *kubeletService
 	// svcMgr is used to interact with the Windows service API
 	svcMgr *mgr.Mgr
 	// installDir is the directory the the kubelet service will be installed
@@ -178,9 +179,16 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 		}
 	}
 
+	var dependents []*mgr.Service
 	// If there is already a kubelet service running, find it
 	if ksvc, err := svcMgr.OpenService(KubeletServiceName); err == nil {
-		bootstrapper.kubeletSVC = ksvc
+		if dependentSvc, err := svcMgr.OpenService(kubeletDependentSvc); err == nil {
+			dependents = append(dependents, dependentSvc)
+		}
+		bootstrapper.kubeletSVC, err = newKubeletService(ksvc, dependents)
+		if err != nil {
+			return nil, fmt.Errorf("could not initialize struct kubeletService: %v", err)
+		}
 	}
 	return &bootstrapper, nil
 }
@@ -490,96 +498,25 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 		Password:         "",
 		Description:      "OpenShift Kubelet",
 	}
-	wmcb.kubeletSVC, err = wmcb.svcMgr.CreateService(KubeletServiceName, filepath.Join(wmcb.installDir, "kubelet.exe"), c, kubeletArgs...)
+	ksvc, err := wmcb.svcMgr.CreateService(KubeletServiceName, filepath.Join(wmcb.installDir, "kubelet.exe"), c, kubeletArgs...)
 	if err != nil {
 		return err
 	}
-	err = wmcb.kubeletSVC.SetRecoveryActions([]mgr.RecoveryAction{
-		{Type: mgr.ServiceRestart, Delay: 5},
-	}, 600)
-	if err != nil {
-		return err
+
+	var dependents []*mgr.Service
+	if dependentSvc, err := wmcb.svcMgr.OpenService(kubeletDependentSvc); err == nil {
+		dependents = append(dependents, dependentSvc)
 	}
+	wmcb.kubeletSVC, err = newKubeletService(ksvc, dependents)
+	if err != nil {
+		return fmt.Errorf("could not initialize struct kubeletService: %v", err)
+	}
+
+	if err := wmcb.kubeletSVC.setRecoveryActions(); err != nil {
+		return fmt.Errorf("failed to set recovery actions for Windows service %s", KubeletServiceName)
+	}
+
 	return nil
-}
-
-// startKubeletService starts the kubelet as a Windows service
-func (wmcb *winNodeBootstrapper) startKubeletService() error {
-	if wmcb.kubeletSVC == nil {
-		return fmt.Errorf("no kubelet service")
-	}
-	err := wmcb.kubeletSVC.Start()
-	if err != nil {
-		return err
-	}
-	return nil
-}
-
-// removeKubeletService deletes the kubelet service via the Windows service API
-func (wmcb *winNodeBootstrapper) removeKubeletService() error {
-	if wmcb.kubeletSVC == nil {
-		return nil
-	}
-	return wmcb.kubeletSVC.Delete()
-}
-
-// controlService sends a signal to the service and waits until it changes state in response to the signal
-func (wmcb *winNodeBootstrapper) controlService(cmd svc.Cmd, desiredState svc.State) error {
-	status, err := wmcb.kubeletSVC.Control(cmd)
-	if err != nil {
-		return err
-	}
-	// Most of the rest of the function borrowed from the package (golang.org/x/sys/windows/svc/mgr) example
-	// Arbitrary wait time
-	timeout := time.Now().Add(serviceWaitTime)
-	for status.State != desiredState {
-		if timeout.Before(time.Now()) {
-			return fmt.Errorf("timeout waiting for service to go to state=%d", desiredState)
-		}
-		time.Sleep(300 * time.Millisecond)
-		status, err = wmcb.kubeletSVC.Query()
-		if err != nil {
-			return fmt.Errorf("could not retrieve service status: %v", err)
-		}
-	}
-	return nil
-}
-
-// stopKubeletService stops the kubelet via the Windows service API
-func (wmcb *winNodeBootstrapper) stopKubeletService() error {
-	if wmcb.kubeletSVC == nil {
-		return nil
-	}
-
-	kubeletIsRunning, err := wmcb.isKubeletServiceRunning()
-	if err != nil {
-		return fmt.Errorf("unable to check if kubelet service is running: %v", err)
-	}
-
-	if !kubeletIsRunning {
-		return nil
-	}
-
-	return wmcb.controlService(svc.Stop, svc.Stopped)
-}
-
-// isKubeletServiceRunning returns true if the kubelet service is running
-func (wmcb *winNodeBootstrapper) isKubeletServiceRunning() (bool, error) {
-	status, err := wmcb.kubeletSVC.Query()
-	if err != nil {
-		return false, err
-	}
-	return status.State == svc.Running, nil
-}
-
-// StopAndRemoveServices stops and removes the kubelet service
-func (wmcb *winNodeBootstrapper) StopAndRemoveServices() error {
-	if wmcb.kubeletSVC == nil {
-		return nil
-	}
-	wmcb.stopKubeletService()
-	return wmcb.removeKubeletService()
-
 }
 
 // refreshServiceManager will disconnect and reconnect from the Windows service API. In order to complete certain
@@ -595,30 +532,13 @@ func (wmcb *winNodeBootstrapper) refreshServiceManager() error {
 	return err
 }
 
-// refreshKubeletService updates the kubelet service with the given config and restarts the service
-func (wmcb *winNodeBootstrapper) refreshKubeletService(config mgr.Config) error {
-	if err := wmcb.stopKubeletService(); err != nil {
-		return fmt.Errorf("error stopping kubelet service: %v", err)
-	}
-
-	if err := wmcb.kubeletSVC.UpdateConfig(config); err != nil {
-		return fmt.Errorf("error updating kubelet service: %v", err)
-	}
-
-	if err := wmcb.startKubeletService(); err != nil {
-		return fmt.Errorf("error starting kubelet service: %v", err)
-	}
-
-	return nil
-}
-
 // InitializeKubelet performs the initial kubelet configuration. It sets up the install directory, creates the kubelet
 // service, and then starts the kubelet service
 func (wmcb *winNodeBootstrapper) InitializeKubelet() error {
 	var err error
 	if wmcb.kubeletSVC != nil {
 		// if the kubelet service exists, we silently remove it and continue, to preserve idempotency
-		err = wmcb.StopAndRemoveServices()
+		err = wmcb.kubeletSVC.stopAndRemove()
 		if err != nil {
 			return err
 		}
@@ -636,7 +556,7 @@ func (wmcb *winNodeBootstrapper) InitializeKubelet() error {
 	if err != nil {
 		return fmt.Errorf("failed to create kubelet windows service: %v", err)
 	}
-	err = wmcb.startKubeletService()
+	err = wmcb.kubeletSVC.start()
 	if err != nil {
 		return fmt.Errorf("failed to start kubelet windows service: %v", err)
 	}
@@ -657,11 +577,11 @@ func (wmcb *winNodeBootstrapper) Configure() error {
 	}
 
 	// Stop the kubelet service as there could be open file handles from kubelet.exe on the plugin files
-	if err := wmcb.stopKubeletService(); err != nil {
+	if err := wmcb.kubeletSVC.stop(); err != nil {
 		return fmt.Errorf("unable to stop kubelet service: %v", err)
 	}
 
-	config, err := wmcb.kubeletSVC.Config()
+	config, err := wmcb.kubeletSVC.config()
 	if err != nil {
 		return fmt.Errorf("error getting kubelet service config: %v", err)
 	}
@@ -671,7 +591,7 @@ func (wmcb *winNodeBootstrapper) Configure() error {
 		return fmt.Errorf("error configuring kubelet service for CNI: %v", err)
 	}
 
-	if err = wmcb.refreshKubeletService(config); err != nil {
+	if err = wmcb.kubeletSVC.refresh(config); err != nil {
 		return fmt.Errorf("unable to refresh kubelet service: %v", err)
 	}
 
@@ -680,11 +600,8 @@ func (wmcb *winNodeBootstrapper) Configure() error {
 
 // Disconnect removes all connections to the Windows service manager api, and allows services to be deleted
 func (wmcb *winNodeBootstrapper) Disconnect() error {
-	if wmcb.kubeletSVC != nil {
-		err := wmcb.kubeletSVC.Close()
-		if err != nil {
-			return err
-		}
+	if err := wmcb.kubeletSVC.disconnect(); err != nil {
+		return err
 	}
 	err := wmcb.svcMgr.Disconnect()
 	wmcb.svcMgr = nil
--- a/pkg/bootstrapper/kubelet.go
+++ b/pkg/bootstrapper/kubelet.go
@@ -1,0 +1,163 @@
+package bootstrapper
+
+import (
+	"fmt"
+	"time"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+// kubeletService struct contains the kubelet specific service information
+type kubeletService struct {
+	// obj is a pointer to the Windows service object
+	obj *mgr.Service
+	// dependents contains a list of services dependent on the current service
+	dependents []*mgr.Service
+}
+
+// newKubeletService creates and returns a new kubeletService object
+func newKubeletService(ksvc *mgr.Service, dependents []*mgr.Service) (*kubeletService, error) {
+	if ksvc == nil {
+		return nil, fmt.Errorf("service object should not be nil")
+	}
+	return &kubeletService{
+		obj:        ksvc,
+		dependents: dependents,
+	}, nil
+}
+
+// config retrieves service config from service object Config()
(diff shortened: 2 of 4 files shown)


pkg/bootstrapper/bootstrapper.go
```diff
@@ -117,10 +117,7 @@ type winNodeBootstrapper struct {
 	//initialKubeletPath is the path to the kubelet that we'll be using to bootstrap this node
 	initialKubeletPath string
 	// TODO: When more services are added consider decomposing the services to a separate Service struct with common functions
-	// kubeletSVC is a pointer to the kubelet Windows service object
-	kubeletSVC *mgr.Service
-	// svcMgr is used to interact with the Windows service API
-	svcMgr *mgr.Mgr
+	service *service
```
VaishnaviHire, line 120: Missing doc

pkg/bootstrapper/bootstrapper.go
```diff
@@ -537,48 +571,51 @@ func (wmcb *winNodeBootstrapper) controlService(cmd svc.Cmd, desiredState svc.St
...
+	}
+
+	return nil
 }
 
-// isKubeletServiceRunning returns true if the kubelet service is running
-func (wmcb *winNodeBootstrapper) isKubeletServiceRunning() (bool, error) {
-	status, err := wmcb.kubeletSVC.Query()
+// isServiceRunning returns true if the service is running
+func (wmcb *winNodeBootstrapper) isServiceRunning() (bool, error) {
```
VaishnaviHire, line 642: Instead of returning `bool` and `error` should we just return `error` if the service is not running and `nil` otherwise?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -537,48 +571,51 @@ func (wmcb *winNodeBootstrapper) controlService(cmd svc.Cmd, desiredState svc.St
...
-func (wmcb *winNodeBootstrapper) StopAndRemoveServices() error {
-	if wmcb.kubeletSVC == nil {
+// StopAndRemoveServices stops and removes the Windows service
+func (wmcb *winNodeBootstrapper) StopAndRemoveServices(serviceName string) error {
+	if wmcb.service.svcObj == nil {
 		return nil
 	}
-	wmcb.stopKubeletService()
-	return wmcb.removeKubeletService()
+	wmcb.stopService(serviceName)
```
VaishnaviHire, line 655: Why are we not capturing the error here ?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -178,13 +183,24 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
...
-	// If there is already a kubelet service running, find it
-	if ksvc, err := svcMgr.OpenService(KubeletServiceName); err == nil {
-		bootstrapper.kubeletSVC = ksvc
+	bootstrapper.service, err = newService(svcMgr)
+	if err != nil {
+		return nil, err
 	}
+
+	// If there is already a kubelet service running, find it and assign it
+	bootstrapper.service.setService(KubeletServiceName)
```
VaishnaviHire, line 193: Why are we not capturing the error here ?

VaishnaviHire (commented):
Thanks @mansikulkarni96 for working on this. PTAL at my comments

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: the way we use this method is we check for the bool value and decide if there is anything to be done in case of stopping and starting a service, if the service is already in expected state we return nil. eg: If the service is not running and we want to stop it, we could just return nil as the variable value is false(returning an error because its not running doesn't fit here)

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: the reason is (// If there is already a kubelet service running, find it and assign it) if there isn't a kubelet service already present we don't need to make sure it is assigned here. It will get created in createService and be assigned there.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -593,13 +594,50 @@ func (wmcb *winNodeBootstrapper) stopService(serviceName string) error {
...
 }
 
+// getDependingServices returns a list of services depending on the given service
+func (wmcb *winNodeBootstrapper) getDependingServices(serviceName string) ([]string, error) {
+	var svcDependents []string
+	services, err := wmcb.service.svcMgr.ListServices()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list all services: %v", err)
+	}
+	if len(services) != 0 {
```
aravindhp, line 623:
```suggestion
	if len(services) == 0 {
	    return nil, nil
	}
	// rest of the code here with less indentation
```

pkg/bootstrapper/bootstrapper.go
```diff
@@ -641,9 +679,12 @@ func (wmcb *winNodeBootstrapper) refreshKubeletService(config mgr.Config) error
...
 	}
-
-	if err := wmcb.startService(); err != nil {
-		return fmt.Errorf("error starting kubelet service: %v", err)
+	// starting the services in dependency list will also start the services they are dependent on
+	for _, dependentSvc := range wmcb.service.svcDependents {
+		wmcb.service.setService(dependentSvc)
+		if err := wmcb.startService(); err != nil {
+			return fmt.Errorf(" error starting dependent service %s for kubelet service", dependentSvc)
+		}
```
aravindhp, line 687: For each service you find, you should be instantiating a new service object and calling `start` on it. This again looks a bit odd to me otherwise.

aravindhp (changes requested):
Thanks for working on this PR, @mansikulkarni96. Please address my comments.

pkg/bootstrapper/bootstrapper.go
sebsoto: I think we should be careful about doing this, what if there is a kubelet service running and we cannot open it for whatever reason? (permissions or something).
We would need to capture that error. 

If setService is being used in this way I think it should either be fed a service object, or we should check that the service exists before calling it, then only call it if the service does exist, and error out if it returns an error

pkg/bootstrapper/bootstrapper.go
```diff
@@ -178,13 +184,24 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
...
+
+	// If there is already a kubelet service running, find it and assign it
+	bootstrapper.service.setService(KubeletServiceName)
 	return &bootstrapper, nil
 }
 
+// newService takes the service manager and returns a new Windows service object
+func newService(mgr *mgr.Mgr) (*service, error) {
+	if mgr == nil {
+		return nil, fmt.Errorf("insuficient arguments to create new service object")
```
sebsoto, line 200: nit: would be more descriptive to say `service manager is nil`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -116,11 +116,8 @@ type winNodeBootstrapper struct {
 	ignitionFilePath string
 	//initialKubeletPath is the path to the kubelet that we'll be using to bootstrap this node
 	initialKubeletPath string
-	// TODO: When more services are added consider decomposing the services to a separate Service struct with common functions
-	// kubeletSVC is a pointer to the kubelet Windows service object
-	kubeletSVC *mgr.Service
-	// svcMgr is used to interact with the Windows service API
-	svcMgr *mgr.Mgr
+	// service struct holds all service specific information
+	service *service
```
sebsoto, line 120: It seems like this holds all the service specific information strictly for the kubelet service

pkg/bootstrapper/bootstrapper.go
```diff
@@ -503,29 +526,41 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 	return nil
 }
 
-// startKubeletService starts the kubelet as a Windows service
-func (wmcb *winNodeBootstrapper) startKubeletService() error {
-	if wmcb.kubeletSVC == nil {
-		return fmt.Errorf("no kubelet service")
+// startService starts a Windows service
+func (wmcb *winNodeBootstrapper) startService() error {
```
sebsoto, line 530: This function confuses me. It says it `starts a Windows service`, but doesn't give input for which service.
I feel like this should be a method of the `service` object. It can then be renamed to `start()`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -503,29 +526,41 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 		return err
 	}
 	return nil
 }
 
-// removeKubeletService deletes the kubelet service via the Windows service API
-func (wmcb *winNodeBootstrapper) removeKubeletService() error {
-	if wmcb.kubeletSVC == nil {
+// removeService deletes the Windows service via the Windows service API
+func (wmcb *winNodeBootstrapper) removeService() error {
```
sebsoto, line 551: Same as the comment on `startService`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -132,6 +129,16 @@ type winNodeBootstrapper struct {
...
 
+// service struct contains the service information
+type service struct {
+	// svcObj is a pointer to the Windows service object
+	svcObj *mgr.Service
+	// svcMgr is used to interact with the Windows service API
+	svcMgr *mgr.Mgr
+	// svcDependents contains a list of services dependent on the current service
+	svcDependents []string
+}
```
sebsoto, line 140: This struct is written like it is meant for a generic service, however it includes the service manager within it. If we had multiple services, we end up with multiple service managers. IMO this struct should either become specific to the kubelet service, or have the manager moved out of it.

sebsoto (changes requested):

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: I am capturing the error for setService(), it is not possible to do it in this particular call as the service is initialized here and service object will always be nil, hence we cannot validate it here.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -593,13 +594,50 @@ func (wmcb *winNodeBootstrapper) stopService(serviceName string) error {
...
+	if err != nil {
+		return err
+	}
+
+	if len(svcDependents) != 0 {
+		for _, svcDependent := range svcDependents {
+			wmcb.service.svcDependents = append(wmcb.service.svcDependents, svcDependent)
+			wmcb.stopService(svcDependent)
+		}
+	}
```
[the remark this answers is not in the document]
mansikulkarni96: I am calling the stop service recursively hence I am creating a new list here which gets updated on each call, also added some more comments around it.

LorbusChris (approved):

pkg/bootstrapper/service.go
```diff
@@ -90,6 +117,28 @@ func (service *kubeletService) controlService(cmd svc.Cmd, desiredState svc.Stat
 	return nil
 }
 
+// controlDependentService sends a control signal to the kubelet dependent service
```
aravindhp, line 120: `kubelet's dependent service`

Please call out that you are only going one level deep. i.e. you are not recursively checking the dependencies.

pkg/bootstrapper/service.go
```diff
@@ -65,6 +65,33 @@ func (service *kubeletService) startKubeletService() error {
 	if err := service.obj.Start(); err != nil {
 		return err
 	}
+
+	if len(service.svcDependents) != 0 {
```
aravindhp, line 69: Same trick mentioned before can be done here:
```suggestion
	if len(service.svcDependents) == 0 {
	    return nil
	}
	// rest of the code with less indentation
```

pkg/bootstrapper/service.go
```diff
@@ -90,6 +117,28 @@ func (service *kubeletService) controlService(cmd svc.Cmd, desiredState svc.Stat
 	return nil
 }
 
+// controlDependentService sends a control signal to the kubelet dependent service
+func controlDependentService(serviceObj *mgr.Service, cmd svc.Cmd, desiredState svc.State) error {
+	status, err := serviceObj.Control(cmd)
+	if err != nil {
+		return err
+	}
+	// Most of the rest of the function borrowed from the package (golang.org/x/sys/windows/svc/mgr) example
```
aravindhp, line 126: Is it possible to link to the example itself?

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+import (
+	"fmt"
+	"time"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+// kubeletService struct contains the kubelet specific service information
+type kubeletService struct {
```
ravisantoshgudimetla, line 12: Technically you specialized here but the fields are generic enough. However there is no problem with this approach considering we just have a single service for now.

pkg/bootstrapper/service.go
```diff
@@ -90,6 +117,28 @@ func (service *kubeletService) controlService(cmd svc.Cmd, desiredState svc.Stat
...
 }
 
+// controlDependentService sends a control signal to the kubelet dependent service
+func controlDependentService(serviceObj *mgr.Service, cmd svc.Cmd, desiredState svc.State) error {
+	status, err := serviceObj.Control(cmd)
+	if err != nil {
+		return err
+	}
+	// Most of the rest of the function borrowed from the package (golang.org/x/sys/windows/svc/mgr) example
+	// Arbitrary wait time
```
aravindhp, line 127: Is it? There is definitely a time you are waiting here. Only `serviceWaitTime` is arbitrary.

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// stopAndRemoveServices stops and removes the kubelet service
+func (service *kubeletService) stopAndRemoveServices() error {
+	if service.obj == nil {
+		return nil
+	}
+	service.stopKubeletService()
+	return service.removeKubeletService()
+
```
sebsoto, line 232: extra newline

aravindhp (changes requested):
Thanks for working on this, @mansikulkarni96. Please address my comments. I am leaving the struct design review to @ravisantoshgudimetla and @sebsoto.

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+func (service *kubeletService) isKubeletServiceRunning() (bool, error) {
+	status, err := service.obj.Query()
+	if err != nil {
+		return false, err
+	}
+	return status.State == svc.Running, nil
+}
+
+// isDependentRunning returns true if the kubelet dependent service is running
+func isDependentRunning(serviceObj *mgr.Service) (bool, error) {
```
sebsoto, line 217: This mentions kubelet but seems generic

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+	status, err := serviceObj.Query()
+	if err != nil {
+		return false, err
+	}
+	return status.State == svc.Running, nil
+}
+
+// stopAndRemoveServices stops and removes the kubelet service
+func (service *kubeletService) stopAndRemoveServices() error {
```
sebsoto, line 226: I think this should be renamed to `stopAndRemove()` the service is implicit as it is a method of the kubeletService struct

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+// removeKubeletService deletes the kubelet service via the Windows service API
+func (service *kubeletService) removeKubeletService() error {
+	if service.obj == nil {
+		return nil
+	}
+	return service.obj.Delete()
+}
+
+// isKubeletServiceRunning returns true if the kubelet service is running
+func (service *kubeletService) isKubeletServiceRunning() (bool, error) {
```
sebsoto, line 208: This can be renamed to `isRunning()` as it is a method of kubeletService

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := service.startKubeletService(); err != nil {
+		return fmt.Errorf("error starting kubelet service: %v", err)
+	}
+
+	return nil
+}
+
+// removeKubeletService deletes the kubelet service via the Windows service API
+func (service *kubeletService) removeKubeletService() error {
```
sebsoto, line 200: Can be renamed to remove() as it is a method of kubelet service

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+		err := controlDependentService(serviceObj, svc.Stop, svc.Stopped)
+		if err != nil {
+			return fmt.Errorf("unable to stop %s dependent service", serviceObj.Name)
+		}
+	}
+	return nil
+}
+
+// refreshKubeletService updates the kubelet service with the given config and restarts the service
+func (service *kubeletService) refreshKubeletService(config mgr.Config) error {
```
sebsoto, line 183: Can be renamed to refresh() as it is a method of kubelet service

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := service.controlService(svc.Stop, svc.Stopped); err != nil {
+		return fmt.Errorf("unable to stop Windows Service %s", KubeletServiceName)
+	}
+
+	return nil
+}
+
+// stopDependentService stops the kubelet dependent service
+func stopDependentService(serviceObj *mgr.Service) error {
```
sebsoto, line 168: Mentions kubelet service but this seems generic.

in addition should this be renamed to `stopDependantServices`? As there could be more than one dependant service

pkg/bootstrapper/service.go
sebsoto: Does this function actually check the dependants, or does it just check the status of whats passed into it?
If the latter, this can be renamed to `isRunning`.

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+		status, err = serviceObj.Query()
+		if err != nil {
+			return fmt.Errorf("could not retrieve service status: %v", err)
+		}
+	}
+	return nil
+}
+
+// stopKubeletService stops the Windows service via the Windows service API
+func (service *kubeletService) stopKubeletService() error {
```
sebsoto, line 143: Can be renamed to stop() as it is a method of kubelet service

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+		status, err = service.obj.Query()
+		if err != nil {
+			return fmt.Errorf("could not retrieve service status: %v", err)
+		}
+	}
+	return nil
+}
+
+// controlDependentService sends a control signal to the kubelet dependent service
+func controlDependentService(serviceObj *mgr.Service, cmd svc.Cmd, desiredState svc.State) error {
```
sebsoto, line 121: Mentions kubelet but seems generic

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+			if err != nil {
+				return fmt.Errorf("unable to start dependent service for %s", KubeletServiceName)
+			}
+		}
+	}
+	return nil
+}
+
+// startDependentService starts the kubelet dependent Windows service
+func startDependentService(serviceObj *mgr.Service) error {
```
sebsoto, line 81: Does this function actually start the dependants, or does it just start the service of whats passed into it?
If the latter, this should be renamed to startService. A function should be named according to what it does, not how its intended to be used

pkg/bootstrapper/service.go
```diff
@@ -0,0 +1,258 @@
...
+func (service *kubeletService) getConfig() (mgr.Config, error) {
+	config, err := service.obj.Config()
+	if err != nil {
+		return mgr.Config{}, err
+	}
+	return config, nil
+}
+
+// startKubeletService starts the kubelet as a Windows service
+func (service *kubeletService) startKubeletService() error {
```
sebsoto, line 53: can be renamed to `start()` as its a method of kubeletService

sebsoto (changes requested):

pkg/bootstrapper/kubelet.go
```diff
@@ -0,0 +1,245 @@
...
+// stop ensures that the kubelet service is stopped
+func (k *kubeletService) stop() error {
+	isServiceRunning, err := k.isRunning()
+	if err != nil {
+		return fmt.Errorf("unable to check if kubelet service is running: %v", err)
+	}
+	if !isServiceRunning {
+		return nil
+	}
+	// the list of dependents is static here and contains one level of dependencies
```
mansikulkarni96, line 98: mentioning the approach here is not recursive/goes one level deep

pkg/bootstrapper/bootstrapper.go
```diff
@@ -37,13 +36,15 @@ const (
...
+	// kubeletDependentSvc is the name of the service dependent on kubelet Windows service
+	kubeletDependentSvc = "hybrid-overlay-node"
 	// kubeletSystemdName is the name of the systemd service that the kubelet runs under,
 	// this is used to parse the kubelet args
 	kubeletSystemdName = "kubelet.service"
 	// kubeletPauseContainerImage is the location of the image we will use for the kubelet pause container
 	kubeletPauseContainerImage = "mcr.microsoft.com/oss/kubernetes/pause:1.3.0"
 	// serviceWaitTime is an arbitrary amount of time to wait for the Windows service API to complete requests
-	serviceWaitTime = time.Second * 10
+	serviceWaitTime = time.Second * 20
```
mansikulkarni96, line 47: timeout required now to stop two services and refresh.

aravindhp: Please update the comment to reflect that.

pkg/bootstrapper/kubelet.go
aravindhp: Please mention it in the `stop()` docs too. A reader of the docs should be able to see that.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -481,107 +489,37 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 		// Path to kubelet.exe
-		BinaryPathName:   filepath.Join(wmcb.installDir, "kubelet.exe"),
-		LoadOrderGroup:   "",
-		TagId:            0,
-		Dependencies:     nil,
+		BinaryPathName: filepath.Join(wmcb.installDir, "kubelet.exe"),
+		LoadOrderGroup: "",
+		TagId:          0,
+		// set dependency on docker
+		Dependencies:     []string{"docker"},
```
aravindhp, line 495: Is this service going to be named `docker` across cloud providers / Windows images?

pkg/bootstrapper/kubelet.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, fmt.Errorf("service object should not be nil")
+	}
+	return &kubeletService{
+		obj:        ksvc,
+		dependents: dependents,
+	}, nil
+}
+
+// getConfig retrieves service config from service object Config()
+func (k *kubeletService) getConfig() (mgr.Config, error) {
```
aravindhp, line 31: Sorry I should have caught this before. Shouldn't this be just `Config()` as per go naming conventions?

pkg/bootstrapper/kubelet.go
```diff
@@ -52,6 +52,16 @@ func (k *kubeletService) start() error {
...
 		return err
 	}
+
+	if len(k.dependents) == 0 {
+		return nil
+	}
+	for _, dependent := range k.dependents {
+		err := startService(dependent)
+		if err != nil {
+			return fmt.Errorf("unable to start dependent service for %s", KubeletServiceName)
```
aravindhp, line 62: Please mention the dependent service in the error message.

pkg/bootstrapper/kubelet.go
```diff
@@ -161,3 +179,76 @@ func (k *kubeletService) setRecoveryActions() error {
...
+func controlService(serviceObj *mgr.Service, cmd svc.Cmd, desiredState svc.State) error {
+	if serviceObj == nil {
+		return fmt.Errorf("service object should not be nil")
+	}
+	status, err := serviceObj.Control(cmd)
+	if err != nil {
+		return err
+	}
+	// Most of the rest of the function borrowed from the package (golang.org/x/sys/windows/svc/mgr) example
+	// Arbitrary wait time
```
aravindhp, line 211: Please mention what that arbitrary wait time is.

aravindhp (changes requested):

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: done

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: The name docker is independent of cloud providers, the package management msft used to install docker package on any Windows server images, installs them with service named "docker"

pkg/bootstrapper/kubelet.go
mansikulkarni96: done

pkg/bootstrapper/kubelet.go
mansikulkarni96: done

pkg/bootstrapper/kubelet.go
mansikulkarni96: done

pkg/bootstrapper/kubelet.go
mansikulkarni96: done

pkg/bootstrapper/bootstrapper.go
```diff
@@ -481,107 +489,37 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
 		StartType:    mgr.StartAutomatic,
 		ErrorControl: 0,
 		// Path to kubelet.exe
-		BinaryPathName:   filepath.Join(wmcb.installDir, "kubelet.exe"),
-		LoadOrderGroup:   "",
-		TagId:            0,
-		Dependencies:     nil,
+		BinaryPathName: filepath.Join(wmcb.installDir, "kubelet.exe"),
+		LoadOrderGroup: "",
+		TagId:          0,
```
aravindhp: Nit: this formatting seems inconsistent

aravindhp (commented):
LGTM from me. I will leave it to @ravisantoshgudimetla to approve and @sebsoto to officially LGTM.

sebsoto:
Thanks for making the changes @mansikulkarni96 i will LGTM once this is approved

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: this change is a result of adding a dependency param, if the formatting is left untouched verify job fails, hence had to update it.

pkg/bootstrapper/kubelet.go
```diff
@@ -0,0 +1,255 @@
...
+	"time"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+// kubeletService struct contains the kubelet specific service information
+type kubeletService struct {
+	// obj is a pointer to the Windows service object
+	obj *mgr.Service
```
ravisantoshgudimetla, line 14: I thought you wouldn't need this field, you could have directly embedded the mgr.Service and that would have given the necessary options?

ravisantoshgudimetla (commented):
Mostly LGTM. I just have question around why the embedding didn't work.

pkg/bootstrapper/kubelet.go
mansikulkarni96: that is right, but I saw no harm in keeping this field to access the service object as I had already implemented using it.

pkg/bootstrapper/bootstrapper.go
aravindhp: Yeah, adding the comment has caused things to be inconsistent. But it is fine to leave it as is. This is the case to add comments on the same line.

ravisantoshgudimetla (approved):
/approve

Thank you for working on this @mansikulkarni96

PR #236, opened 2020-11-18
WINC-490: Remove downloading CNI binaries

PratikMahajan:
This PR builds the CNI plugins and then transfers them to the Windows
VM instead of downloading the CNI plugins directly from github.

Also, changes the way we copy files to the Windows VM
Now, instead of taking individual files to be transferred,
we will transfer the folders to the Windows VM according to preset source
and destination similar to what we're doing in WMCO. Nested folders wont
be copied over.


internal/test/wmcb/wmcb_test.go
```diff
@@ -58,9 +58,9 @@ var (
 		Value:  "Windows",
 		Effect: v1.TaintEffectNoSchedule,
 	}
-	// filesToBeTransferred holds the list of files that needs to be transferred to the Windows VM
-	filesToBeTransferred = flag.String("filesToBeTransferred", "",
-		"Comma separated list of files to be transferred")
+	// foldersToBeTransferred holds the list of folders that needs to be transferred to the Windows VM
+	foldersToBeTransferred = flag.String("foldersToBeTransferred", "",
```
LorbusChris, line 59: `directories` or `dirs` would be better imo

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
 	return nil
 }
 
+// CopyDirectory copies the files from the directory on the local host to the remote Windows VM
+// This does not copy nested directories
+func (w *Windows) CopyDirectory(localDir string, remoteDir string) error {
```
LorbusChris, line 92: nit : maybe a better name would be `TransferDirectory` for this func, to be in sync with the flag

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+
+	sftp, err := sftp.NewClient(w.SSHClient)
+	if err != nil {
+		return fmt.Errorf("sftp initialization failed: %v", err)
+	}
+	defer sftp.Close()
+
+	log.Printf("Copying %v directory to Windows : %v", localDir, remoteDir)
+
+	// creating a local directory to store the files and directories from remote directory.
```
PratikMahajan, line 105: change comment

LorbusChris:
A lot of vocabulary here. I think we should agree on one of:
- `Directories`, `Dirs` or `Folders`
- `copy`, `transfer` or `move`

Dockerfile.tools
```diff
@@ -82,6 +92,6 @@ COPY --from=build /build/wmcb_unit_test.exe .
 COPY --from=build /build/wmcb_e2e_test.exe .
 COPY --from=build /build/internal/test/wmcb/test-wmcb .
 
-ENV PATH="${PATH}:/home/test"
+ENV PATH="${PATH}:/payload"
```
sebsoto, line 95: Why does this need to be added to the PATH?

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
+	}
+
+	sftp, err := sftp.NewClient(w.SSHClient)
+	if err != nil {
+		return fmt.Errorf("sftp initialization failed: %v", err)
+	}
+	defer sftp.Close()
+
+	log.Printf("Copying %v directory to Windows : %v", localDir, remoteDir)
```
sebsoto, line 103: nit: %s instead of %v

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
 	return nil
 }
 
+// CopyDirectory copies the files from the directory on the local host to the remote Windows VM
+// This does not copy nested directories
+func (w *Windows) CopyDirectory(localDir string, remoteDir string) error {
+	if w.SSHClient == nil {
+		return fmt.Errorf("cannot retrieve remote directory without a ssh client")
```
sebsoto, line 94: Comment needs to be changed

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+	if err != nil {
+		return fmt.Errorf("sftp initialization failed: %v", err)
+	}
+	defer sftp.Close()
+
+	log.Printf("Copying %v directory to Windows : %v", localDir, remoteDir)
+
+	// creating a local directory to store the files and directories from remote directory.
+	err = sftp.MkdirAll(remoteDir)
+	if err != nil {
```
sebsoto, line 107: can make this into 1 line

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+	// Get the list of all files in the directory
+	localFiles, err := ioutil.ReadDir(localDir)
+	if err != nil {
+		return fmt.Errorf("error opening local directory %s: %v", localDir, err)
+	}
+
+	for _, localFile := range localFiles {
+		localPath := filepath.Join(localDir, localFile.Name())
+		// transfer only files and not directories
+		if !localFile.IsDir() {
```
sebsoto, line 120: nit: it could be preferred to do
```
if localFIle.IsDir() {
  continue
}
```
in order to reduce indentation

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+			if err = w.CopyFile(localPath, remoteDir); err != nil {
+				log.Printf("error while copying %s file to Windows : %v", localPath, err)
```
sebsoto, line 122: This should return an error, as that file could be expected later

internal/test/wmcb/wmcb_test.go
```diff
@@ -256,17 +238,10 @@ func (vm *wmcbVM) initializeTestConfigureCNIFiles(ovnHostSubnet string) error {
...
 
-	cniPkgUrl := framework.pkgs[cniPluginPkgName].getUrl()
-	cniUrl, err := url.Parse(cniPkgUrl)
-	if err != nil {
-		return fmt.Errorf("error parsing %s: %v", cniPkgUrl, err)
-	}
-
-	// Download and extract the CNI binaries on the Windows VM
-	err = vm.remoteDownloadExtract(framework.pkgs[cniPluginPkgName], remoteDir+path.Base(cniUrl.Path), winCNIDir)
+	out, err := vm.Run(fmt.Sprintf("Move-Item -Path %s -Destination %s", remoteDir+path.Base("cni"), winCNIDir), true)
```
sebsoto, line 241: Why do this instead of just copying the files to where we want in the first place?

sebsoto (changes requested):

internal/test/wmcb/wmcb_test.go
```diff
@@ -112,10 +90,14 @@ func TestWMCB(t *testing.T) {
 	for _, vm := range framework.WinVMs {
 		log.Printf("Testing VM: %s", vm.GetCredentials().InstanceId())
 		wVM := &wmcbVM{vm}
-		files := strings.Split(*filesToBeTransferred, ",")
-		for _, file := range files {
-			err := wVM.CopyFile(file, remoteDir)
-			require.NoError(t, err, "error copying %s to the Windows VM", file)
+		folders := strings.Split(*foldersToBeTransferred, ",")
```
VaishnaviHire, line 93: `folder` and `directory` are used interchangeably here. Should we just go with `directory` / `dir`?

VaishnaviHire (commented):
Thanks @PratikMahajan for working on this. PTAL at my comment

internal/test/windows/windows.go
aravindhp: But this would put it in out of sync with `CopyFile`. I suggest leaving it as `CopyDirectory`.

internal/test/windows/windows.go
```diff
@@ -81,6 +87,45 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
 	return nil
 }
 
+// CopyDirectory copies the files from the directory on the local host to the remote Windows VM
+// This does not copy nested directories
```
aravindhp, line 91: Not needed here if you are describing it in the interface.

internal/test/windows/windows.go
aravindhp: `Windows VM:`

aravindhp (changes requested):

internal/test/wmcb/wmcb_test.go
PratikMahajan: We dont know the destination while transferring the files to Windows VM. So Transferring it to a temp directory and then copying it to required directory.

internal/test/wmcb/wmcb_test.go
```diff
@@ -58,9 +55,9 @@ var (
 		Value:  "Windows",
 		Effect: v1.TaintEffectNoSchedule,
 	}
-	// filesToBeTransferred holds the list of files that needs to be transferred to the Windows VM
-	filesToBeTransferred = flag.String("filesToBeTransferred", "",
-		"Comma separated list of files to be transferred")
+	// dirsToBeTransferred holds the list of folders that needs to be transferred to the Windows VM
```
PratikMahajan, line 58: directories

internal/test/wmcb/wmcb_test.go
```diff
@@ -112,10 +90,14 @@ func TestWMCB(t *testing.T) {
 	for _, vm := range framework.WinVMs {
 		log.Printf("Testing VM: %s", vm.GetCredentials().InstanceId())
 		wVM := &wmcbVM{vm}
-		files := strings.Split(*filesToBeTransferred, ",")
-		for _, file := range files {
-			err := wVM.CopyFile(file, remoteDir)
-			require.NoError(t, err, "error copying %s to the Windows VM", file)
+		folders := strings.Split(*dirsToBeTransferred, ",")
```
PratikMahajan, line 93: directories

internal/test/windows/windows.go
```diff
@@ -55,6 +59,8 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
 	}
 	defer ftp.Close()
 
+	log.Printf("Copying %s file to Windows : %v", filePath, remoteDir)
```
mansikulkarni96, line 62: Windows vm*

internal/test/wmcb/wmcb_test.go
```diff
@@ -256,17 +238,10 @@ func (vm *wmcbVM) initializeTestConfigureCNIFiles(ovnHostSubnet string) error {
...
 	}
 
-	cniPkgUrl := framework.pkgs[cniPluginPkgName].getUrl()
-	cniUrl, err := url.Parse(cniPkgUrl)
-	if err != nil {
-		return fmt.Errorf("error parsing %s: %v", cniPkgUrl, err)
-	}
-
-	// Download and extract the CNI binaries on the Windows VM
-	err = vm.remoteDownloadExtract(framework.pkgs[cniPluginPkgName], remoteDir+path.Base(cniUrl.Path), winCNIDir)
```
mansikulkarni96, line 266: can we remove removeDownload and removeDownloadExtract methods, they don't seem to be used anywhere?

internal/test/wmcb/pkginfo.go
```diff
@@ -52,31 +52,11 @@ type pkgInfo struct {
 // pkgInfoFactory returns PkgInfo specific to the package name
 func pkgInfoFactory(name pkgName, shaType string, baseUrl string, version string) (PkgInfo, error) {
 	switch name {
-	case cniPluginPkgName:
```
mansikulkarni96, line 55: is this file pkginfo required anymore?

internal/test/windows/windows.go
mansikulkarni96, line 62: Windows VM*

mansikulkarni96 (commented):
Thanks for working on this @PratikMahajan PTAL ta my comments.

Dockerfile.tools
```diff
@@ -72,14 +72,15 @@ ENV GOCACHE="/go/.cache"
...
-COPY internal/test/wmcb/templates templates
+WORKDIR /payload
 COPY internal/test/wmcb/powershell/ .
-
 COPY --from=build /build/ovn-kubernetes/go-controller/_output/go/bin/windows/hybrid-overlay-node.exe .
 COPY --from=build /build/kubernetes/_output/local/bin/windows/amd64/kubelet.exe .
 COPY --from=build /build/wmcb_unit_test.exe .
 COPY --from=build /build/wmcb_e2e_test.exe .
+
+WORKDIR /home/test
```
aravindhp, line 92: Why do we need a separate directory for the template? We should definitely move away from `/home/test/` either way.

internal/test/windows/windows.go
```diff
@@ -29,6 +30,9 @@ type Windows struct {
 
 // WindowsVM is the interface for interacting with a Windows object created by the cloud provider
 type WindowsVM interface {
+	// CopyDirectory copies the files from the directory on the local host to the remote Windows VM
```
aravindhp, line 33: `...to the directory on the remote Windows VM ...`

internal/test/windows/windows.go
```diff
@@ -81,6 +87,43 @@ func (w *Windows) CopyFile(filePath, remoteDir string) error {
...
+	localFiles, err := ioutil.ReadDir(localDir)
+	if err != nil {
+		return fmt.Errorf("error opening local directory %s: %v", localDir, err)
+	}
+
+	for _, localFile := range localFiles {
+		localPath := filepath.Join(localDir, localFile.Name())
+		// transfer only files and not directories
+		if localFile.IsDir() {
+			continue
```
aravindhp, line 118: Add a log saying we are skipping this directory,

internal/test/wmcb/wmcb_test.go
aravindhp: Not sure what you mean by saying `directories`.

Please replace `folders` with `directories`.

internal/test/wmcb/wmcb_test.go
```diff
@@ -112,10 +112,14 @@ func TestWMCB(t *testing.T) {
 	for _, vm := range framework.WinVMs {
 		log.Printf("Testing VM: %s", vm.GetCredentials().InstanceId())
 		wVM := &wmcbVM{vm}
-		files := strings.Split(*filesToBeTransferred, ",")
-		for _, file := range files {
-			err := wVM.CopyFile(file, remoteDir)
-			require.NoError(t, err, "error copying %s to the Windows VM", file)
+		folders := strings.Split(*dirsToBeTransferred, ",")
+		// all the files in `/payload` directory will be transferred to `C:\Temp`
+		// for files not in `/payload`, new directory will be created in `C:\Temp` which resembles the linux folder path
```
aravindhp, line 95: I don't understand this comment. For one, this comment will need to be updated every time the input list of directories changes. So you need to say something to the effect of `not in the list of directories`. Moreover what do you mean by `for files not in /payload`?

>		// for files not in `/payload`, new directory will be created in `C:\Temp` which resembles the linux folder path

Where are we doing this? I think confusion is because you are using values instead of the referencing the variables here.

internal/test/wmcb/wmcb_test.go
```diff
@@ -112,10 +112,14 @@ func TestWMCB(t *testing.T) {
...
+		for _, folder := range folders {
```
aravindhp, line 96: Please use `directories` or `dirs` instead of `folder` everywhere including comments. In comments please use `directories`.

Dockerfile.tools
```diff
@@ -36,8 +36,12 @@ RUN git clone --branch release-4.6 --single-branch https://github.com/openshift/
 WORKDIR /build/ovn-kubernetes/go-controller/
 RUN make windows
 
-# TODO: Add building the CNI plugins to dockerfile and remove downloading binary
-# https://issues.redhat.com/browse/WINC-490
+# Build CNI plugins
+WORKDIR /build/
+RUN git clone --branch release-4.6 --single-branch https://github.com/openshift/containernetworking-plugins
```
aravindhp, line 41: I think we should clone WMCO and use the sub-module there. We need to do this for other components too but that can be a separate PR.

@LorbusChris maybe you can tackle this as part of updating sub-modules in WMCO.

internal/test/wmcb/pkginfo.go
aravindhp: If we are removing the last user of `pkgInfo`, then `pkginfo.go` should be deleted using `git rm`.

aravindhp (changes requested):
Thanks for working on this, @PratikMahajan. Please address my comments.

internal/test/wmcb/wmcb_test.go
PratikMahajan: Updated the comment to make it better to understand. LMK if its still confusing

internal/test/wmcb/wmcb_test.go
```diff
@@ -112,10 +112,17 @@ func TestWMCB(t *testing.T) {
...
+		// all the files in `/payload` directory will be transferred to `remoteDir`
+		// for files not in `/payload`, new directory will be created in `remoteDir` which resembles the linux folder path
+		// for eg. `/payload/cni` will be stored in `remoteDir\cni` and `/foo/bar` will be stored in `remoteDir\foo\bar`
+		for _, dir := range dirs {
+			// remove `/payload` from the path and split the path
+			dirPath := strings.Split(strings.Replace(dir, "/payload", "", -1), "/")
+			// join the path with `\\` and place it in `remoteDir`
+			remotePath := remoteDir + "\\" + strings.Join(dirPath, "\\")
+			err := wVM.CopyDirectory(dir, remotePath)
+			require.NoError(t, err, "error copying %s to the Windows VM", dir)
```
aravindhp, line 99: I am a nack on this approach. You are assuming that the input will be `/payload`. Sure we are building the image that way but you cannot leak such assumptions into the test code.

aravindhp (changes requested):

LorbusChris (approved):

aravindhp (approved):
Thanks for working on this, @PratikMahajan 

/approve

PR #239, opened 2020-11-30
Bug 1894649: Update Kubelet service in-place instead of recreating

VaishnaviHire:
This commit ensures that the idempotency of kubelet service is
mainitained. Recreating the kubelet service caused the bug where
wmcb tried to access kubelet service that was marked for deletion.
The changes include removing deletion of kubelet service, and
ensuring that the kubelet service is created if it is not present,
else updating the existing one with custom specifications.

[Bug 1894649](https://issues.redhat.com/browse/OCPBUGSM-20185)


pkg/bootstrapper/bootstrapper.go
```diff
@@ -830,3 +828,58 @@ func (cni *cniOptions) configure(kubeletCmd *string) error {
...
+	// Index 0 of kubeletArgs will hold the kubelet.exe.
+	kubeletPath := strings.Split(*kubeletCmd, " ")[0]
+	if !strings.Contains(kubeletPath, "kubelet.exe") {
+		return fmt.Errorf("kubelet command does not start with kubelet.exe")
+	}
+	kubeletKeyValueArgs[kubeletExeKey] = kubeletPath
+
+	if *kubeletCmd, err = reconstructKubeletCmd(kubeletKeyValueArgs); err != nil {
+		return fmt.Errorf("unable to reconstruct kubelet command %v: %v", kubeletKeyValueArgs, err)
+	}
```
sebsoto, line 882: The benefit of converting the args into a map is that if we want to change specific arguments that is made easier. In this case you are not doing that, and it seems like you are creating a new function to convert a []string into a map so that the `reconstructKubeletCmd` can be reused. While its good to reuse functions, this is adding unnecessary logic and obfuscating a simple operation. 

Why was the below solution we talked about unviable?
```	
        s := syscall.EscapeArg(kubeletpath)
	for _, v := range args {
		s += " " + syscall.EscapeArg(v)
	}
```
This code was borrowed from the [svc mgr library](https://github.com/golang/sys/blob/master/windows/svc/mgr/mgr.go#L126-L129)

pkg/bootstrapper/bootstrapper.go
```diff
@@ -182,9 +182,12 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 	var dependents []*mgr.Service
```
sebsoto, line 182: I think the approach taken here requires a commit/PR title change, something like `Update kubelet service in place instead of recreating`

sebsoto (changes requested):

pkg/bootstrapper/bootstrapper.go
```diff
@@ -445,9 +448,8 @@ func (wmcb *winNodeBootstrapper) initializeKubeletFiles() error {
...
 }
 
-// createKubeletService creates a new kubelet service to our specifications
-func (wmcb *winNodeBootstrapper) createKubeletService() error {
-	var err error
+// createInitialKubeletArgs defines the kubelet args required during initial kubelet start up.
+func (wmcb *winNodeBootstrapper) createInitialKubeletArgs() []string {
 	// If initialize-kubelet is run after configure-cni, the kubelet args will be overwritten and the CNI
 	// configuration will be lost. The assumption is that every time initialize-kubelet is run, configure-cni needs to
 	// be run again. This is how the WSU playbook is written and we don't expect users to execute WMCB directly.
```
aravindhp, line 455: Given you are in the area, please update the comment

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +506,74 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 	}
 
 	var dependents []*mgr.Service
-	if dependentSvc, err := wmcb.svcMgr.OpenService(kubeletDependentSvc); err == nil {
-		dependents = append(dependents, dependentSvc)
+	dependentSvc, err := wmcb.svcMgr.OpenService(kubeletDependentSvc)
+	if err != nil {
+		return fmt.Errorf("colud not get kubelet dependent services %s : %v", kubeletDependentSvc, err)
 	}
+	dependents = append(dependents, dependentSvc)
```
aravindhp, line 540: This is a repeated pattern and should be modularized.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -445,9 +448,8 @@ func (wmcb *winNodeBootstrapper) initializeKubeletFiles() error {
 	return nil
 }
 
-// createKubeletService creates a new kubelet service to our specifications
-func (wmcb *winNodeBootstrapper) createKubeletService() error {
-	var err error
+// createInitialKubeletArgs defines the kubelet args required during initial kubelet start up.
+func (wmcb *winNodeBootstrapper) createInitialKubeletArgs() []string {
```
aravindhp, line 452: This is more like `getInitialKubeletArgs()`.  You should update the comment stating that this function assumes `wmcb.kubeletArgs` have been populated. Ideally there should be a guard but I am not sure what that guard looks like.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +506,74 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
-}
+	if err := configureKubelet(&c.BinaryPathName, kubeletArgs); err != nil {
+		return fmt.Errorf("error configuring kubelet service: %v", err)
+	}
 
-// refreshServiceManager will disconnect and reconnect from the Windows service API. In order to complete certain
-// operations, there must be zero handlers to the API present on the system.
-func (wmcb *winNodeBootstrapper) refreshServiceManager() error {
-	var err error
-	if err = wmcb.Disconnect(); err != nil {
```
aravindhp, line 527: Do we need to call this after we finish configuration?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -830,3 +828,58 @@ func (cni *cniOptions) configure(kubeletCmd *string) error {
...
+
+	// Remove the trailing space
+	if standaloneArgs, found := kubeletKeyValueArgs[kubeletStandAloneArgsKey]; found {
+		kubeletKeyValueArgs[kubeletStandAloneArgsKey] = strings.TrimSpace(standaloneArgs)
+	}
+	return kubeletKeyValueArgs, nil
+
+}
+
+// configureKubelet updates the given kubelet command with the required args.
```
aravindhp, line 862: If this is updating the kubelet cmd then why is it called `configureKubelet`? This should be `updateKubeletCmd`

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Given this is a bug, I am not sure if the title should be same as in Bugzilla

test/e2e/bootstrapper_test.go
```diff
@@ -75,6 +75,13 @@ func TestBootstrapper(t *testing.T) {
 		assert.True(t, isKubeletRunning(t, kubeletLogPath))
 	})
 
+	t.Run("Update already running kubelet service", func(t *testing.T) {
+		err := wmcb.InitializeKubelet()
+		assert.NoError(t, err, "unable to update kubelet service")
+		time.Sleep(1 * time.Minute)
```
aravindhp, line 81: Any reason to use 1 minute vs 30 seconds here?

aravindhp (changes requested):
Thanks for working on this, @VaishnaviHire. Please address my comments.

test/e2e/bootstrapper_test.go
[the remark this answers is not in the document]
VaishnaviHire: At the start of the `Poll`, the kubelet service would be running, resulting in immediate return

pkg/bootstrapper/bootstrapper.go
```diff
@@ -182,9 +182,12 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 	var dependents []*mgr.Service
 	// If there is already a kubelet service running, find it
 	if ksvc, err := svcMgr.OpenService(KubeletServiceName); err == nil {
-		if dependentSvc, err := svcMgr.OpenService(kubeletDependentSvc); err == nil {
-			dependents = append(dependents, dependentSvc)
+		dependentSvc, err := svcMgr.OpenService(kubeletDependentSvc)
+		if err != nil {
```
mansikulkarni96, line 186: Correct me if I am wrong but isn't this function called on intialize kubelet as well and when the hybrid-overlay is not yet installed? Won't that always result in an error being returned here?

pkg/bootstrapper/bootstrapper.go
aravindhp: I missed this. Please don't repeat the title of the bug in the PR. That is akin to saying you are introducing the bug. Mention **how** you are fixing the bug.

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: We are also calling [refresh](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/pkg/bootstrapper/kubelet.go#L116) after configuration. Do we still go ahead and call this function ?

aravindhp: I am going to defer to @sebsoto on this. I want to ensure we are closing all open handles.

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: You might be right. How do we handle errors returned by `OpenService` when the dependent services are installed ?

mansikulkarni96: we might have to check by ignoring a specific error here.

pkg/bootstrapper/bootstrapper.go
sebsoto: The refresh was important since we needed to close handles to the service manager in order to allow the service to be deleted. Now all we need to make sure to do is that we are closing it at the end. The process exiting should do that automatically, but to be safe we should explicitly disconnect

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: I agree that we can get away without converting the []string to map, and can update that.

However I am not sure why the above code snippet is required when we can pass `&c.BinaryPath` where c is the `inital config`

test/e2e/bootstrapper_test.go
VaishnaviHire: Calling InitializeKubelet() on already running service involves stopping the service, updating the config and restarting the service, hence the increased buffer time

LorbusChris:
nit: two typos in commit message:
```diff
-Bug 1894649: Idempotentcy of kubelet service is broken
+Bug 1894649: Idempotency of kubelet service is broken

-This commit ensures that the idempotentcy of kubelet service is
+This commit ensures that the idempotency of kubelet service is
-mainitained.The changes include ensuring that the kubelet service is
+maintained.The changes include ensuring that the kubelet service is
created if it is not present, else updating the existing kubelet
service with custom specifications.
```

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: We do have some guards [here](https://github.com/openshift/windows-machine-config-bootstrapper/pull/239/files#diff-e9ddf87a450b985b61b945593604139617a74d4dbacff8b018acfe18c5880c2cR483). Show we return error if the key doesn`t exist ?

pkg/bootstrapper/bootstrapper.go
aravindhp: Never assume exiting will clean things up automatically, especially on Windows ;-)

test/e2e/bootstrapper_test.go
aravindhp: Not a fan of such sleeps. I would prefer we wait for the expected result with a timeout.

pkg/bootstrapper/bootstrapper.go
aravindhp: That is not a guard. A guard would something that returns an error if the function is called before the args have been populated. 

>Should we return error if the key doesn`t exist ?

Not sure if we can do that. Some of the args are optional.

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: sorry I missed this, but this is only called when kubelet service is already running and not when we are creating it for the first time. Thinking of this scenario I think its fine checking for any error here.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +506,74 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 	wmcb.kubeletSVC, err = newKubeletService(ksvc, dependents)
 	if err != nil {
 		return fmt.Errorf("could not initialize struct kubeletService: %v", err)
 	}
+	return nil
+}
 
-	if err := wmcb.kubeletSVC.setRecoveryActions(); err != nil {
-		return fmt.Errorf("failed to set recovery actions for Windows service %s", KubeletServiceName)
+// createKubeletService creates a new kubelet service to our specifications
```
mansikulkarni96, line 548: comment needs to be updated

mansikulkarni96 (commented):
Thanks for fixing this @VaishnaviHire PTAL ta my comments.

pkg/bootstrapper/bootstrapper.go
sebsoto: Its important to escape the arguments in case we are sourcing any problematic strings from the ignition file

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Won`t the refresh function be required here since it calls, [svcobj.UpdateConfig](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/pkg/bootstrapper/kubelet.go#L121). We are also calling [Disconnect](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/cmd/bootstrapper/initialize_kubelet.go#L72) after IntializeKubelet, that should clean things up

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Discussed this, neither of the input is needed, we could generate the string and update the c.BinaryPath

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Ignoring `service does not exist` 

> we might have to check by ignoring a specific error here.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +502,99 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
+	kubeletcmd := " " + filepath.Join(wmcb.installDir, "kubelet.exe") + " "
+	var standaloneArgs string
+	var keyvalueArgs string
+	// Add rest of the args
+	for _, args := range kubeletArgs {
+		if strings.Contains(args, "=") {
+			keyvalueArgs += args + " "
+		} else {
+			standaloneArgs += args + " "
+		}
```
sebsoto, line 566: Why is this separation being done?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +502,99 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
-	// We need to give Windows time to clean up the services we've marked for deletion
-	time.Sleep(serviceWaitTime)
-	wmcb.svcMgr, err = mgr.Connect()
-	return err
+	existingConfig.BinaryPathName = strings.TrimSpace(kubeletcmd + standaloneArgs + keyvalueArgs)
+
+	// Update service config and restart
+	if err := wmcb.kubeletSVC.refresh(existingConfig); err != nil {
+		return fmt.Errorf("unable to refresh kubelet service: %v", err)
+	}
```
sebsoto, line 573: This shouldnt be necessary

pkg/bootstrapper/bootstrapper.go
```diff
@@ -830,3 +869,22 @@ func (cni *cniOptions) configure(kubeletCmd *string) error {
 
 	return nil
 }
+
+// updateKubeletDependents updates the dependents field of the kubeletService struct
+// to reflect current list of dependent services. This function assumes that the kubelet service is running
+func updateKubeletDependents(svcMgr *mgr.Mgr) ([]*mgr.Service, error) {
+
```
sebsoto, line 876: extra space

sebsoto (changes requested):

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Wouldn`t we need to call [updateConfig](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/pkg/bootstrapper/kubelet.go#L121) ?

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Thanks for catching that. I need to revert it. Added it as part of a debug.

pkg/bootstrapper/bootstrapper.go
sebsoto: sorry you are right, read it too quick and thought that was refreshing the manager

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +502,99 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 
-// refreshServiceManager will disconnect and reconnect from the Windows service API. In order to complete certain
-// operations, there must be zero handlers to the API present on the system.
-func (wmcb *winNodeBootstrapper) refreshServiceManager() error {
-	var err error
-	if err = wmcb.Disconnect(); err != nil {
-		return err
+	// Create kubelet command to populate config.BinaryPathName
+	// Add a space after kubelet.exe followed by the stand alone args
+	kubeletcmd := " " + filepath.Join(wmcb.installDir, "kubelet.exe") + " "
```
aravindhp, line 557: Why do you need a space before the command?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -499,63 +502,99 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
 	}
-	err = wmcb.initializeKubeletFiles()
-	if err != nil {
-		return fmt.Errorf("failed to initialize kubelet: %v", err)
-	}
-	err = wmcb.createKubeletService()
+	err = wmcb.ensureKubeletService()
 	if err != nil {
-		return fmt.Errorf("failed to create kubelet windows service: %v", err)
+		return fmt.Errorf("failed to ensure that kubelet windows service is created: %v", err)
```
aravindhp, line 597: s/created/present

pkg/bootstrapper/kubelet.go
```diff
@@ -125,18 +126,21 @@ func (k *kubeletService) refresh(config mgr.Config) error {
 	if err := k.start(); err != nil {
 		return fmt.Errorf("error starting kubelet service: %v", err)
 	}
+	// Wait for service to go to Running state
+	err := wait.Poll(30*time.Second, 2*time.Minute, func() (done bool, err error) {
```
aravindhp, line 130: Such times should be expressed as constants.

aravindhp (changes requested):

pkg/bootstrapper/bootstrapper.go
```diff
@@ -481,15 +484,18 @@ func (wmcb *winNodeBootstrapper) createKubeletService() error {
...
+// it updates the existing kubelet service with our specifications.
+func (wmcb *winNodeBootstrapper) ensureKubeletService() error {
 	// Mostly default values here
 	c := mgr.Config{
 		ServiceType: 0,
 		// StartAutomatic will start the service again if the node restarts
-		StartType:    mgr.StartAutomatic,
-		ErrorControl: 0,
-		// Path to kubelet.exe
-		BinaryPathName: filepath.Join(wmcb.installDir, "kubelet.exe"),
```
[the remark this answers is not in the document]
VaishnaviHire: Yes, we pass it as exepath [here](https://github.com/openshift/windows-machine-config-bootstrapper/blob/7720182cdc978c38df06eed22cf36abbf7c9ee0a/vendor/golang.org/x/sys/windows/svc/mgr/mgr.go#L116)

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: Yes. [ConfigureCNICmd](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/cmd/bootstrapper/configure_cni.go#L72)

pkg/bootstrapper/kubelet.go
```diff
@@ -6,6 +6,14 @@ import (
 
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// waitTimeKubeletService is the time required for kubelet service to go to running state
+	waitTimeKubeletService = 30 * time.Second
```
aravindhp, line 14: This is not the wait time but rather the interval at which you will poll the service. `svcPollInterval`. Given it is is in `kubelet.go` no reason to have `kubelet`.

pkg/bootstrapper/kubelet.go
```diff
@@ -6,6 +6,14 @@ import (
...
+	// TimeoutKubeletService is the maximum duration to wait for the kubelet service to go to running state
+	timeoutKubeletService = 2 * time.Minute
```
aravindhp, line 16: Similarly I suggest using `svcRunTimeout`. Please set comment correctly too.

test/e2e/bootstrapper_test.go
```diff
@@ -71,10 +74,22 @@ func TestBootstrapper(t *testing.T) {
...
 		assert.True(t, isKubeletRunning(t, kubeletLogPath))
 	})
 
+	t.Run("Update already running kubelet service", func(t *testing.T) {
+		err := wmcb.InitializeKubelet()
+		assert.NoError(t, err, "unable to update kubelet service")
+		err = wmcb.Disconnect()
+		assert.NoErrorf(t, err, "Could not disconnect from windows svc API: %s", err)
+
+		err = wait.Poll(waitTimeKubeletLog, 2*time.Minute, func() (done bool, err error) {
```
aravindhp, line 87: Please add a constant for `2*time.Minute`. `waitTimeKubeletLog`.

test/e2e/bootstrapper_test.go
```diff
@@ -20,7 +21,11 @@ var ignitionFilePath string
 var kubeletPath string
 var installDir string
 
-const kubeletLogPath = "C:\\var\\log\\kubelet\\kubelet.log"
+const (
+	kubeletLogPath = "C:\\var\\log\\kubelet\\kubelet.log"
+	// waitTimeKubeletLog is the time required to populate kubelet log
+	waitTimeKubeletLog = 30 * time.Second
```
aravindhp, line 27: `pollIntervalKubeletLog`

aravindhp (approved):
/approve

with the caveat that the variable names are fixed up.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -830,3 +863,21 @@ func (cni *cniOptions) configure(kubeletCmd *string) error {
...
+// updateKubeletDependents updates the dependents field of the kubeletService struct
+// to reflect current list of dependent services. This function assumes that the kubelet service is running
+func updateKubeletDependents(svcMgr *mgr.Mgr) ([]*mgr.Service, error) {
+	var dependents []*mgr.Service
+	// If there is already a kubelet service running, find it
+	dependentSvc, err := svcMgr.OpenService(kubeletDependentSvc)
+	if err != nil {
+		// Do not return error if the services are not installed.
+		if !strings.Contains(err.Error(), "service does not exist") {
+			return nil, fmt.Errorf("error get dependent services for kubelet %v", err)
```
LorbusChris, line 876: tiny nit:
```suggestion
			return nil, fmt.Errorf("error getting dependent services for kubelet %v", err)
```

LorbusChris (approved):
/lgtm

LorbusChris:
@VaishnaviHire the nit above, plus the ones mentioned in https://github.com/openshift/windows-machine-config-bootstrapper/pull/239#issuecomment-737392572 - not blocking on it, though :)

PR #242, opened 2020-12-07
Dockerfile.tools: Clone WMCO repo and build submodules from there

LorbusChris:
This ensures the testing in this repository uses the same commits of the
submodule components as the operator does.

### DIFF
--- a/Dockerfile.tools
+++ b/Dockerfile.tools
@@ -1,24 +1,22 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 as build
 LABEL stage=build
+
+# Clone WMCO repo and initialize submodules for building of windows binaries payload for e2e testing 
 WORKDIR /build/
+RUN git clone --branch release-4.7 --single-branch https://github.com/openshift/windows-machine-config-operator.git
+WORKDIR /build/windows-machine-config-operator/
+RUN git submodule update --init containernetworking-plugins kubernetes ovn-kubernetes
 
 # Build kubelet.exe
-RUN git clone --branch release-4.6 https://github.com/openshift/kubernetes.git
-WORKDIR /build/kubernetes/
-# TODO: Checking out commit before go1.15 became a requirement, revert checkout in https://issues.redhat.com/browse/WINC-460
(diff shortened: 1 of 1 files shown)


aravindhp (approved):

/approve

Dockerfile.tools
```diff
@@ -2,23 +2,17 @@ FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6
 LABEL stage=build
 WORKDIR /build/
 
-# Build kubelet.exe
-RUN git clone --branch release-4.6 https://github.com/openshift/kubernetes.git
-WORKDIR /build/kubernetes/
-# TODO: Checking out commit before go1.15 became a requirement, revert checkout in https://issues.redhat.com/browse/WINC-460
-RUN git checkout f5121a6a6a02ddfafd2bfbf5201b092dc5097ab0
+RUN git clone --branch release-4.7 --single-branch --recurse-submodules https://github.com/openshift/windows-machine-config-operator.git
```
aravindhp, line 5: Should we add an "inception" warning here given the WMCB repo is a sub-module? :-P

mansikulkarni96:
Thanks for working on this @LorbusChris 
/lgtm

Dockerfile.tools
LorbusChris: Hm, I'm starting to think I we should probably only initialize the submodule repos that we actually need/build, otherwise we're cloning all of them each time.

LorbusChris:
/hold
per https://github.com/openshift/windows-machine-config-bootstrapper/pull/242#discussion_r539440181

Dockerfile.tools
aravindhp: I like that approach

Dockerfile.tools
```diff
@@ -1,24 +1,22 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 as build
```
LorbusChris, line 1: @aravindhp should I bump this to 4.7?
```suggestion
FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 as build
```

aravindhp: Yes please given this is going into master.

PR #243, opened 2020-12-07
[release-4.6] Dockerfile.tools: Clone WMCO repo and build submodules from there

LorbusChris:
This ensures the testing in this repository uses the same commits of the
submodule components as the operator does.


aravindhp (approved):

/approve

PR #244, opened 2020-12-08
Bug 1894649: Fix Node NotReady state when updating Kubelet service.

VaishnaviHire:
This commmit ensures that we fix the issue created by PR [#239](https://github.com/openshift/windows-machine-config-bootstrapper/pull/239)
where a Windows Node goes to a `NotReady` state, when kubelet
is updated. The issue was caused by cloud-provider kubelet arg
getting overridden by an empty string. `initializeKubeletFiles` function
was not being called during update kubelet, since all files are already
present in the VM, however the function was also responsible for setting
cloud-provider arg. To resolve this issue, this commit makes changes
to call the  `initializeKubeletFiles` function during update. It is
also required to stop an already running kubelet service before
`initializeKubeletFiles` is called to avoid file access errors. The
commit also adds tests to check for `Ready` state for Windows Node.

### DIFF
--- a/internal/test/wmcb/wmcb_test.go
+++ b/internal/test/wmcb/wmcb_test.go
@@ -412,4 +412,16 @@ func testWMCBCluster(t *testing.T) {
 		metav1.ListOptions{LabelSelector: e2ef.WindowsLabel})
 	require.NoErrorf(t, err, "error while getting Windows node: %v", err)
 	assert.Lenf(t, winNodes.Items, 1, "expected one node to have node label but found: %v", len(winNodes.Items))
+	// Test Windows Nodes for Ready status
+	for _, node := range winNodes.Items {
+		readyCondition := false
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == v1.NodeReady {
+				readyCondition = true
+				assert.Equalf(t, v1.ConditionTrue, condition.Status, "expected Windows node %v should be in %v State",
+					node.Name, condition.Status)
+			}
+		}
+		assert.Truef(t, readyCondition, "expected node Status to have condition type Ready for node %v", node.Name)
+	}
 }
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -580,12 +580,21 @@ func (wmcb *winNodeBootstrapper) updateKubeletService(config mgr.Config, kubelet
 // service, and then starts the kubelet service
 func (wmcb *winNodeBootstrapper) InitializeKubelet() error {
 	var err error
-	if wmcb.kubeletSVC == nil {
-		err = wmcb.initializeKubeletFiles()
+
+	if wmcb.kubeletSVC != nil {
+		// Stop kubelet service if it is in Running state. This is required to access kubelet files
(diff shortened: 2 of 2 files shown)


internal/test/wmcb/wmcb_test.go
```diff
@@ -412,4 +412,14 @@ func testWMCBCluster(t *testing.T) {
 		metav1.ListOptions{LabelSelector: e2ef.WindowsLabel})
 	require.NoErrorf(t, err, "error while getting Windows node: %v", err)
 	assert.Lenf(t, winNodes.Items, 1, "expected one node to have node label but found: %v", len(winNodes.Items))
+	// Test Windows Nodes for Ready status
+	for _, node := range winNodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == v1.NodeReady {
```
aravindhp, line 418: Won't `condition.Type` not being present result in a false positive?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -543,10 +543,6 @@ func (wmcb *winNodeBootstrapper) updateKubeletService(config mgr.Config, kubelet
 		return fmt.Errorf("no existing config found")
 	}
 
-	// Stop the kubelet service as there could be open file handles from kubelet.exe on the plugin files
-	if err := wmcb.kubeletSVC.stop(); err != nil {
-		return fmt.Errorf("unable to stop kubelet service: %v", err)
-	}
```
aravindhp, line 549: The assumption now is that you need to call `updateKubeletService()` only after you have stopped the service. Don't you need a guard for this to be on the safe side?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -580,12 +576,21 @@ func (wmcb *winNodeBootstrapper) updateKubeletService(config mgr.Config, kubelet
...
-	if wmcb.kubeletSVC == nil {
-		err = wmcb.initializeKubeletFiles()
+
+	if wmcb.kubeletSVC != nil {
+		// Stop kubelet service if it is in Running state. This is required to access kubelet files
+		// without getting 'The process cannot access the file because it is being used by another process.' error
+		err := wmcb.kubeletSVC.stop()
 		if err != nil {
-			return fmt.Errorf("failed to initialize kubelet: %v", err)
+			return fmt.Errorf("failed to stop existing kubelet service: %v", err)
```
aravindhp, line 585: Please remove `existing` as it is redundant.

aravindhp (changes requested):
Thanks for working on this, @VaishnaviHire. Please address my comments.

pkg/bootstrapper/bootstrapper.go
VaishnaviHire: I will revert this change

internal/test/wmcb/wmcb_test.go
sebsoto: Agreed, the NodeReady label being missing should cause a test failure

sebsoto (approved):
/lgtm

mansikulkarni96:
LGTM

aravindhp (approved):
/approve

PR #245, opened 2020-12-23
WINC-479: [wni] Remove wni from bootstrapper

VaishnaviHire:
This commit ensures that all unused wni code is
deleted and is not used by any other component.

PR #249, opened 2021-01-05
WINC-479: [release-4.6] Remove WNI from WMCB

VaishnaviHire:
Backport #245


aravindhp (changes requested):
Please open separate PRs. We should ideally should have 1-1 mapping of PRs between branches.
Moreover `WINC-479/480:` will break Jira linking.

aravindhp (approved):
/approve

PR #251, opened 2021-01-06
 [ci] Define build_root_image within WMCB repo

VaishnaviHire:
This commit allows us to change the build_root_image with commits to
this repo. This is to maintain consistency with the WMCO repo and to
not require explict changes to openshift/release during go version update.


aravindhp (approved):
/approve

PR #253, opened 2021-01-11
Bug 1894649: [community-4.6] Fix Node NotReady state when updating Kubelet service.

VaishnaviHire:
This commmit ensures that we fix the issue created by PR #239
where a Windows Node goes to a NotReady state, when kubelet
is updated. The issue was caused by cloud-provider kubelet arg
getting overrided by empty string. initializeKubeletFiles function
was not being called during update kubelet, since all files are already
present in the VM, however the function was also responsible for setting
cloud-provider arg. To resolve this issue, this commit makes changes
to call the  initializeKubeletFiles function during update. It is
also required to stop an already running kubelet service before
initializeKubeletFiles is called to avoid file access errors. The
commit also adds tests to check for Ready state for Windows Node.

(cherry picked from commit a22ec439c9a447b26f414b7ee11a5ec1b61c4920)


aravindhp (approved):
/approve

PR #258, opened 2021-02-18
WINC-555: [wmcb] Support Windows Server 20H2

aravindhp:
This requires upgrading the pause image to 3.4.1 which has support for Windows Server 20H2.


mansikulkarni96:
Thanks for working on this @aravindhp 
/lgtm

PR #259, opened 2021-02-18
[wmcb] Add uninstall kubelet command line option

mansikulkarni96:
This PR  adds a command line option "uninstall-kubelet " to uninstall the existing kubelet service from the Windows node. This will help in BYOH implementation to uninstall all services.

### DIFF
--- a/cmd/bootstrapper/uninstall_kubelet.go
+++ b/cmd/bootstrapper/uninstall_kubelet.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"flag"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/windows-machine-config-bootstrapper/pkg/bootstrapper"
+)
+
+var (
+	// uninstallKubeletCmd describes the uninstall-kubelet command
+	uninstallKubeletCmd = &cobra.Command{
+		Use:   "uninstall-kubelet",
+		Short: "Stops and removes the kubelet service",
+		Long:  "Stops and removes the kubelet service",
+		Run:   runUninstallKubeletCmd,
+	}
+)
+
+func init() {
+	rootCmd.AddCommand(uninstallKubeletCmd)
+}
+
+// runUninstallKubeletCmd uninstalls kubelet service from the Windows node
+func runUninstallKubeletCmd(cmd *cobra.Command, args []string) {
+	flag.Parse()
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper("", "", "", "", "")
+	if err != nil {
+		log.Error(err, "could not create bootstrapper")
+		os.Exit(1)
+	}
+
+	// uninstall kubelet Windows service
+	if err = wmcb.UninstallKubelet(); err != nil {
+		log.Error(err, "could not uninstall kubelet")
+		os.Exit(1)
+	}
+
+	// Send success message to StdOut to ascertain that kubelet removal was successful
+	os.Stdout.WriteString("kubelet uninstalled successfully")
+
+	if err = wmcb.Disconnect(); err != nil {
+		log.Error(err, "can't clean up bootstrapper")
+	}
+}
--- a/internal/test/wmcb/wmcb_test.go
+++ b/internal/test/wmcb/wmcb_test.go
@@ -109,6 +109,9 @@ func (vm *wmcbVM) runE2ETestSuite(t *testing.T) {
 	vm.runTestBootstrapper(t)
 
 	vm.runTestConfigureCNI(t)
+
+	// Run this test only after TestBoostrapper() to ensure kubelet service is present.
+	vm.runTestKubeletUninstall(t)
 }
 
 // runTest runs the testCmd in the given VM
@@ -285,6 +288,11 @@ func (vm *wmcbVM) waitForHybridOverlayToRun() error {
 
 	// hybrid-overlay-node never started running
 	return fmt.Errorf("timeout waiting for hybrid-overlay-node: %v", err)
+}
+
+func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
+	err := vm.runTest(e2eExecutable + " --test.run TestKubeletUninstall --test.v")
+	require.NoError(t, err, "TestKubeletUninstall failed")
 }
 
 // mkdirCmd returns the Windows command to create a directory if it does not exists
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -179,18 +179,34 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 		}
 	}
 
-	// If there is already a kubelet service running, find it
-	if ksvc, err := svcMgr.OpenService(KubeletServiceName); err == nil {
-		dependents, err := updateKubeletDependents(svcMgr)
-		if err != nil {
-			return nil, fmt.Errorf("error updating kubelet dependents field %v", err)
-		}
-		bootstrapper.kubeletSVC, err = newKubeletService(ksvc, dependents)
-		if err != nil {
-			return nil, fmt.Errorf("could not initialize struct kubeletService: %v", err)
-		}
+	// If there is already a kubelet service running, find and assign it
+	bootstrapper.kubeletSVC, err = assignExistingKubelet(svcMgr)
+	if err != nil {
+		return nil, fmt.Errorf("could not assign existing kubelet service: %v", err)
 	}
 	return &bootstrapper, nil
+}
+
+// assignExistingKubelet finds the existing kubelet service from the Windows Service Manager,
+// assigns its value to the kubeletService struct and returns it.
+func assignExistingKubelet(svcMgr *mgr.Mgr) (*kubeletService, error) {
+	ksvc, err := svcMgr.OpenService(KubeletServiceName)
+	if err != nil {
+		// Do not return error if the service is not installed.
+		if !strings.Contains(err.Error(), "service does not exist") {
+			return nil, fmt.Errorf("error getting existing kubelet service %v", err)
+		}
+		return nil, nil
+	}
+	dependents, err := updateKubeletDependents(svcMgr)
+	if err != nil {
+		return nil, fmt.Errorf("error updating kubelet dependents field %v", err)
+	}
+	kubeletSVC, err := newKubeletService(ksvc, dependents)
+	if err != nil {
+		return nil, fmt.Errorf("could not initialize struct kubeletService: %v", err)
+	}
+	return kubeletSVC, nil
 }
 
 // newCNIOptions takes the paths to the kubelet installation and the CNI files as input and returns the cniOptions
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -149,7 +149,7 @@ type cniOptions struct {
 
 // NewWinNodeBootstrapper takes the dir to install the kubelet to, and paths to the ignition and kubelet files along
 // with the CNI options as inputs, and generates the winNodeBootstrapper object. The CNI options are populated only in
-// the configure-cni command.
+// the configure-cni command. The inputs to NewWinNodeBootstrapper are ignored while using the uninstall kubelet functionality.
 func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cniDir string,
 	cniConfig string) (*winNodeBootstrapper, error) {
 	// Check if cniDir or cniConfig is empty when the other is not
@@ -665,6 +665,19 @@ func (wmcb *winNodeBootstrapper) Disconnect() error {
 	err := wmcb.svcMgr.Disconnect()
 	wmcb.svcMgr = nil
 	return err
+}
+
+// UninstallKubelet uninstalls the kubelet service from Windows node
+func (wmcb *winNodeBootstrapper) UninstallKubelet() error {
+	if wmcb.kubeletSVC == nil {
+		return fmt.Errorf("kubelet service is not present")
+	}
+	// Stop and remove kubelet service if it is in Running state.
+	err := wmcb.kubeletSVC.stopAndRemove()
+	if err != nil {
+		return fmt.Errorf("failed to stop and remove kubelet service: %v", err)
+	}
+	return nil
 }
 
 func copyFile(src, dest string) error {
--- a/pkg/bootstrapper/kubelet.go
+++ b/pkg/bootstrapper/kubelet.go
@@ -183,6 +183,17 @@ func (k *kubeletService) setRecoveryActions() error {
 	return nil
 }
 
+// stopAndRemove stops and removes the kubelet service
+func (k *kubeletService) stopAndRemove() error {
+	if k.obj == nil {
+		return nil
+	}
+	if err := k.stop(); err != nil {
+		return fmt.Errorf("failed to stop kubelet service: %v", err)
+	}
+	return k.obj.Delete()
+}
+
 // startService is a helper to start a given service
 func startService(serviceObj *mgr.Service) error {
 	if serviceObj == nil {
--- a/test/e2e/uninstall_kubelet_test.go
+++ b/test/e2e/uninstall_kubelet_test.go
@@ -1,0 +1,38 @@
+package e2e
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openshift/windows-machine-config-bootstrapper/pkg/bootstrapper"
+)
+
+// TestKubeletUninstall tests if WMCB returns an error if the kubelet is uninstalled
+func TestKubeletUninstall(t *testing.T) {
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper("", "", "", "", "")
+	require.NoError(t, err, "could not create wmcb")
+
+	err = wmcb.UninstallKubelet()
+	assert.NoError(t, err, "unable to uninstall kubelet service")
+
+	err = wmcb.Disconnect()
+	assert.NoError(t, err, "could not disconnect from windows svc API")
+
+	assert.False(t, svcExists(t, bootstrapper.KubeletServiceName))
+}
+
+// testUninstallWithoutKubeletSvc tests if WMCB returns an error if the kubelet is uninstalled with no kubelet service present
+func testUninstallWithoutKubeletSvc(t *testing.T) {
+	if svcExists(t, bootstrapper.KubeletServiceName) {
+		t.Skip("Skipping as kubelet service already exists")
+	}
+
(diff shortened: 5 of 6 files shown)


internal/test/wmcb/wmcb_test.go
```diff
@@ -122,6 +124,7 @@ func (vm *wmcbVM) runTest(testCmd string) error {
 		return fmt.Errorf("error running test: %v: %s", err, output)
 	}
 	if strings.Contains(output, "FAIL") {
+		log.Print(output)
```
mansikulkarni96, line 127: note: remove on next push

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,54 @@
+package main
+
+import (
+	"flag"
+	"os"
+
+	"github.com/openshift/windows-machine-config-bootstrapper/pkg/bootstrapper"
```
VaishnaviHire, line 7: this import can be moved to a separate section

pkg/bootstrapper/bootstrapper.go
```diff
@@ -38,6 +38,8 @@ const (
 	KubeletServiceName = "kubelet"
 	// kubeletDependentSvc is the name of the service dependent on kubelet Windows service
 	kubeletDependentSvc = "hybrid-overlay-node"
+	// hybridOverlayDependentSvc is the name of the service dependent on hybrid-overlay-node Windows service
```
VaishnaviHire, line 41: Should this change be a part of separate commit ?

VaishnaviHire (commented):
Thanks @mansikulkarni96 for working on this. PTAL at my comments

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,54 @@
...
+		Run: runUninstallKubeletCmd,
+	}
+)
+
+func init() {
+	rootCmd.AddCommand(uninstallKubeletCmd)
+}
+
+// runUninstallKubeletCmd uninstalls kubelet service from the Windows node
+func runUninstallKubeletCmd(cmd *cobra.Command, args []string) {
```
aravindhp, line 28: I suggest following the existing pattern of:
```go
bootstrapper.NewWinNodeBootstrapper()
wmcb.UninstallKubelet()
wmcb.Disconnect()
```
Otherwise you are leaking implementation details here.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -179,20 +181,33 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
...
-			return nil, fmt.Errorf("error updating kubelet dependents field %v", err)
-		}
-		bootstrapper.kubeletSVC, err = newKubeletService(ksvc, dependents)
-		if err != nil {
-			return nil, fmt.Errorf("could not initialize struct kubeletService: %v", err)
-		}
+	// If there is already a kubelet service running, find and assign it
+	bootstrapper.kubeletSVC, err = AssignExistingKubelet(svcMgr)
+	if err != nil {
+		return nil, fmt.Errorf("could not assign existing kubelet service: %v", err)
```
aravindhp, line 187: This refactor can be a separate commit.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -179,20 +181,33 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
...
 	}
 	return &bootstrapper, nil
 }
 
+// AssignExistingKubelet finds the existing kubelet service from the Windows Service Manager,
+// assigns its value to the kubeletService struct and returns it.
+func AssignExistingKubelet(svcMgr *mgr.Mgr) (*kubeletService, error) {
+	ksvc, err := svcMgr.OpenService(KubeletServiceName)
+	if err != nil {
+		// Do not return error if the service is not already installed
```
aravindhp, line 197: This is dangerous as the caller could try to dereference the kubelet service object which is nil at this point. You should rather return a specific error.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -876,17 +891,23 @@ func (cni *cniOptions) configure(kubeletCmd *string) error {
...
-	if err != nil {
-		// Do not return error if the services are not installed.
-		if !strings.Contains(err.Error(), "service does not exist") {
-			return nil, fmt.Errorf("error getting dependent services for kubelet %v", err)
+	// existingDependents is a list of dependent services that exist on the node
+	var existingDependents []*mgr.Service
+	// expectedDependents is a list of dependent services that could exist
+	// on the node, loop through this list to find the existing list of dependents
+	expectedDependents := []string{hybridOverlayDependentSvc, kubeletDependentSvc}
+	// find the existing service dependents
```
aravindhp, line 899: For one, this sort of refactor should be a separate commit. It is also not clear why you are considering the transitive dependency, `hybridOverlayDependentSvc`, a direct dependent of the kubelet service.

pkg/bootstrapper/kubelet.go
```diff
@@ -183,6 +183,36 @@ func (k *kubeletService) setRecoveryActions() error {
 	return nil
 }
 
+// remove deletes the kubelet service via the Windows service API
+func (k *kubeletService) remove() error {
```
aravindhp, line 187: Why do you need this method if it is never being used?

test/e2e/configure_cni_test.go
```diff
@@ -25,6 +26,10 @@ func TestConfigureCNI(t *testing.T) {
 	t.Run("Configure CNI", testConfigureCNI)
 }
 
+func TestKubeletUninstall(t *testing.T) {
```
aravindhp, line 29: It is odd that `TestKubeletUninstall` is added to `configure_cni_test.go`. Shouldn't this be a separate test file?

test/e2e/configure_cni_test.go
```diff
@@ -86,6 +91,22 @@ func testConfigureCNI(t *testing.T) {
...
+		svcMgr, err := mgr.Connect()
+		require.NoError(t, err)
+		defer svcMgr.Disconnect()
+		kubeletSvc, err := bootstrapper.AssignExistingKubelet(svcMgr)
+		require.NoError(t, err)
+		require.NotNil(t, kubeletSvc)
+
+		err = kubeletSvc.UninstallKubeletService()
+		assert.NoError(t, err, "unable to uninstall kubelet service")
+		assert.False(t, svcExists(t, bootstrapper.KubeletServiceName))
```
aravindhp, line 106: Here too you should be using the pattern:
```go
bootstrapper.NewWinNodeBootstrapper()
wmcb.UninstallKubelet()
wmcb.Disconnect()
```
This is how someone using WMCB as a library would use it.

aravindhp (changes requested):
@mansikulkarni96 thanks for working on this. Please address my comments.

cmd/bootstrapper/uninstall_kubelet.go
mansikulkarni96: I agree that was my initial thought but there would be a lot of parameters in calling bootstrapper.NewWinNodeBootstrapper() which would be not be required for uninstall eg: k8sInstallDir, ignitionFile, kubeletPath and would have to be passed as empty strings.

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: I ll check for a specific error here something like 'not installed'.

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: Doing this in simplest form possible here, I need to stop kube-proxy-> hybrid-overlay -> kubelet to stop kubelet service due to the dependency order. I can add comments to clarify this and add a seperate commit.

pkg/bootstrapper/kubelet.go
mansikulkarni96: this method is being used in k.stopAndRemove()

mansikulkarni96 (commented):
Thank you for the review @aravindhp , I have addressed the comments that could require discussion, will add changes for the rest.

cmd/bootstrapper/uninstall_kubelet.go
aravindhp: Think of WMCB as a library and how a user would call. This makes it way too complicated from that perspective. Having empty strings is not a reason for doing this. We can document that the inputs will be ignored if using the uninstall kubelet functionality.

pkg/bootstrapper/bootstrapper.go
aravindhp: I feel that the WMCB should return an error and not try to figure out the dependency graph. It should only be aware of the kubelet service. If it cannot stop that then it should return an error. It is up to the caller (WMCO)  to ensure that kube-proxy and hybrid-overlay are stopped before making this call.

pkg/bootstrapper/kubelet.go
aravindhp: I should have been more clear. It is only being used in `k.stopAndRemove()`. Given how little code there is it is fine to just have `k.stopAndRemove()`. But if `k.remove()` was becoming large then yes, you should split it out, which is not the case here.

pkg/bootstrapper/kubelet.go
```diff
@@ -183,6 +183,36 @@ func (k *kubeletService) setRecoveryActions() error {
...
+	}
+	return k.obj.Delete()
+}
+
+// stopAndRemove stops and removes the kubelet service
+func (k *kubeletService) stopAndRemove() error {
+	if k.obj == nil {
+		return nil
+	}
+	k.stop()
```
aravindhp, line 199: Don't you need to check for error here?

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: @aravindhp that makes sense, I ll drop handling of dependencies from here.

pkg/bootstrapper/kubelet.go
mansikulkarni96: thanks for clarifying, I ll drop this method.

cmd/bootstrapper/uninstall_kubelet.go
mansikulkarni96: that works.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -179,20 +179,35 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
...
 	}
 	return &bootstrapper, nil
 }
 
+// assignExistingKubelet finds the existing kubelet service from the Windows Service Manager,
+// assigns its value to the kubeletService struct and returns it.
+func assignExistingKubelet(svcMgr *mgr.Mgr) (*kubeletService, error) {
+	ksvc, err := svcMgr.OpenService(KubeletServiceName)
+	if err != nil {
+		// Do not return error if the services are not installed.
```
aravindhp, line 195: `service is not installed`

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,51 @@
...
+	flag.Parse()
+	// The inputs to NewWinNodeBootstrapper() are ignored while using the uninstall kubelet functionality
```
aravindhp, line 29: This doc needs to be added to `NewWinNodeBootstrapper`

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,51 @@
...
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper("", "", "", "",
+		"")
```
aravindhp, line 31: Please move to the previous line

pkg/bootstrapper/bootstrapper.go
```diff
@@ -666,6 +666,16 @@ func (wmcb *winNodeBootstrapper) Disconnect() error {
 	return err
 }
 
+// UninstallKubelet uninstalls the kubelet service from Windows node
+func (wmcb *winNodeBootstrapper) UninstallKubelet() error {
+	// Stop and remove kubelet service if it is in Running state.
+	err := wmcb.kubeletSVC.stopAndRemove()
```
aravindhp, line 672: Please add a `nil` check for `kubeletSVC`?

internal/test/wmcb/wmcb_test.go
```diff
@@ -109,6 +109,8 @@ func (vm *wmcbVM) runE2ETestSuite(t *testing.T) {
 	vm.runTestBootstrapper(t)
```
aravindhp, line 109: Shouldn't we add a test that tries to uninstall the kubelet and asserts that it is handled gracefully when no kubelet service is present?

aravindhp (changes requested):
Looks much better now, @mansikulkarni96

pkg/bootstrapper/bootstrapper.go
```diff
@@ -179,7 +179,7 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 		}
 	}
 
-	// If there is already a kubelet service running, find and assign it
+	// If there is already a kubelet service running, find it
```
aravindhp, line 182: Nit: this should be in the previous commit.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -667,6 +667,19 @@ func (wmcb *winNodeBootstrapper) Disconnect() error {
 	return err
 }
 
+// UninstallKubelet uninstalls the kubelet service from Windows node
+func (wmcb *winNodeBootstrapper) UninstallKubelet() error {
+	if wmcb.kubeletSVC == nil {
+		return fmt.Errorf("kubelet service is not present")
```
aravindhp, line 673: We should remember to ignore this error when using this in WMCO.

aravindhp (approved):
/approve

pkg/bootstrapper/bootstrapper.go
mansikulkarni96: done

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/openshift/windows-machine-config-bootstrapper/pkg/bootstrapper"
+)
+
+var (
+	// uninstallKubeletCmd describes the uninstall-kubelet command
+	uninstallKubeletCmd = &cobra.Command{
+		Use:   "uninstall-kubelet",
+		Short: "Uninstalls the kubelet service from the Windows node",
+		Long:  "Uninstalls the kubelet service from the Windows node",
```
sebsoto, line 17: `Stops and removes the kubelet service` maybe?

Windows documentation seems to use the term `remove` when it comes to services and not `uninstall`.
Dont think mentioning the Windows node is necessary as this is run directly on the Windows VM.

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// uninstall kubelet Windows service
+	err = wmcb.UninstallKubelet()
+	if err != nil {
+		log.Error(err, "could not uninstall kubelet service")
+		os.Exit(1)
+	}
+
+	// Send success message to StdOut to ascertain that kubelet removal was successful
+	os.Stdout.WriteString("kubelet service uninstalled successfully")
```
sebsoto, line 43: I think its fine to either say `kubelet service removed sucessfully` or `kubelet uninstalled sucessfully`

cmd/bootstrapper/uninstall_kubelet.go
```diff
@@ -0,0 +1,49 @@
...
+
+	err = wmcb.Disconnect()
+	if err != nil {
```
sebsoto, line 46: nit: I would prefer using this format if  nothing but err is being returned
```suggestion
	if err = wmcb.Disconnect(); err != nil {
```

sebsoto: Applies to https://github.com/openshift/windows-machine-config-bootstrapper/pull/259/files#diff-7bd684eea45943c86e1fa622be0ac5b58ebb684b7059e40d5b332b942b5ed64aR36 as well
`err = wmcb.UninstallKubelet()`

test/e2e/uninstall_kubelet_test.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		wmcb, err := bootstrapper.NewWinNodeBootstrapper("", "", "", "", "")
+		require.NoError(t, err, "could not create wmcb")
+
+		err = wmcb.UninstallKubelet()
+		assert.Error(t, err, "no error when attempting to uninstall kubelet without kubelet svc")
+		assert.Contains(t, err.Error(), "kubelet service is not present", "incorrect error thrown")
+	})
+}
```
sebsoto, line 46: You shouldnt be starting test functions with a `t.Run`. In general `t.Run` should be used to divide separate test cases under a common test parent.
In this code you are introducing essentially 2 tests. `Uninstall kubelet without kubelet service present` and `Uninstall Kubelet`. But these tests are being introduced as `TestBootstrapper/Uninstall_kubelet_without_kubelet_service_present/Kubelet_service_uninstalls` and  `TestKubeletUninstall/Uninstall_Kubelet/Kubelet_service_unistalls`. Theres no need for this complexity.

Instead you can have the two new tests be `TestBootstrapper/Kubelet_service_uninstalls` and `TestKubeletUninstallWithoutKubeletSVC`

sebsoto (changes requested):

test/e2e/uninstall_kubelet_test.go
sebsoto: `You shouldnt be starting test functions with a t.Run.` This statement is true only in this context, where there is only one test occuring within the the function.

mansikulkarni96: thanks for pointing this out, modified the test calls.

test/e2e/uninstall_kubelet_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openshift/windows-machine-config-bootstrapper/pkg/bootstrapper"
+)
+
+func TestKubeletUninstall(t *testing.T) {
+	t.Run("Uninstall Kubelet", testKubeletUninstall)
+}
```
sebsoto, line 14: I think you can take this one step further and remove this function and rename `testUninstallWithoutKubeletSvc` to `TestUninstallWithoutKubeletSvc`

mansikulkarni96: I think you meant TestKubeletUninstall, I have already modified TestBootstrapper/testUninstallWithoutKubeletSvc

sebsoto (commented):
/lgtm

PR #278, opened 2021-08-18
[wmcb] Allow forcing Windows node object IP 

sebsoto:
* Refactor to make kubelet args unit-testable 
* Allow forcing Windows node object IP

### DIFF
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -127,8 +127,8 @@ type winNodeBootstrapper struct {
 	// logDir is the directory that captures log outputs of Kubelet
 	// TODO: make this directory available in Artifacts
 	logDir string
-	// kubeletArgs is a map of the variable arguments that will be passed to the kubelet
-	kubeletArgs map[string]string
+	// kubeletArgs is a slice of the arguments that will be passed to the kubelet
+	kubeletArgs []string
 	// cni holds all the CNI specific information
 	cni *cniOptions
 }
@@ -169,7 +169,6 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath string, cni
 		logDir:             "C:\\var\\log\\kubelet",
 		initialKubeletPath: kubeletPath,
 		svcMgr:             svcMgr,
-		kubeletArgs:        make(map[string]string),
 	}
 	// populate the CNI struct if CNI options are present
 	if cniDir != "" && cniConfig != "" {
@@ -314,8 +313,8 @@ func convertIgnition2to3(ign2config ignitionCfgv2_4Types.Config) (ignitionCfgv3T
 	return ign3_1config, nil
 }
 
-// parseIgnitionFileContents parses the ignition file contents and writes the contents of the described files to the k8s
-// installation directory
+// parseIgnitionFileContents parses the ignition file contents gathering the required kubelet args, and writing
+// the contents of the described files to the k8s installation directory
 func (wmcb *winNodeBootstrapper) parseIgnitionFileContents(ignitionFileContents []byte,
 	filesToTranslate map[string]fileTranslation) error {
 	// Parse raw file contents for Ignition spec v3.1 config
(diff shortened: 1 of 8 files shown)


pkg/bootstrapper/bootstrapper.go
```diff
@@ -522,6 +526,9 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
 	if nodeWorkerLabel, ok := args["node-labels"]; ok {
 		kubeletArgs = append(kubeletArgs, "--"+"node-labels"+"="+nodeWorkerLabel)
 	}
+	if wmcb.nodeIP != "" {
+		kubeletArgs = append(kubeletArgs, "--node-ip="+wmcb.nodeIP)
```
saifshaikh48, line 530: Currently it's not clear to me where `nodeIP` is being set. I assume the value will be passed in by WMCO when it runs `wmcb.exe --node-ip= ...`?

saifshaikh48 (commented):
/lgtm

Left a question just to solidify my understanding. Also, I am fine with the limited scope refactor, as long as we make sure that the further work is tracked in Jira.

pkg/bootstrapper/bootstrapper.go
sebsoto: Yes that will passed in via WMCO when running WMCB

pkg/bootstrapper/bootstrapper.go
```diff
@@ -522,15 +543,13 @@ func (wmcb *winNodeBootstrapper) ensureKubeletService() error {
 		Password:         "",
 		Description:      "OpenShift Kubelet",
 	}
-	// Get kubelet args
-	kubeletArgs := wmcb.getInitialKubeletArgs()
 
 	if wmcb.kubeletSVC == nil {
-		if err := wmcb.createKubeletService(c, kubeletArgs); err != nil {
+		if err := wmcb.createKubeletService(c, wmcb.kubeletArgs); err != nil {
```
aravindhp, line 555: If you are making the actual `kubeletArgs` part of the `winNodeBootstrapper`, it should not be passed as an argument to a struct method

pkg/bootstrapper/bootstrapper.go
```diff
@@ -336,50 +335,42 @@ func (wmcb *winNodeBootstrapper) parseIgnitionFileContents(ignitionFileContents
...
-			filesToTranslate[results[1]] = fileTranslation{
-				dest: filepath.Join(wmcb.installDir, cloudConfFilename),
-			}
-
-			// Set the --cloud-config option value
-			wmcb.kubeletArgs[cloudConfigOption] = filepath.Join(wmcb.installDir, cloudConfFilename)
+	// Check for the presence of "--cloud-config" option and if it is present append the value to
+	// filesToTranslate. This option is only present for Azure and hence we cannot assume it as a file that
+	// requires translation across clouds.
+	if cloudConfigPath, ok := args[cloudConfigOption]; ok {
```
aravindhp, line 360: Shouldn't this now be done in `parseKubeletArgs()` and have one spot where there args are being created?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -522,6 +526,9 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
 	if nodeWorkerLabel, ok := args["node-labels"]; ok {
 		kubeletArgs = append(kubeletArgs, "--"+"node-labels"+"="+nodeWorkerLabel)
 	}
+	if wmcb.nodeIP != "" {
```
aravindhp, line 529: Shouldn't we validate the `nodeIP` rather than blindly using it?

aravindhp (changes requested):
Thanks for working on this, @sebsoto. PTAL at my comments.

pkg/bootstrapper/bootstrapper.go
sebsoto: Added it to the `newWinNodeBootstrapper` function

pkg/bootstrapper/bootstrapper.go
sebsoto: I added a TODO explaining why it is currently like that and link to the new refactor ticket https://issues.redhat.com/browse/WINC-670

sebsoto: @saifshaikh48 ticket that you asked for^

saifshaikh48 (commented):
/lgtm

Thanks for working on this @sebsoto

pkg/bootstrapper/bootstrapper.go
```diff
@@ -461,9 +485,9 @@ func (wmcb *winNodeBootstrapper) initializeKubeletFiles() error {
 	return nil
 }
 
-// getInitialKubeletArgs returns the kubelet args required during initial kubelet start up.
-// This function assumes that wmcb.kubeletArgs are populated.
-func (wmcb *winNodeBootstrapper) getInitialKubeletArgs() []string {
+// generateInitialKubeletArgs returns the kubelet args required during initial kubelet start up. args should be a map
```
mansikulkarni96, line 500: args -> Args

mansikulkarni96 (commented):
Thanks for working on this @sebsoto 
LGTM except for a nit.

pkg/bootstrapper/bootstrapper.go
sebsoto: I did lowercase since I was referring to the variable `args` which is lowercase

aravindhp (approved):
/approve

sebsoto:
/hold cancel
Seeing no change in flags besides --node-ip flag

PR #279, opened 2021-08-18
[test] Fix handleHybridOverlay()

aravindhp:
Sync up to match up with how we are [configuring hybrid-overlay in WMCO](https://github.com/openshift/windows-machine-config-operator/blob/a094884abffac047c1199dce1281969a1e320682/pkg/windows/windows.go#L362):
- Add a sleep to wait for the network reconfiguration to complete
- Change the ordering of reinitialization of the SSH connection
- Remove a stale comment which is remnant of the WNI / WinRM days

### DIFF
--- a/internal/test/wmcb/wmcb_test.go
+++ b/internal/test/wmcb/wmcb_test.go
@@ -12,6 +12,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
@@ -48,6 +49,9 @@ const (
 	hybridOverlayName = "hybrid-overlay-node.exe"
 	// hybridOverExecutable is the remote location of the hybrid overlay binary
 	hybridOverlayExecutable = remoteDir + hybridOverlayName
+	// hybridOverlayConfigurationTime is the approximate time taken for the hybrid-overlay to complete reconfiguring
+	// the Windows VM's network
+	hybridOverlayConfigurationTime = 2 * time.Minute
 )
 
 var (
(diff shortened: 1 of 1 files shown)


mansikulkarni96:
Changes LGTM.
Do we still need to run hybrid overlay in the [background](https://github.com/openshift/windows-machine-config-bootstrapper/blob/0d5d2f4fce523dfb6978b78c64705c9a123836ea/internal/test/wmcb/wmcb_test.go#L234), was the reason for doing that because we weren't doing this reinitialization or will that require further investigation?

saifshaikh48:
/lgtm
/hold

Hold for Mansi's question above.

aravindhp:
> Changes LGTM.
> Do we still need to run hybrid overlay in the [background](https://github.com/openshift/windows-machine-config-bootstrapper/blob/0d5d2f4fce523dfb6978b78c64705c9a123836ea/internal/test/wmcb/wmcb_test.go#L234), was the reason for doing that because we weren't doing this reinitialization or will that require further investigation?

@mansikulkarni96, yes we need still need to run hybrid-overlay in the background or as a service. That is what happens when you run it as a service like we are doing in WMCO. I don't fully understand rest of the question though. We were always doing the re initialization but in a different order.

mansikulkarni96:
Thanks for the explanation @aravindhp my question came from the comment [here](https://github.com/openshift/windows-machine-config-bootstrapper/blob/0d5d2f4fce523dfb6978b78c64705c9a123836ea/internal/test/wmcb/wmcb_test.go#L231). But I now realize that comment is dated and might not be relevant to the current code.

aravindhp:
Ah yes, that comment is outdated. I can remove it. as part of this PR.

mansikulkarni96:
Thanks for fixing this.
/lgtm

PR #304, opened 2021-10-26
[OCPCLOUD-1247] Allow to override kubelet hostname

Fedosin:
This commit adds a new parameter for the bootstrapper: `--hostname`. It allows to define custom hostname for kubelet on the node.

It is required for deploying clusters on Windows machines with Cloud Controller Manager support, where hostnames must be the same as node names.

Related links:
https://issues.redhat.com/browse/OCPCLOUD-1247
https://bugzilla.redhat.com/show_bug.cgi?id=1995894

### DIFF
--- a/test/e2e/bootstrapper_test.go
+++ b/test/e2e/bootstrapper_test.go
@@ -62,7 +62,7 @@ func TestBootstrapper(t *testing.T) {
 	t.Run("Uninstall kubelet without kubelet service present", testUninstallWithoutKubeletSvc)
 
 	// Run the bootstrapper, which will start the kubelet service
-	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "")
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "", "my-computer")
 	require.NoErrorf(t, err, "Could not create WinNodeBootstrapper: %s", err)
 	err = wmcb.InitializeKubelet()
 	assert.NoErrorf(t, err, "Could not run bootstrapper: %s", err)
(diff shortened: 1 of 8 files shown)


JoelSpeed:
For additional context, on some platforms (eg AWS, OpenStack), the default hostname of the instance (which is what Kubelet uses with `--cloud-provider=external`) does not match the node name which is given to kubelet by the in-tree cloud providers. To ensure we keep the node names consistent between in-tree and out-of-tree providers, we have used the `--hostname-override` flag on linux nodes by fetching data from the providers metadata services to grab the correct node name.

If we upgrade existing nodes and the node name changes between in-tree and out-of-tree, the Kubelet becomes broken as the client credentials renewal flow requests a name change, which is rejected by the CSR approval system within OpenShift. Equally, the kubelet would try to create a new Node object, orphaning the old Node. Therefore, we must keep the names consistent between in-tree and out-of-tree.

This PR is aiming to help us in that flow by allowing us to set the flag for the windows kubelets

test/e2e/bootstrapper_test.go
```diff
@@ -211,3 +214,19 @@ func testPathInKubeletArgs(t *testing.T, checkPathsFor []string, path string) {
...
+// testHostname checks if the hostname given as the argument to kubelet service is correct
+func testHostname(t *testing.T, path string) {
+	// Split the arguments from kubelet path
+	kubeletArg := strings.Split(path, " ")
+	for _, arg := range kubeletArg {
+		// Split the key and value of arg
+		argSplit := strings.SplitN(arg, "=", 2)
+		// Ignore single valued arguments
+		if len(argSplit) > 1 {
+			if argSplit[0] == "--hostname-override" {
```
sebsoto, line 227: If this test is only ran when the `hostname-override` argument is supposed to be set, I  would like to see this test fail if the `hostname-override` flag is missing

test/e2e/bootstrapper_test.go
```diff
@@ -211,3 +214,19 @@ func testPathInKubeletArgs(t *testing.T, checkPathsFor []string, path string) {
...
+				assert.Equal(t, argSplit[1], "my-computer")
```
sebsoto, line 228: `my-computer` should either be an argument or a constant so it only needs to be maintained in one place

sebsoto (changes requested):
Thanks @Fedosin, looks mostly good, just some comments on the tests

PR #306, opened 2021-10-29
Bug 1994859: Parametrize Cluster DNS

jrvaldes:
The hard-coded Cluster DNS value from the kubelet_conf.json
template is replaced with a placeholder that will be populated while
processing the template.

Introduce a required flag to the initialize-kubelet command, to hold the
value used to configure the kubelet with the specified Cluster DNS address.

Fixes BZ#1994859


pkg/bootstrapper/bootstrapper.go
```diff
@@ -172,6 +174,13 @@ func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath, nodeIP, cn
 		}
 	}
 
+	// // If clusterDNSIP is set, ensure that it is a valid IP
```
sebsoto, line 177: double comment

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -24,6 +24,10 @@ var (
 			if err != nil {
 				return err
 			}
+			err = cmd.MarkPersistentFlagRequired("cluster-dns-ip")
+			if err != nil {
+				return err
+			}
```
sebsoto, line 30: You're marking it as required here, but I don't think it should be considering it wasn't needed in the past. Not setting it is valid, in that it will use the default value decided by the kubelet

pkg/bootstrapper/bootstrapper.go
```diff
@@ -253,6 +263,8 @@ type fileTranslation struct {
 type kubeletConf struct {
 	// ClientCAFile specifies location to client certificate
 	ClientCAFile string
+	// clusterDNSIP is the IP address used to configure the kubelet with a specific clusterDNS IP
```
sebsoto, line 266: Considering this struct describes the kubelet configuration I feel you can drop the `configure the kubelet` part, and instead go with something like:
`clusterDNSIP is the IP address of the DNS server used for all containers`

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -51,6 +57,8 @@ func init() {
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.nodeIP, "node-ip", "",
 		"nodeIP is the IP that should be used as the node object's IP. "+
 			"If unset, kubelet will determine the IP itself.")
+	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.clusterDNSIP, "cluster-dns-ip", "",
+		"IP address used to configure the kubelet with a specific clusterDNS IP")
```
sebsoto, line 61: Would prefer this to describe the end results for the user. So something like:
`The DNS server IP which will be used by all containers, instead of the host's DNS servers`

Paraphrasing from https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

docs/README.md
```diff
@@ -40,7 +40,7 @@ make build
 ```
 
 ```
-wmcb initialize-kubelet --ignition-file $IGNITION_FILE_PATH --kubelet-path $KUBELET_PATH
+wmcb initialize-kubelet --cluster-dns-ip $CLUSTER_DNS_IP --ignition-file $IGNITION_FILE_PATH --kubelet-path $KUBELET_PATH
```
sebsoto, line 43: You need to document how to get the value for this flag as its not immediately obvious

sebsoto: Additionally it would be nice to make it optional as suggested above

pkg/bootstrapper/bootstrapper_test.go
```diff
@@ -176,7 +176,10 @@ func TestCreateKubeletConf(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			bs := winNodeBootstrapper{installDir: instDir}
+			bs := winNodeBootstrapper{
+				installDir:   instDir,
+				clusterDNSIP: "172.30.0.10",
```
sebsoto, line 181: Where is this IP coming from? If this specific value is needed please make it a constant, if any value works please comment.

sebsoto (changes requested):
Thanks @jrvaldes looks mostly good

cmd/bootstrapper/initialize_kubelet.go
jrvaldes: Valid point; which I considered. 

The default value decided by the kubelet  is :[the host's DNS servers](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/)

pkg/bootstrapper/bootstrapper_test.go
jrvaldes: The `172.30.0.10` value  is coming from the [template](https://github.com/openshift/windows-machine-config-bootstrapper/pull/306/files/39bc32c9bc83fd020254cfb3be0633f8f6b79980#diff-c28fa61c93e3f117dea1a808854d6b9309175dc85d833e2a67b6e33ff2d92bfaL174) used in the base case test.

saifshaikh48 (commented):
/lgtm

Thanks for working on this @jrvaldes

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -37,6 +37,8 @@ var (
 		installDir string
 		// nodeIP directs the kubelet to use a specific IP for the node object
 		nodeIP string
+		// clusterDNSIP is the IP address of the DNS server used for all containers
+		clusterDNSIP string
```
aravindhp, line 41: https://github.com/openshift/windows-machine-config-operator/pull/777#discussion_r741493915 applies here also.

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -51,6 +53,9 @@ func init() {
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.nodeIP, "node-ip", "",
 		"nodeIP is the IP that should be used as the node object's IP. "+
 			"If unset, kubelet will determine the IP itself.")
+	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.clusterDNSIP, "cluster-dns-ip", "",
+		"The DNS server IP which will be used by all containers. "+
+			"If unset, kubelet will sue the host's DNS servers")
```
aravindhp, line 58: You need say how WMCB is going to use this field and not how the kubelet uses it. This also needs to be reflected in the README.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -158,7 +160,7 @@ type cniOptions struct {
 // optional node IP, along with the CNI options as inputs, and generates the winNodeBootstrapper object. The CNI options
 // are populated only in the configure-cni command. The inputs to NewWinNodeBootstrapper are ignored while using the
 // uninstall kubelet functionality.
-func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath, nodeIP, cniDir,
+func NewWinNodeBootstrapper(k8sInstallDir, ignitionFile, kubeletPath, nodeIP, clusterDNSIP, cniDir,
```
aravindhp, line 163: Please update the doc wrt the new parameter addition

aravindhp (changes requested):
Thanks for working on this, @jrvaldes. Please address my comments.

saifshaikh48 (commented):
/lgtm

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -51,6 +53,9 @@ func init() {
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.nodeIP, "node-ip", "",
 		"nodeIP is the IP that should be used as the node object's IP. "+
 			"If unset, kubelet will determine the IP itself.")
+	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.clusterDNS, "cluster-dns", "",
+		"The DNS server IP passed to kubelet, that will be used to configure all containers for DNS resolution "+
+			"instead of the host's DNS server. The latter, is the default kubelet configuration.")
```
aravindhp, line 58: You should call out what happens when it is not specified similar to the `node-ip`.

>The DNS server IP passed to kubelet, that will be used to configure all containers for DNS resolution. If unset, kubelet will determine the DNS server to use.

The determination should be left to the kubelet

docs/README.md
```diff
@@ -48,6 +48,10 @@ wmcb configure-cni --cni-dir $CNI_BIN_DIR --cni-config $CNI_CONFIG
 after `configure-cni` is executed, all the CNI options will be removed. This is to give the user a chance to change
 network configuration to something other than CNI after the initial setup.
 
+The `initialize-kubelet` command provides the following optional parameters:
+- `--cluster-dns` is the DNS server IP passed to kubelet, that will be used to configure all containers for DNS 
+  resolution instead of the host's DNS server. The latter, is the default [KubeletConfiguration](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration).
```
aravindhp, line 53: >The DNS server IP passed to kubelet, that will be used to configure all containers for DNS resolution. If unset, kubelet will [determine](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration) the DNS server to use.

Can be used here also.

jrvaldes: Good point, updated both.

aravindhp (approved):
/approve

I suggest calling out `clusterDNS` is the kubelet option in the commit and README. The reference link points at the full list.

saifshaikh48 (commented):
/lgtm

@jrvaldes Thanks for this fix!

PR #308, opened 2021-11-11
OCPCLOUD-1247: Set correct kubelet hostname for AWS platform

Fedosin:
It allows to define type for the platform where the cluster is deployed.

If platform type is AWS, then we explicitly override kubelet hostname with the value from Metadata.

It is required for deploying clusters on Windows machines with Cloud Controller Manager support, where hostnames must be the same as node names.

Related links:
https://issues.redhat.com/browse/OCPCLOUD-1247
https://bugzilla.redhat.com/show_bug.cgi?id=1995894


pkg/bootstrapper/bootstrapper.go
```diff
@@ -403,7 +406,10 @@ func (wmcb *winNodeBootstrapper) parseIgnitionFileContents(ignitionFileContents
 	}
 
 	// Generate the full list of kubelet arguments from the arguments present in the ignition file
-	wmcb.kubeletArgs = wmcb.generateInitialKubeletArgs(args)
+	wmcb.kubeletArgs, err = wmcb.generateInitialKubeletArgs(args)
+	if err != nil {
+		return fmt.Errorf("cannot generate initial kubelet args: %w\n", err)
```
JoelSpeed, line 411: It's not usual to add a new line to an error string like this, what's the intention here?

Also, is there a particular purpose to wrapping the error? Are you likely to unwrap later? If not, just use `%v` instead, it's more idiomatic

pkg/bootstrapper/bootstrapper.go
```diff
@@ -558,7 +564,21 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
...
+
+	var hostname string
+	var err error
+	switch wmcb.platformType {
+	case "AWS":
+		hostname, err = getAWSMetadataHostname()
+		if err != nil {
+			return nil, err
+		}
+	}
```
JoelSpeed, line 576: This switch could grow over time, so as not to distract from the purpose of this function, why not make this a helper?

```
func getKubeletHostname(platform configv1.PlatformType) (string, error) {
  switch platform {
  case configv1.AWSPlatformType:
    return getAWSMetadataHostname()
  default:
    return "", nil
  }
}
```

pkg/bootstrapper/hostname.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+)
+
+func getAWSMetadataHostname() (string, error) {
```
JoelSpeed, line 12: Do we want to have the AWS specific functions within either a file with AWS in the name or a separate provider specific package?

pkg/bootstrapper/hostname.go
```diff
@@ -0,0 +1,34 @@
...
+	})
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve the hostname from the EC2 instance: %w\n", err)
+	}
+
+	defer hostnameRes.Content.Close()
+	hostname, err := io.ReadAll(hostnameRes.Content)
+	if err != nil {
+		return "", fmt.Errorf("cannot read hostname from the EC2 instance: %w\n", err)
+	}
```
JoelSpeed, line 31: Would recommend not wrapping the errors here and removing the the new lines as well

pkg/bootstrapper/bootstrapper.go
sebsoto: Agreed, I'd also like to see this moved to a separate package

pkg/bootstrapper/hostname.go
sebsoto: Another package for all providers would be preferred with a separate AWS file.

test/e2e/bootstrapper_test.go
```diff
@@ -62,7 +62,7 @@ func TestBootstrapper(t *testing.T) {
 	t.Run("Uninstall kubelet without kubelet service present", testUninstallWithoutKubeletSvc)
 
 	// Run the bootstrapper, which will start the kubelet service
-	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "")
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "", "")
```
sebsoto, line 65: The provider should be discovered as part of the test and then passed in to all invocations of `NewWinNodeBootstrapper`, excluding the uninstall kubelet tests, as arguments are ignored there.

A test should be added to check that the `hostname-override` argument is set.

Do we need to check the value? What is the expected behavior if the value is incorrect?

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -56,6 +58,8 @@ func init() {
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.clusterDNS, "cluster-dns", "",
 		"The DNS server IP passed to kubelet, that will be used to configure all containers for DNS resolution. "+
 			"If unset, kubelet will determine the DNS server to use.")
+	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.platformType, "platform-type", "",
+		"Type of the platform where the cluster is deployed.")
```
sebsoto, line 62: Can you add example values, such as `aws`?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -558,7 +564,21 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
 	if wmcb.nodeIP != "" {
 		kubeletArgs = append(kubeletArgs, "--node-ip="+wmcb.nodeIP)
 	}
-	return kubeletArgs
+
+	var hostname string
+	var err error
+	switch wmcb.platformType {
+	case "AWS":
```
sebsoto, line 571: lets make this  `aws` instead of `AWS`, as this is a flag being passed in, and all lowercase is more commonly used

sebsoto: This could potentially be made a constant in the new package you will introduce.

sebsoto (changes requested):

pkg/bootstrapper/bootstrapper.go
Fedosin: moved to a new package "cloud"

pkg/bootstrapper/hostname.go
Fedosin: done

pkg/bootstrapper/hostname.go
Fedosin: done

test/e2e/bootstrapper_test.go
Fedosin: I put "azure" platform as the parameter value there at the moment, in this case wcmb doesn't override the hostname. In case of aws, it will send requests to AWS metadata service, which is hard to emulate in these tests.

pkg/bootstrapper/cloud/cloud.go
```diff
@@ -0,0 +1,25 @@
...
+	"strings"
+)
+
+const (
+	awsPlatformType = "aws"
+)
+
+// GetOverriddenKubeletHostname returns correct hostname for kubelet if it should
+// be overridden, or an empty string otherwise.
+func GetOverriddenKubeletHostname(platformType string) (string, error) {
```
JoelSpeed: Nit: I'd probably call this `GetKubeletHostnameOverride` so that it looks more like the kubelet flag `hostname-override`, but I'll defer to WMCO folks as to whether they think that's a required change

pkg/bootstrapper/cloud/aws.go
```diff
@@ -0,0 +1,35 @@
...
+	cfg, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+		return "", fmt.Errorf("unable to load config: %w", err)
+	}
+
+	client := imds.NewFromConfig(cfg)
+
+	hostnameRes, err := client.GetMetadata(context.TODO(), &imds.GetMetadataInput{
+		Path: "hostname",
+	})
```
JoelSpeed: Is there a reference somewhere for why we are using this value? Perhaps a link to the old cloud provider code or something? It would be good to link the original source if possible

JoelSpeed:
Having watched the recording from the monday meeting, I wanted to add some notes about this to alleviate some concerns about supportability that were raised and some clarifications on technical details:
- With in-tree, Kubelet asks the cloud what name it should use for the node. In the AWS implementation, this uses the [PrivateDNSName](https://github.com/kubernetes/legacy-cloud-providers/blob/c0697affa6e0a5445bbde000c89e4d928d44969d/aws/aws.go#L4926-L4928) of the instance which is also used to populate the `hostname` field within the metadata. If you remove the cloud provider from the situation, we fall back to the OS hostname which may not match.
- The node name preference in kubelet, highest to lowest is: `hostname-override` > cloud provider `GetCurrentNodeName` > OS Hostname
- This PR maintains the existing node name in a backwards compatible way, and as such, should not affect any existing installations.

pkg/bootstrapper/cloud/cloud.go
Fedosin: Done

Fedosin:
"local-hostname" path is taken to comply with the MCO request: https://github.com/openshift/machine-config-operator/blob/master/templates/common/aws/files/usr-local-bin-aws-kubelet-nodename.yaml#L19

JoelSpeed:
Perhaps we should replicate the comment from https://github.com/openshift/machine-config-operator/blob/b567492ff2b14444c805a3d8210768540d8a64eb/templates/common/aws/files/usr-local-bin-aws-kubelet-nodename.yaml#L15-L18 to this code as well, and/or link to the MCO counterpart?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -558,7 +566,16 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
 	if wmcb.nodeIP != "" {
 		kubeletArgs = append(kubeletArgs, "--node-ip="+wmcb.nodeIP)
 	}
-	return kubeletArgs
+
+	hostname, err := cloud.GetKubeletHostnameOverride(wmcb.platformType)
+	if err != nil {
+		return nil, err
+	}
```
sebsoto, line 573: After seeing this, I am wondering if it would make more sense to introduce a cloud interface
with `GetKubeletHostnameOverride` as method

My original and current thoughts are that it would need to be done only if there were more cloud specific functions to be introduced. Is this the only cloud specific WMCB function required for the CCM work @Fedosin?

I could go either way, I will let others weigh in, if they have an opinion on this.

sebsoto (changes requested):
Thanks @Fedosin 
a few requests/questions

Makefile
```diff
@@ -11,7 +11,7 @@ build:
 
 .PHONY: build-wmcb-unit-test
 build-wmcb-unit-test:
-	$(GO_BUILD_ARGS) GOOS=windows GOFLAGS=-v go test -c ./pkg/... -o wmcb_unit_test.exe
+	$(GO_BUILD_ARGS) GOOS=windows GOFLAGS=-v go test -c ./pkg/bootstrapper/... -o wmcb_unit_test.exe
```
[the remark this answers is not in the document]
Fedosin: I think the problem here is that we `cannot use -c flag with multiple packages`. I.e. we can build a binary for just one package.
If we add unit tests for the "cloud" package, another binary has to be built.

test/e2e/bootstrapper_test.go
```diff
@@ -62,7 +62,7 @@ func TestBootstrapper(t *testing.T) {
 	t.Run("Uninstall kubelet without kubelet service present", testUninstallWithoutKubeletSvc)
 
 	// Run the bootstrapper, which will start the kubelet service
-	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "")
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "", "azure")
```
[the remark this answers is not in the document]
Fedosin: Oh, I got it, you run these test on AWS, so metadata service will be available. In this case, I changed the platform type to "aws" in the tests

pkg/cloud/cloud.go
```diff
@@ -0,0 +1,25 @@
...
+	awsPlatformType = "aws"
+)
+
+// GetKubeletHostnameOverride returns correct hostname for kubelet if it should
+// be overridden, or an empty string otherwise.
+func GetKubeletHostnameOverride(platformType string) (string, error) {
+	var hostname string
+	var err error
+	switch strings.ToLower(platformType) {
+	case awsPlatformType:
```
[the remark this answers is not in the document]
Fedosin: Done

pkg/bootstrapper/bootstrapper.go
Fedosin: Yeah, nothing else besides that is required. So, I don't think this interface is necessary.

pkg/cloud/cloud.go
```diff
@@ -0,0 +1,31 @@
...
+func GetKubeletHostnameOverride(platformType string) (string, error) {
+	var hostname string
+	var err error
+	platformType = strings.ToLower(platformType)
+	switch platformType {
+	case awsPlatformType:
+		hostname, err = getAWSMetadataHostname()
+		if err != nil {
+			return "", err
+		}
```
JoelSpeed, line 23: Why not just return here, then you don't need the final return or to have the vars on 15 and 16. I think this would make it cleaner
```suggestion
		return getAWSMetadataHostname()
```

test/e2e/bootstrapper_test.go
```diff
@@ -62,7 +62,7 @@ func TestBootstrapper(t *testing.T) {
 	t.Run("Uninstall kubelet without kubelet service present", testUninstallWithoutKubeletSvc)
 
 	// Run the bootstrapper, which will start the kubelet service
-	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "")
+	wmcb, err := bootstrapper.NewWinNodeBootstrapper(installDir, ignitionFilePath, kubeletPath, "", "", "", "", "aws")
```
JoelSpeed, line 65: Why not use the constant `configv1.AWSPlatformType` for this instead of a raw string?

Fedosin: we need to add openshift/api in the dependencies then...

Fedosin: with all its dependencies it is 110k lines of code (146 files changed, 109510 insertions(+), 553 deletions(-)) just for a string constant.

Fedosin: that's why I prefer not to do this

Makefile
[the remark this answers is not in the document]
aravindhp: Is there any option that allows us to build a unit test binary that runs through all the packages?

pkg/bootstrapper/bootstrapper.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/sys/windows/svc/mgr"
+
```
aravindhp, line 25: Unnecessary new line. This will prevent the formatter from organizing this correctly.

pkg/cloud/cloud.go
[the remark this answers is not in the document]
aravindhp: This is a bit strange. It makes it seem that the only supported platfrom is AWS. Should we think of `platformType` to be a required parameter? Internally it can be a no-op for certain platforms.

pkg/bootstrapper/bootstrapper.go
sebsoto: This is compliant with the regular formatting we do for imports

first group built in, second external imports, third imports from other packages within the module.

aravindhp: Ah yes. Sorry my bad. Please ignore.

Makefile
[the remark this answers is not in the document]
Fedosin: I don't think so, here they explain why it's not possible: https://github.com/golang/go/issues/15513#issuecomment-216694348

aravindhp:
/retitle OCPCLOUD-1247: Set correct kubelet hostname for AWS platform

To keep the current patterns we have in the repo. I also suggest prefixing the commits with `[vendor]` and `[bootstrapper]`. We try not to use Jira prefixes for the individual commits.

JoelSpeed (commented):
/lgtm

pkg/cloud/cloud.go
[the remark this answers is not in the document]
sebsoto: @Fedosin
@aravindhp made a good point here, the default case should return no error, as we should be able to pass any supported cloud platform from WMCO without error. Sorry for the misdirection :pray:

sebsoto (changes requested):
One last comment, will approve after its change

pkg/cloud/cloud.go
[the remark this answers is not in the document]
Fedosin: Ack! It's fixed now.

sebsoto:
/approve
Thank you for your work @Fedosin !

sebsoto:
/hold
to ensure ccm test passes

PR #312, opened 2022-01-07
WINC-699: [build] Upgrade to Golang 1.17

saifshaikh48:
- [build] Bump to Go 1.17 
- [vendor] Update go.mod for Go 1.17 bump


Dockerfile.tools
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/openshift/origin-release:golang-1.16 as build
+FROM registry.ci.openshift.org/openshift/release:golang-1.17 as build
```
aravindhp, line 1: Please add a line to the commit message saying that we no longer mirror to docker.io/openshift/ and are using registry.ci.openshift.org/openshift/ as that is also public. Otherwise we will forget why we made this change.

aravindhp (changes requested):
Thanks for doing this, @saifshaikh48. LGTM.

Dockerfile.tools
saifshaikh48: Sure, done.

PR #315, opened 2022-02-01
WINC-724: Add OpenShift managed to Kubelet SVC description

sebsoto:
The "openshift-managed" string must be added to the description of all
Windows services created as part of the Windows Node configuration
process. This is a requirement of properly upgrading Nodes. During the
upgrade process, a search can be done for all Windows services with the
string in their description, in order to identify and potentially remove
Services created by us.

### DIFF
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -95,6 +95,9 @@ const (
 	cniBinDirOption = "--cni-bin-dir"
 	// cniConfDirOption is to specify the CNI conf directory
 	cniConfDirOption = "--cni-conf-dir"
+	// managedServicePrefix indicates that the service being described is managed by OpenShift. This ensures that all
+	// services created as part of Node configuration can be searched for by checking their description for this string
+	managedServicePrefix = "OpenShift managed"
 )
 
 // These regex are global, so that we only need to compile them once
@@ -594,7 +597,7 @@ func (wmcb *winNodeBootstrapper) ensureKubeletService() error {
 		ServiceStartName: "",
 		DisplayName:      "",
 		Password:         "",
-		Description:      "OpenShift Kubelet",
+		Description:      fmt.Sprintf("%s kubelet", managedServicePrefix),
 	}
 
 	if wmcb.kubeletSVC == nil {
--- a/test/e2e/bootstrapper_test.go
+++ b/test/e2e/bootstrapper_test.go
@@ -111,6 +111,11 @@ func TestBootstrapper(t *testing.T) {
 	t.Run("Test the config dependencies in Kubelet arguments", func(t *testing.T) {
 		assert.ElementsMatch(t, expectedDependencies, actualDependencies)
 	})
+	t.Run("Kubelet service has 'OpenShift managed' in description", func(t *testing.T) {
+		config, err := getSvcConfig(bootstrapper.KubeletServiceName)
+		require.NoError(t, err, "error getting Kubelet Config")
+		assert.Contains(t, config.Description, "OpenShift managed")
+	})
 }
 
 // ensureIgnitionFileExists will create a generic ignition file if one is not provided on the node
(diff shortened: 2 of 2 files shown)


test/e2e/bootstrapper_test.go
```diff
@@ -180,6 +185,23 @@ func getSvcInfo(svcName string) (svc.State, string, []string, error) {
...
+	if err != nil {
+		return mgr.Config{}, fmt.Errorf("could not connect to Windows SCM: %s", err)
+	}
+	defer svcMgr.Disconnect()
+
+	svcHandle, err := svcMgr.OpenService(svcName)
+	if err != nil {
+		return mgr.Config{}, err
+	}
+	defer svcHandle.Close()
```
jrvaldes, line 200: This block can be modularized. See lines [160-170](https://github.com/openshift/windows-machine-config-bootstrapper/pull/315/files#diff-2f248ac78cf47e8f98e4c678f4208351bc12c9da943275b4be0a3ce500e31bfeR160-R170)

jrvaldes (commented):
@sebsoto  thanks for working on this, added a minor suggestion. Otherwise LGTM.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -594,7 +597,7 @@ func (wmcb *winNodeBootstrapper) ensureKubeletService() error {
 		ServiceStartName: "",
 		DisplayName:      "",
 		Password:         "",
-		Description:      "OpenShift Kubelet",
+		Description:      managedString,
```
saifshaikh48, line 600: Any reason not to just add "openshift-managed" to the existing service description?

sebsoto: The description was arbitrarily added when writing WMCB
it would look weird if it was changed to `OpenShift Kubelet openshift-managed`

Especially since the service name already calls it `kubelet`

test/e2e/bootstrapper_test.go
sebsoto: That change would require a larger refactor, across all the `getSvc*` functions, I thought about doing it, but decided against it, especially considering that this code will be taken out of service soon.

pkg/bootstrapper/bootstrapper.go
```diff
@@ -95,6 +95,9 @@ const (
 	cniBinDirOption = "--cni-bin-dir"
 	// cniConfDirOption is to specify the CNI conf directory
 	cniConfDirOption = "--cni-conf-dir"
+	// managedString indicates that the service being described is managed by OpenShift. This ensures that all services
+	// created as part of Node configuration can be searched for by checking their description for this string
+	managedString = "openshift-managed"
```
mansikulkarni96, line 100: nit: can remove string from the name here.

mansikulkarni96 (commented):
LGTM except a nit.

pkg/bootstrapper/bootstrapper.go
saifshaikh48: I'm also thinking in general, for other services with more informative descriptions, maybe prepending `managedString`? So here we could fix up kubelet description e.g. 
```
"openshift-managed: registers an instance as a node with the cluster's api-server"
```

aravindhp: I suggest something like `OpenShift managed Kubelet`.

aravindhp (approved):
Thanks for working on this, @sebsoto 

/approve

aravindhp:
/hold 
```
 === RUN   TestBootstrapper/Kubelet_service_has_'OpenShift_managed'_in_description
    bootstrapper_test.go:117: 
        	Error Trace:	bootstrapper_test.go:117
        	Error:      	"OpenShift managed" does not contain "OpenShift managed kubelet"
        	Test:       	TestBootstrapper/Kubelet_service_has_'OpenShift_managed'_in_description 
```

selansen (approved):
/lgtm

PR #322, opened 2022-02-03
WINC-696: Introduce dockerRuntime flag option to update kubelet arguments 

mansikulkarni96:
This PR adds the --dockerRuntime flag option to allow passing containerd specific kubelet config params
while starting kubelet service. This will also allow kubelet to set a dependency on the underlying container runtime present.
It also starts using the DOCKER_RUNTIME env var passed from the release repo in the CCM job config to set the flag value for --dockerRuntime and install containerd Windows service to test the kubelet config changes.

### DIFF
--- a/cmd/bootstrapper/initialize_kubelet.go
+++ b/cmd/bootstrapper/initialize_kubelet.go
@@ -41,6 +41,9 @@ var (
 		clusterDNS string
 		// platformType contains type of the platform where the cluster is deployed
 		platformType string
+		// dockerRuntime set to true indicates the container runtime to be used is docker.
+		// If set to false containerd will be the container runtime.
+		dockerRuntime bool
 	}
 )
 
@@ -60,6 +63,8 @@ func init() {
 			"If unset, kubelet will determine the DNS server to use.")
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.platformType, "platform-type", "",
 		"Type of the platform where the cluster is deployed. Example: AWS, Azure, GCP")
+	initializeKubeletCmd.PersistentFlags().BoolVar(&initializeKubeletOpts.dockerRuntime, "dockerRuntime", true,
+		"Use docker as container runtime. If unset, defaults to true. If false, containerd will be used as the container runtime.")
 }
 
 // runInitializeKubeletCmd starts the Windows Machine Config Bootstrapper
@@ -70,7 +75,7 @@ func runInitializeKubeletCmd(cmd *cobra.Command, args []string) {
 	wmcb, err := bootstrapper.NewWinNodeBootstrapper(initializeKubeletOpts.installDir,
 		initializeKubeletOpts.ignitionFile, initializeKubeletOpts.kubeletPath,
 		initializeKubeletOpts.nodeIP, initializeKubeletOpts.clusterDNS, "", "",
-		initializeKubeletOpts.platformType)
+		initializeKubeletOpts.platformType, initializeKubeletOpts.dockerRuntime)
 	if err != nil {
 		log.Error(err, "could not create bootstrapper")
 		os.Exit(1)
--- a/hack/run-wmcb-ci-e2e-test.sh
+++ b/hack/run-wmcb-ci-e2e-test.sh
@@ -57,6 +57,11 @@ fi
 sed -i "s~ARTIFACT_DIR_VALUE~${ARTIFACT_DIR}~g" internal/test/wmcb/deploy/job.yaml
 sed -i "s~REPLACE_IMAGE~${WMCB_IMAGE}~g" internal/test/wmcb/deploy/job.yaml
 
+# If DOCKER_RUNTIME is set, set that value for flag -dockerRuntime in job.yaml.
+# If it is unset, default value taken will be true.
+DOCKER_RUNTIME=${DOCKER_RUNTIME:-true}
+sed -i "s~DOCKER_RUNTIME~${DOCKER_RUNTIME}~g" internal/test/wmcb/deploy/job.yaml
+
 # deploy the test pod on test cluster
 if ! $OC apply -f internal/test/wmcb/deploy/job.yaml -n default; then
     echo "job already deployed"
--- a/internal/test/wmcb/wmcb_test.go
+++ b/internal/test/wmcb/wmcb_test.go
@@ -216,20 +216,18 @@ func (vm *wmcbVM) initializeTestConfigureCNIFiles(ovnHostSubnet string) error {
 // handleHybridOverlay ensures that the hybrid overlay is running on the node
 func (vm *wmcbVM) handleHybridOverlay(nodeName string) error {
 	// Check if the hybrid-overlay-node is running
-	output, err := vm.Run("Get-Process -Name \"hybrid-overlay-node\"", true)
-
-	// stderr being empty implies that an hybrid-overlay-node was running. This is to help with local development.
-	if err == nil || output == "" {
+	running := vm.isRunning(hybridOverlayServiceName)
+	if running {
 		return nil
 	}
 
 	// Wait until the node object has the hybrid overlay subnet annotation. Otherwise the hybrid-overlay-node will fail
 	// to start
-	if err = waitForNodeAnnotation(nodeName, test.HybridOverlaySubnet); err != nil {
+	if err := waitForNodeAnnotation(nodeName, test.HybridOverlaySubnet); err != nil {
 		return fmt.Errorf("error waiting for hybrid overlay subnet annotation: %v", err)
 	}
 
-	output, err = vm.Run(mkdirCmd(kLog), false)
+	output, err := vm.Run(mkdirCmd(kLog), false)
 	if err != nil {
 		return fmt.Errorf("unable to create remote directory %s: %v\n%s", kLog, err, output)
 	}
@@ -237,11 +235,9 @@ func (vm *wmcbVM) handleHybridOverlay(nodeName string) error {
 	// Start the hybrid-overlay-node Windows service
 	binPath := fmt.Sprintf("%s --node %s --windows-service --k8s-kubeconfig c:\\k\\kubeconfig "+
 		"--logfile %s\\hybrid-overlay.log", hybridOverlayExecutable, nodeName, kLog)
-	startCommand := fmt.Sprintf("sc.exe create %s binPath=\"%s\" start=auto && sc.exe start %s",
-		hybridOverlayServiceName, binPath, hybridOverlayServiceName)
-	out, err := vm.Run(startCommand, false)
-	if err != nil {
-		return errors.Wrapf(err, "failed to start service with output: %s", out)
+
+	if err = vm.ensureServiceIsRunning(hybridOverlayServiceName, binPath); err != nil {
+		fmt.Errorf("unable to start Windows service %s: %v", hybridOverlayServiceName, err)
 	}
 
 	// Wait for the hybrid-overlay to complete reconfiguring the network. The only way to detect that it has completed
@@ -295,6 +291,25 @@ func (vm *wmcbVM) waitForOpenShiftHNSNetworks() error {
 func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
 	err := vm.runTest(e2eExecutable + " --test.run TestKubeletUninstall --test.v")
 	require.NoError(t, err, "TestKubeletUninstall failed")
+}
+
+// isRunning checks if the given service is running
+func (vm *wmcbVM) isRunning(serviceName string) bool {
+	svcCmd := fmt.Sprintf("sc.exe qc %s", serviceName)
+	out, _ := vm.Run(svcCmd, false)
+	return strings.Contains(out, "RUNNING")
+}
+
+// ensureServiceIsRunning creates and starts a given Windows service
+func (vm *wmcbVM) ensureServiceIsRunning(serviceName string, binPath string) error {
+	startCmd := fmt.Sprintf("sc.exe create %s binPath=\"%s\" start=auto && sc.exe start %s", serviceName, binPath,
+		serviceName)
+
+	out, err := vm.Run(startCmd, false)
+	if err != nil {
+		return errors.Wrapf(err, "failed to start service with output: %s", out)
+	}
+	return nil
 }
 
 // mkdirCmd returns the Windows command to create a directory if it does not exists
--- a/internal/test/wmcb/wmcb_test.go
+++ b/internal/test/wmcb/wmcb_test.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"text/template"
@@ -54,6 +55,16 @@ const (
 	hybridOverlayConfigurationTime = 2 * time.Minute
 	// hybridOverlayServiceName is the name of hybrid-overlay Windows service
 	hybridOverlayServiceName = "hybrid-overlay-node"
+	//containerdServiceName is containerd Windows service name
+	containerdServiceName = "containerd"
+	// containerdDir is the directory for storing the containerd files in the payload directory
+	containerdDir = payloadDirectory + containerdServiceName
+	// winContainerdDir is the directory where the containerd files are placed in remote directory
+	winContainerdDir = remoteDir + containerdServiceName
+	// containerdPath is the location of the containerd exe
+	containerdPath = remoteDir + containerdServiceName + "\\containerd.exe"
+	// containerdConfPath is the location of containerd config file
+	containerdConfPath = remoteDir + containerdServiceName + "\\containerd_conf.toml"
 )
 
 var (
@@ -91,6 +102,9 @@ func TestWMCB(t *testing.T) {
 	srcDestPairs := map[string]string{
 		payloadDirectory: remoteDir,
 		cniDirectory:     winCNIDir,
+	}
+	if !dockerRuntime {
+		srcDestPairs[containerdDir] = winContainerdDir
 	}
 
 	for _, vm := range framework.WinVMs {
@@ -144,7 +158,13 @@ func (vm *wmcbVM) runTestBootstrapper(t *testing.T) {
 	err := vm.initializeTestBootstrapperFiles()
 	require.NoError(t, err, "error initializing files required for TestBootstrapper")
 
-	err = vm.runTest(e2eExecutable + " --test.run TestBootstrapper --test.v --platform-type=aws")
+	if !dockerRuntime {
+		err = vm.configureContainerd()
+		require.NoError(t, err, "error running containerd Windows service")
+	}
+
+	err = vm.runTest(e2eExecutable + " --test.run TestBootstrapper --test.v --platform-type=aws --dockerRuntime=" +
+		strconv.FormatBool(dockerRuntime))
 	require.NoError(t, err, "TestBootstrapper failed")
 }
 
@@ -161,7 +181,8 @@ func (vm *wmcbVM) runTestConfigureCNI(t *testing.T) {
 	err = vm.initializeTestConfigureCNIFiles(hybridOverlayAnnotation)
 	require.NoError(t, err, "error initializing files required for TestConfigureCNI")
 
-	err = vm.runTest(e2eExecutable + " --test.run TestConfigureCNI --test.v --platform-type=aws")
+	err = vm.runTest(e2eExecutable + " --test.run TestConfigureCNI --test.v --platform-type=aws --dockerRuntime=" +
+		strconv.FormatBool(dockerRuntime))
 	require.NoError(t, err, "TestConfigureCNI failed")
 }
 
@@ -177,6 +198,14 @@ func (vm *wmcbVM) initializeTestBootstrapperFiles() error {
 	_, err = vm.Run("cp "+remoteDir+"kubelet.exe "+winTemp, true)
 	if err != nil {
 		return fmt.Errorf("unable to copy kubelet.exe to %s: %v", winTemp, err)
+	}
+
+	if !dockerRuntime {
+		// create the containerd directory on Windows VM.
+		output, err := vm.Run(mkdirCmd(winContainerdDir), false)
+		if err != nil {
+			return fmt.Errorf("unable to create remote directory %s: %v\n%s", winContainerdDir, err, output)
+		}
 	}
 
 	// Ignition v2.3.0 maps to Ignition config spec v3.1.0.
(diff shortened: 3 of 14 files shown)


hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -39,6 +39,12 @@ if [ -z "$ARTIFACT_DIR" ]; then
   export ARTIFACT_DIR=$ARTIFACT_DIR
 fi
 
+DOCKER_RUNTIME=${DOCKER_RUNTIME:-}
+if [ -z "$DOCKER_RUNTIME" ]; then
+  echo "DOCKER_RUNTIME is not set. Defaulting to true"
```
jrvaldes, line 44: The `--dockerRuntime` flag option defaults to true [already](https://github.com/openshift/windows-machine-config-bootstrapper/pull/322/commits/de59d0c49effd43b67dd8a90223fb281715463ee#diff-933b7d0fd790e2d3de8ea9dc1185e3441f8ea6aea0adac993cadb0567b9296fbR65) if is not provided.

Ideally, if `DOCKER_RUNTIME` is provided you should set the `-dockerRuntime=true` arg in  `internal/test/wmcb/deploy/job.yaml`.

test/e2e/bootstrapper_test.go
```diff
@@ -101,7 +103,12 @@ func TestBootstrapper(t *testing.T) {
 	// Does not include node-labels and container-image since their paths do not depend on underlying OS
 	checkPathsFor := []string{"--bootstrap-kubeconfig", "--cloud-config", "--config", "--kubeconfig", "--log-file",
 		"--cert-dir"}
-	expectedDependencies := []string{"docker"}
```
jrvaldes, line 103: You can keep docker as default in `expectedDependencies` and overwrite if `wmcb.dockerRuntime` evaluates to false.

internal/test/wmcb/wmcb_test.go
```diff
@@ -144,7 +160,13 @@ func (vm *wmcbVM) runTestBootstrapper(t *testing.T) {
 	err := vm.initializeTestBootstrapperFiles()
 	require.NoError(t, err, "error initializing files required for TestBootstrapper")
 
-	err = vm.runTest(e2eExecutable + " --test.run TestBootstrapper --test.v --platform-type=aws")
+	if !*dockerRuntime {
```
jrvaldes, line 163: After reading the proposed changes, I feel the flag name should be `--containerdRuntime`. It will lean the execution flow towards the positive side, instead of always negating the proposed flag.

I know it's late for this comment 😞

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +328,48 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
...
+
+	// set Windows defender exclusions for containerd
+	setExclusionCmd := fmt.Sprintf("Add-MpPreference -ExclusionProcess %s", containerdPath)
+	out, err = vm.Run(setExclusionCmd, true)
+	if err != nil {
+		return fmt.Errorf("setting Windows defender process exclusion failed: %v", err)
+	}
+
+	// Start the containerd Windows service
+	binPath := fmt.Sprintf("%s --config %s --log-file %s\\containerd.log --log-level=info --run-service",
```
jrvaldes, line 354: Considering this is an in-development feature, should we start with `--log-level=debug` or will be too much?

jrvaldes (commented):
@mansikulkarni96 Great work! Thanks for doing this, comments added. PTAL

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -60,6 +62,8 @@ func init() {
 			"If unset, kubelet will determine the DNS server to use.")
 	initializeKubeletCmd.PersistentFlags().StringVar(&initializeKubeletOpts.platformType, "platform-type", "",
 		"Type of the platform where the cluster is deployed. Example: AWS, Azure, GCP")
+	initializeKubeletCmd.PersistentFlags().BoolVar(&initializeKubeletOpts.dockerRuntime, "dockerRuntime", true,
+		"Container runtime to be used Example: docker, containerd")
```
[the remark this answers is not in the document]
saifshaikh48: Would suggest keeping a containerd mention as that is the alternative.  e.g. appending `If false, containerd will be used as the default container runtime.`

pkg/bootstrapper/bootstrapper.go
```diff
@@ -588,17 +595,21 @@ func (wmcb *winNodeBootstrapper) ensureKubeletService() error {
 	c := mgr.Config{
 		ServiceType: 0,
 		// StartAutomatic will start the service again if the node restarts
-		StartType:      mgr.StartAutomatic,
-		ErrorControl:   0,
-		LoadOrderGroup: "",
-		TagId:          0,
-		// set dependency on docker
-		Dependencies:     []string{"docker"},
```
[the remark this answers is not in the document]
saifshaikh48: +1 to this suggestion. Nit that makes it slightly easier/less code churn when we need to remove the flag

pkg/bootstrapper/bootstrapper_test.go
```diff
@@ -298,6 +300,7 @@ func TestKubeletArgs(t *testing.T) {
 				kubeletConfPath: filepath.Join("/fakepath/kubelet.conf"),
 				logDir:          "/fakepath/",
 				nodeIP:          "192.168.1.1",
+				dockerRuntime:   true,
```
saifshaikh48, line 303: Shouldn't we also add a test for kubelet args when runtime is containerd?

test/e2e/configure_cni_test.go
```diff
@@ -73,9 +73,12 @@ func testConfigureCNI(t *testing.T) {
...
 	assert.True(t, isKubeletRunning(t, kubeletLogPath))
-	isConfiguredCorrectly, err := isCNIConfigured(t, kubeletLogPath)
-	require.NoError(t, err, "Error reading kubelet log")
-	assert.True(t, isConfiguredCorrectly, "CNI was not configured correctly")
+	// if the runtime is containerd, the network plugin is loaded at a later stage.
+	if dockerRuntime {
+		isConfiguredCorrectly, err := isCNIConfigured(t, kubeletLogPath)
+		require.NoError(t, err, "Error reading kubelet log")
+		assert.True(t, isConfiguredCorrectly, "CNI was not configured correctly")
+	}
```
saifshaikh48, line 81: Not sure if I understand correctly -- if the node picks up CNI networking changes later with containerd, would we be able to run this test later on in the test suite? Or wait for some time and then run it? I just don't want to skip this test case entirely if containerd is the runtime.

saifshaikh48 (commented):
Looks good, though some questions on testing.

test/e2e/configure_cni_test.go
mansikulkarni96: We are suspecting this log might be actually coming from docker runtime hence we are not seeing it in containerd case, I am looking into it more closely now will update soon.

hack/run-wmcb-ci-e2e-test.sh
mansikulkarni96: yes, DOCKER_RUNTIME value is appended in internal/test/wmcb/deploy/job.yaml, this was just a safe check to avoid appending null value if the var is not set. Modified it to removing the line if the value is not set.

pkg/internal/containerd_conf.toml
```diff
@@ -0,0 +1,174 @@
...
+    [plugins."io.containerd.grpc.v1.cri".registry]
+      config_path = ""
+
+      [plugins."io.containerd.grpc.v1.cri".registry.auths]
+
+      [plugins."io.containerd.grpc.v1.cri".registry.configs]
+
+      [plugins."io.containerd.grpc.v1.cri".registry.headers]
+
+      [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
```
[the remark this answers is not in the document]
selansen: This is generated file. I would prefer to keep the changes as minimal as possible. We are making changes that are required only for our needs.  In future it will be easy to take a diff and identify our changes with the generated code.

internal/test/wmcb/wmcb_test.go
selansen: Agree. It was just --Runtime and it was easy to read the code flow. Then we changed it to bool and still wanted to give preference to docker runtime and use contained only for testing and community operator. Hence all the confusion :)

cmd/bootstrapper/initialize_kubelet.go
```diff
@@ -41,6 +41,8 @@ var (
 		clusterDNS string
 		// platformType contains type of the platform where the cluster is deployed
 		platformType string
+		// dockerRuntime indicates if the container runtime to be used is docker
```
selansen, line 44: Can you pls add an extra line saying:
If true docker will be the runtime, false means containers will be the runtime.

cmd/bootstrapper/initialize_kubelet.go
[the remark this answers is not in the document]
selansen: > Would suggest keeping a containerd mention as that is the alternative. e.g. appending `If false, containerd will be used as the default container runtime.`

+1

pkg/bootstrapper/bootstrapper.go
```diff
@@ -147,6 +147,8 @@ type winNodeBootstrapper struct {
 	cni *cniOptions
 	// platformType contains type of the platform where the cluster is deployed
 	platformType string
+	// dockerRuntime indicates if the container runtime to be used is docker
```
selansen, line 150: same as initialize_kubelet.go comment

pkg/internal/containerd_conf.toml
```diff
@@ -0,0 +1,174 @@
...
+    stats_collect_period = 10
+    stream_idle_timeout = "4h0m0s"
+    stream_server_address = "127.0.0.1"
+    stream_server_port = "0"
+    systemd_cgroup = false
+    tolerate_missing_hugetlb_controller = false
+    unset_seccomp_profile = ""
+
+    [plugins."io.containerd.grpc.v1.cri".cni]
+      bin_dir = "C:\\k\\cni"
```
selansen, line 67: Could you pls point me where do we update cni.conf file info here?

mansikulkarni96: Not sure what info we need to update here as we are using the same cni.conf file with the same network info as before. If you are referring to renaming the file to` 0-containerd-nat.conf`, I tested it out and that is not a required change we can go ahead with cni.conf naming as long as it is the only config file present in the bin dir.

selansen: That is what I am referring. I was thinking we may need to explicitly mention new cni conf file name. 
It it has been tested, then I am fine. If I remember when we tested, cni.conf was the only file inside the conf folder and it was not working, right?

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +328,48 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
 	require.NoError(t, err, "TestKubeletUninstall failed")
 }
 
+// configureContainerd installs the containerd Windows service
+func (vm *wmcbVM) configureContainerd() error {
+	// Check if containerd is running
+	out, err := vm.Run("Get-Process -Name \"containerd\"", true)
+
+	// stderr being empty implies that containerd was running.
+	if err == nil || out == "" {
```
saifshaikh48, line 337: I am not clear on why the second clause `|| out == ""` is included in this check. What if the command results in an error without any output? I've seen the following for example:
```
error running	{"cmd": "powershell.exe -NonInteractive -ExecutionPolicy Bypass \"Test-Path C:\\k\\cni\\\\win-bridge.exe\"", "out": "", "error": "read tcp 192.168.81.68:36090->192.168.81.111:22: read: connection reset by peer"}
```

saifshaikh48 (commented):
LGTM aside from one concern

pkg/internal/containerd_conf.toml
```diff
@@ -0,0 +1,174 @@
...
+    path = "C:\\ProgramData\\containerd\\root\\opt"
+
+  [plugins."io.containerd.internal.v1.restart"]
+    interval = "10s"
+
+  [plugins."io.containerd.metadata.v1.bolt"]
+    content_sharing_policy = "shared"
+
+  [plugins."io.containerd.runtime.v2.task"]
+    platforms = ["windows/amd64", "linux/amd64"]
```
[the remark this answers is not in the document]
sebsoto: @mansikulkarni96

pkg/internal/containerd_conf.toml
sebsoto: It would be easeier to review if you split the generated config file into a separate commit, and have another commit making changes to it

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -57,6 +57,14 @@ fi
 sed -i "s~ARTIFACT_DIR_VALUE~${ARTIFACT_DIR}~g" internal/test/wmcb/deploy/job.yaml
 sed -i "s~REPLACE_IMAGE~${WMCB_IMAGE}~g" internal/test/wmcb/deploy/job.yaml
 
+DOCKER_RUNTIME=${DOCKER_RUNTIME:-}
+
+if [ ! -z "$DOCKER_RUNTIME" ]; then
+  sed -i "s~DOCKER_RUNTIME~${DOCKER_RUNTIME}~g" internal/test/wmcb/deploy/job.yaml
+else
+  sed -i '/-dockerRuntime=/d' internal/test/wmcb/deploy/job.yaml
+fi
```
sebsoto, line 66: please add comments for what you are doing here

internal/test/wmcb/wmcb_test.go
```diff
@@ -54,6 +56,16 @@ const (
 	hybridOverlayConfigurationTime = 2 * time.Minute
 	// hybridOverlayServiceName is the name of hybrid-overlay Windows service
 	hybridOverlayServiceName = "hybrid-overlay-node"
+	// containerdDir is the directory for storing the CNI plugins
+	containerdDir = payloadDirectory + containerdServiceName
```
sebsoto, line 60: name doesnt match with description, please change to cniDir if description is accurate

internal/test/wmcb/wmcb_test.go
```diff
@@ -92,6 +105,9 @@ func TestWMCB(t *testing.T) {
 		payloadDirectory: remoteDir,
 		cniDirectory:     winCNIDir,
 	}
+	if !*dockerRuntime {
+		srcDestPairs[containerdDir] = winContainerdDir
+	}
```
sebsoto, line 110: You're dereferencing a pointer without checking if it is nil

internal/test/wmcb/wmcb_test.go
```diff
@@ -63,6 +75,7 @@ var (
 		Value:  "Windows",
 		Effect: v1.TaintEffectNoSchedule,
 	}
+	dockerRuntime = flag.Bool("dockerRuntime", true, "Container runtime to be used")
```
sebsoto, line 78: where is flag.Parse() being called?
this flag should be defined where the other flags are being defined, flags come command line arguments and should be processed at the start of the program, not throughout the code.

internal/test/wmcb/wmcb_test.go
sebsoto: Its better to use https://pkg.go.dev/flag#BoolVar
so you dont have to check if it is nil or not

internal/test/wmcb/wmcb_test.go
sebsoto: Please make this a shared function with any Windows process instead of [having duplicate code](https://github.com/openshift/windows-machine-config-bootstrapper/blob/38466b767564bddeeb68a6a9fab708dd73a937ce/internal/test/wmcb/wmcb_test.go#L218-L224)

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +328,48 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
...
+	}
+
+	// set Windows defender exclusions for containerd
+	setExclusionCmd := fmt.Sprintf("Add-MpPreference -ExclusionProcess %s", containerdPath)
+	out, err = vm.Run(setExclusionCmd, true)
+	if err != nil {
+		return fmt.Errorf("setting Windows defender process exclusion failed: %v", err)
+	}
+
+	// Start the containerd Windows service
```
sebsoto, line 353: Starting should also be made into a [shared function](https://github.com/openshift/windows-machine-config-bootstrapper/blob/38466b767564bddeeb68a6a9fab708dd73a937ce/internal/test/wmcb/wmcb_test.go#L237-L245)

sebsoto (changes requested):

internal/test/wmcb/wmcb_test.go
mansikulkarni96: yes initially thought of doing that, missed my later on. Thanks for pointing out the duplicity.

pkg/internal/containerd_conf.toml
alinaryan: +1

internal/test/wmcb/wmcb_test.go
mansikulkarni96: @saifshaikh48 we were always using the same [condition](https://github.com/openshift/windows-machine-config-bootstrapper/blob/38466b767564bddeeb68a6a9fab708dd73a937ce/internal/test/wmcb/wmcb_test.go#L218-L224).
Looking at vm.Run() method, it only returns err with out  "" OR out with an err = nil.
I am trying to change the method to use sc.exe but running into some powershell issues, if it's a low hanging fruit I ll handle it.

pkg/internal/containerd_conf.toml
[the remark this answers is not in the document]
mansikulkarni96: same comment as pointed out by @selansen https://github.com/openshift/windows-machine-config-bootstrapper/pull/322#discussion_r816051481

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -57,6 +57,17 @@ fi
...
+DOCKER_RUNTIME=${DOCKER_RUNTIME:-}
+
+# If DOCKER_RUNTIME is set, set the value for flag -dockerRuntime in job.yaml.
+# If it is unset, remove the line with -dockerRuntime flag to avoid passing
+# flag with nil value. Default value for the flag will be true in this case.
+if [ ! -z "$DOCKER_RUNTIME" ]; then
+  sed -i "s~DOCKER_RUNTIME~${DOCKER_RUNTIME}~g" internal/test/wmcb/deploy/job.yaml
+else
+  sed -i '/-dockerRuntime=/d' internal/test/wmcb/deploy/job.yaml
+fi
```
saifshaikh48, line 69:
```suggestion
DOCKER_RUNTIME=${DOCKER_RUNTIME:-true}

sed -i "s~DOCKER_RUNTIME~${DOCKER_RUNTIME}~g" internal/test/wmcb/deploy/job.yaml
```
Could this simplified by providing a default value?

internal/test/wmcb/wmcb_test.go
```diff
@@ -54,6 +55,16 @@ const (
 	hybridOverlayConfigurationTime = 2 * time.Minute
 	// hybridOverlayServiceName is the name of hybrid-overlay Windows service
 	hybridOverlayServiceName = "hybrid-overlay-node"
+	// containerdDir is the directory for storing the containerd files in the payload directory
+	containerdDir = payloadDirectory + containerdServiceName
+	// winContainerdDir is the directory where the containerd files are placed in remote directory
+	winContainerdDir = remoteDir + containerdServiceName
+	//containerdServiceName is containerd Windows service name
+	containerdServiceName = "containerd"
```
saifshaikh48, line 63: nit, move `containerdServiceName` definition above `containerdDir` and `winContainerdDir` which use it

hack/run-wmcb-ci-e2e-test.sh
mansikulkarni96: yes I proposed that in the first revision of the PR but changed after I received a comment from Jose pointing out the default value would be true even if the flag is not passed. I am fine with either approach whichever is easier to read.

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +293,25 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
...
 
+// isRunning checks if the given service is running
+func (vm *wmcbVM) isRunning(serviceName string) bool {
+	svcCmd := fmt.Sprintf("sc.exe qc %s", serviceName)
+	out, _ := vm.Run(svcCmd, false)
+	return strings.Contains(out, "RUNNING")
+}
+
+// startService starts a given Windows service
+func (vm *wmcbVM) startService(serviceName string, binPath string) error {
```
sebsoto, line 364: This is both starting and creating the service and the name should reflect that

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +293,25 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
 	require.NoError(t, err, "TestKubeletUninstall failed")
 }
 
+// isRunning checks if the given service is running
+func (vm *wmcbVM) isRunning(serviceName string) bool {
+	svcCmd := fmt.Sprintf("sc.exe qc %s", serviceName)
```
sebsoto, line 358: this is a change in functionality, checking if the process is running vs checking if the service is running and should be reflected in the commit message

hack/run-wmcb-ci-e2e-test.sh
mansikulkarni96: updated

hack/run-wmcb-ci-e2e-test.sh
```diff
@@ -57,6 +57,11 @@ fi
 sed -i "s~ARTIFACT_DIR_VALUE~${ARTIFACT_DIR}~g" internal/test/wmcb/deploy/job.yaml
 sed -i "s~REPLACE_IMAGE~${WMCB_IMAGE}~g" internal/test/wmcb/deploy/job.yaml
 
+# If DOCKER_RUNTIME is set, set that value for flag -dockerRuntime in job.yaml.
+# If it is unset, default value taken will be true.
+DOCKER_RUNTIME=${DOCKER_RUNTIME:-true}
```
alinaryan, line 62: For my own learning, why is the DOCKER_RUNTIME env var being taken from the release repo AWS e2e CCM config specifically?

internal/test/wmcb/wmcb_test.go
```diff
@@ -297,6 +293,25 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
...
+	out, _ := vm.Run(svcCmd, false)
+	return strings.Contains(out, "RUNNING")
+}
+
+// ensureServiceIsRunning creates and starts a given Windows service
+func (vm *wmcbVM) ensureServiceIsRunning(serviceName string, binPath string) error {
```
alinaryan, line 364: These two helper functions are great and make the code much cleaner, but shouldn't they be in a separate PR? I'm not sure how they relate to this one.

alinaryan: My mistake, I see them being used in the next commit

hack/run-wmcb-ci-e2e-test.sh
mansikulkarni96: @alinaryan We are passing dockerRuntime=false from CCM job so we can test the installation and setting of containerd as the runtime through kubelet arguments using only CCM job.

internal/test/wmcb/wmcb_test.go
```diff
@@ -293,6 +322,37 @@ func (vm *wmcbVM) runTestKubeletUninstall(t *testing.T) {
 	require.NoError(t, err, "TestKubeletUninstall failed")
 }
 
+// configureContainerd installs the containerd Windows service
+func (vm *wmcbVM) configureContainerd() error {
+	// Check if containerd is running
+	running := vm.isRunning(hybridOverlayServiceName)
```
selansen, line 328: I am confused a bit.  Comment says check if containerd is running but we are checking hybridOverlayServiceName here.

mansikulkarni96: Thanks for the catch, must have messed this up during rebase with copy/paste

pkg/bootstrapper/bootstrapper.go
```diff
@@ -569,6 +575,10 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
 	if wmcb.nodeIP != "" {
 		kubeletArgs = append(kubeletArgs, "--node-ip="+wmcb.nodeIP)
 	}
+	if !wmcb.dockerRuntime {
+		kubeletArgs = append(kubeletArgs, "--container-runtime=remote",
```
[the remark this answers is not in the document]
mansikulkarni96: It was suggested by @sebsoto in this [comment](https://github.com/openshift/windows-machine-config-bootstrapper/pull/322#discussion_r816848050), so it can remain as a reference within code.

selansen:
Thanks for working on this and addressing all comments.

/lgtm

saifshaikh48 (commented):
/lgtm

We should make sure QE is aware of the payload binary additions (containerd, hcsshim).

pkg/internal/containerd_conf.toml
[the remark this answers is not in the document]
jrvaldes: @selansen  How does this file gets generated? Are there any configuration flags to narrow the generation?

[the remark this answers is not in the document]
mansikulkarni96: @jrvaldes generation has been mentioned in the commit message.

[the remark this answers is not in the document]
jrvaldes: Oh! Missed that. Thanks @mansikulkarni96 👍🏼

PR #323, opened 2022-02-03
Run hybrid-overlay as a Windows service in e2e tests

sebsoto:


internal/test/wmcb/wmcb_test.go
```diff
@@ -221,24 +223,25 @@ func (vm *wmcbVM) handleHybridOverlay(nodeName string) error {
 		return nil
 	}
 
-	// Wait until the node object has the hybrid overlay subnet annotation. Otherwise the hybrid-overlay-node will fail to
-	// start
-	if err = waitForHybridOverlayAnnotation(nodeName); err != nil {
-		return fmt.Errorf("error waiting for hybrid overlay node annotation: %v", err)
+	// Wait until the node object has the hybrid overlay subnet annotation. Otherwise the hybrid-overlay-node will fail
```
aravindhp, line 226: You need to fix these lines:
```
	// Check if the hybrid-overlay-node is running
	output, err := vm.Run("Get-Process -Name \"hybrid-overlay-node\"", true)
```
to check if the service is running.

saifshaikh48 (commented):
/lgtm
Thanks for doing this @sebsoto

alinaryan:
/lgtm
Thanks for taking care of this so quickly!

sebsoto:
CI passed, retesting a few times to ensure flake is gone
/retest

internal/test/wmcb/wmcb_test.go
sebsoto: I think thats a potential failure point
running hybrid overlay causes the SSH connection to be broken, so based on timing, that check will fail if done while the network interface is being reset
It could be moved to after the wait period, but at that point we can just check for the network creation/ MAC address and get the same confirmation that hybrid-overlay has done its job.

sebsoto:
Tests passed again retrying one more time

internal/test/wmcb/wmcb_test.go
aravindhp: How is that a potential failure point? This is the first thing we do in handleHybridOverlay(). IOW, hybrid-overlay has not been run yet.

sebsoto: My mistake, was confusing that with the waitForHybridOverlayToRun function.

Its a lot simpler to just check that the process is running, instead of parsing the output of sc.exe qc

aravindhp: You don't need to use `sc.exe` but use `Get-Service -Name kubelet | Where-Object {$_.Status -eq "Running"}` or something similar. But I am fine leaving it as is.

sebsoto: That command would need to be modified to throw an error if the list is empty, i'd prefer to just leave it as it is for the sake of simplicity

aravindhp:
@sebsoto don't you need to backport this to the 4.8 branches also?

PR #347, opened 2022-04-20
Revert addition of dockerRuntime flag

mansikulkarni96:
This PR reverts the addition and usage of dockerRuntime flag option. It was added with the anticipation of using docker and containerd runtimes in the same release.
However containerd feature is now expected to be fully supported in 6.0 time-frame while removing support for docker container runtime which makes this flag option no longer useful.


selansen (commented):
/lgtm

sebsoto:
This is a breaking change that needs to be merged in only after WMCO is set up to do all the required containerd steps by default

PR #358, opened 2022-08-22
Add optional parameter for kubelet verbosity

jrvaldes:
This change introduces an optional program argument to
set the kubelet log level. If unset, wmcb.exe will use
the verbosity value in the kubelet systemd unit provided
by the ignition file (if any) or default to 3. The novel
program argument takes precedence over the later two.

### DIFF
--- a/pkg/bootstrapper/bootstrapper.go
+++ b/pkg/bootstrapper/bootstrapper.go
@@ -40,6 +40,8 @@ const (
 	// KubeletServiceName is the name will we run the kubelet Windows service under. It is required to be named "kubelet":
 	// https://github.com/kubernetes/kubernetes/blob/v1.16.0/cmd/kubelet/app/init_windows.go#L26
 	KubeletServiceName = "kubelet"
+	// KubeletDefaultVerbosity is the recommended default log level for kubelet. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
+	KubeletDefaultVerbosity = "3"
 	// kubeletDependentSvc is the name of the service dependent on kubelet Windows service
 	kubeletDependentSvc = "hybrid-overlay-node"
 	// kubeletSystemdName is the name of the systemd service that the kubelet runs under,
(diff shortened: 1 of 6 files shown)


pkg/bootstrapper/bootstrapper.go
```diff
@@ -482,12 +487,19 @@ func (wmcb *winNodeBootstrapper) generateInitialKubeletArgs(args map[string]stri
...
-		kubeletArgs = append(kubeletArgs, "--v="+v)
-	} else {
-		// In case the verbosity argument is missing, use a default value
-		kubeletArgs = append(kubeletArgs, "--v="+"3")
+		// and update, if found
+		kubeletVerbosity = v
+	}
+	// set the verbosity value from the program argument if any, this takes precedence
+	if wmcb.kubeletVerbosity != "" {
+		kubeletVerbosity = wmcb.kubeletVerbosity
```
aravindhp, line 499: This looks confusing. You should key off `wmcb.kubeletVerbosity` being set or not. 
```suggestion
	// set default verbosity for kubelet
	kubeletVerbosity := KubeletDefaultVerbosity
    // set the verbosity value from the program argument if any, this takes precedence
	if wmcb.kubeletVerbosity != "" {
		kubeletVerbosity = wmcb.kubeletVerbosity
	} else {
    	// look for verbosity value in kubelet systemd unit file configuration
    	if v, ok := args["v"]; ok && v != "" {
	    	// and update, if found
		    kubeletVerbosity = v
    	}
    }
```
You shouldn't have to look at the systemd unit if the verbosity is specified.

aravindhp (changes requested):

pkg/bootstrapper/bootstrapper.go
jrvaldes: Good catch! Addressed in 1637182a088084487a20c257f9846789231ee89d

aravindhp (commented):
/lgtm

I will defer to @sebsoto for approval

aravindhp (approved):

aravindhp:
> /cherry-pick release-4.12

@jrvaldes there is no reason to cherry pick to release-4.12 as it is still being ff from master.

Is there a reason not to backport this to all supported releases?

jrvaldes:
> > /cherry-pick release-4.12
> 
> @jrvaldes there is no reason to cherry pick to release-4.12 as it is still being ff from master.
> 


 cherry pick to release-4.12  was a typo.

jrvaldes:
> Is there a reason not to backport this to all supported releases?

No, plan to do it one at a time.

PR #363, opened 2022-10-17
OCPBUGS-2028:  Override kubelet hostname for GCP platform

jrvaldes:
This change fetches the hostname from the GCP instance metadata service
and overrides the kubelet hostname, so that the name of the Windows node
matches the hostname.

### DIFF
--- a/pkg/cloud/gcp.go
+++ b/pkg/cloud/gcp.go
@@ -1,0 +1,39 @@
+package cloud
+
+import (
+	"strings"
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/pkg/errors"
+)
+
+// gcpHostnameMaxLength is the maximum number of character allowed for the instance's hostname in GCP
+const gcpHostnameMaxLength = 63
+
+// processGCPInstanceHostname truncates the given hostname if the length is longer 63 characters, otherwise returns
+// the original value
+func processGCPInstanceHostname(hostname string) string {
+	// check hostname length
+	if len(hostname) > gcpHostnameMaxLength {
+		// check if hostname shorter than 63 characters before the first dot in the FQDN
+		firstDotIndex := strings.Index(hostname, ".")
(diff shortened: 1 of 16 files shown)


mansikulkarni96:
For my understanding, what is the difference between the default behaviour that truncates the hostname over 63 characters and the code introduced in this PR?

jrvaldes:
> For my understanding, what is the difference between the default behaviour that truncates the hostname over 63 characters and the code introduced in this PR?


@mansikulkarni96, there is no default behavior that truncates the hostname for Windows nodes, it only [exists](https://github.com/openshift/machine-config-operator/blob/e46a003cdb705f30ace3acdfff830ca36446f4e5/templates/common/_base/files/usr-local-bin-mco-hostname.yaml#L25) for Linux instances. This PR introduces this functionality for Windows instances in GCP.

jrvaldes:
/close

As WMCB is getting deprecated, this change should be part of WICD's kubelet service implementation and proposed in a separate PR after the following PRs merge:
- https://github.com/openshift/windows-machine-config-operator/pull/1260
- https://github.com/openshift/windows-machine-config-operator/pull/1279

pkg/cloud/gcp.go
```diff
@@ -0,0 +1,39 @@
...
+	"cloud.google.com/go/compute/metadata"
+	"github.com/pkg/errors"
+)
+
+// gcpHostnameMaxLength is the maximum number of character allowed for the instance's hostname in GCP
+const gcpHostnameMaxLength = 63
+
+// processGCPInstanceHostname truncates the given hostname if the length is longer 63 characters, otherwise returns
+// the original value
+func processGCPInstanceHostname(hostname string) string {
```
aravindhp, line 15: Is there an existing function in MAO or MCO that you can leverage?

jrvaldes: I looked and couldn't find anything.

